PR #2, opened 2016-05-02
Implement custom GAPIC header in PHP.

shinfan:
- Move all descriptors from CallSettings to API level.
- Testing: unit tests

### DIFF
--- a/src/AgentHeaderDescriptor.php
+++ b/src/AgentHeaderDescriptor.php
@@ -29,58 +29,65 @@
  * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
  * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
+
 namespace Google\GAX;
 
 use InvalidArgumentException;
 
 /**
- * Holds the description information used for page streaming.
+ * Encapsulates the custom GAPIC header information.
  */
-class PageStreamingDescriptor
+class AgentHeaderDescriptor
 {
-    private $requestPageTokenField;
-    private $responsePageTokenField;
-    private $resourceField;
+    const AGENT_HEADER_KEY = 'x-google-apis-agent';
+
+    private $clientName;
+    private $clientVersion;
+    private $codeGenName;
+    private $codeGenVersion;
+    private $gaxVersion;
+    private $phpVersion;
 
     /**
-     * @param array $descriptor {
+     * @param array $headerInfo {
      *     Required.
      *
-     *     @type string $requestPageTokenField the page token field in the request object.
-     *     @type string $responsePageTokenField the page token field in the response object.
-     *     @type string $resourceField the resource field in the response object.
+     *     @type string $clientName the name of the client application.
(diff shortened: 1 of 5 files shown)


src/CallSettings.php
```diff
@@ -145,12 +138,9 @@ private static function constructRetry($methodConfig, $statusCodes,
     /**
      * Used internally to create instances of CallSettings.
      */
-    public static function createInternal($settings)
+    public static function createInternal($settings = [])
```
garrettjonesgoogle: Delete this function.

src/ApiCallable.php
```diff
@@ -111,12 +112,34 @@ private static function setPageStreaming($callable, $pageStreamingDescriptor)
         return $inner;
     }
 
+    private static function setCustomHeader($callable, $headerDescriptor)
+    {
+        $inner = function() use ($callable, $headerDescriptor) {
+            $params = func_get_args();
+            if (count($params) != self::GRPC_CALLABLE_PARAM_COUNT ||
+                gettype($params[self::GRPC_CALLABLE_METADATA_INDEX]) != 'array') {
```
garrettjonesgoogle: I know this pattern was already used above, but I think it would be better to use the built-in function is_array (http://php.net/manual/en/function.is-array.php)

src/AgentHeaderDescriptor.php
```diff
@@ -0,0 +1,94 @@
...
+    {
+        $requiredFields = ['clientName', 'clientVersion', 'codeGenName',
+            'codeGenVersion', 'gaxVersion', 'phpVersion'];
+        foreach ($requiredFields as $field) {
+            if (empty($descriptor[$field])) {
+                throw new InvalidArgumentException(
+                    "$field is required for AgentHeaderDescriptor");
+            }
+        }
+    }
```
garrettjonesgoogle: I think we have enough examples of this pattern to create a utility function for it (validating required fields)

src/ApiCallable.php
```diff
@@ -111,12 +112,34 @@ private static function setPageStreaming($callable, $pageStreamingDescriptor)
...
      * @param string $methodName the method name on the stub to call.
      * @param \Google\GAX\CallSettings the call settings to use for this call.
+     * @param \Google\GAX\PageStreamingDescriptor
+     *     the descriptor used for page-streaming. Default is null.
+     * @param \Google\GAX\AgentHeaderDescriptor
+     *     the descriptor used for creating GAPIC header. Default is null.
      */
-    public static function createApiCall($stub, $methodName, CallSettings $settings)
+    public static function createApiCall($stub, $methodName, CallSettings $settings,
+        $pageStreamingDescriptor = null, $headerDescriptor = null)
```
garrettjonesgoogle: I think I would prefer to take the descriptors as an options array, since the parameter list is getting a little long.

src/AgentHeaderDescriptor.php
shinfan: I can fix this in a separate PR.

garrettjonesgoogle: ok.

shinfan:
PTAL

garrettjonesgoogle:
LGTM

PR #3, opened 2016-05-03
Fix incorrect header format in API callable.

shinfan:


garrettjonesgoogle:
LGTM

src/ApiCallable.php
```diff
@@ -121,9 +121,12 @@ private static function setCustomHeader($callable, $headerDescriptor)
                 throw new InvalidArgumentException('Metadata argument is not found.');
             } else {
                 $metadata = $params[self::GRPC_CALLABLE_METADATA_INDEX];
-                $params[self::GRPC_CALLABLE_METADATA_INDEX] =
-                    array_merge($metadata, $headerDescriptor->getHeader());
-                call_user_func_array($callable, $params);
+                $headers = $headerDescriptor->getHeader();
+                if (array_key_exists('headers', $metadata)) {
+                    $headers = array_merge($headers, $metadata['headers']);
```
michaelbausor: Do we want the values from $headers to overwrite the existing values in $metadata['headers']? Because it seems that array_merge takes values from the later argument when there is duplication of keys.
http://php.net/manual/en/function.array-merge.php

shinfan: I think it would very rare that the keys are duplicate unless the user specifically sets the header key as ''x-google-apis-agent", which I am not sure if we should override it.

garrettjonesgoogle: Our generated wrappers won't provide a way to wrapper callers to override it; it might be handy for our wrappers themselves to be able to override it though. It might be worth clarifying this in the docs of ApiCallable, in any case.

Issue #7, opened 2016-05-16
Rewrite CallSettings.load to accept config_overrides


jmuk:
Based on the discussion in https://github.com/googleapis/gax-python/issues/95, we are going to accept a parameter `config_overrides` to construct_settings, and retire `retrying_overrides`.

`config_overrides` is in the same structure as `client_config` is, and expected to be supplied from the API users to override the config. Typically, I expect that users will copy the default JSON config file and modify it and pass it.

The returned config has to fall back to the default config, so the `config_overrides` can be a subset of `client_config`. The new `load` will merge these configs before constructing the config.

See also: https://github.com/googleapis/gax-python/pull/103

michaelbausor:
Closing out old issues. Call settings overrides are handled by the RetrySettings object.

PR #8, opened 2016-05-17
Update CallSettings::load()

jmuk:
Based on #7, we want to support the same structure for overriding
configs, so that the users can create a copy of config JSON file
and pass it.


michaelbausor:
Closing old PR

PR #10, opened 2016-05-24
Added API common types

michaelbausor:
Added API common types generated using packman
Updated composer file


composer.json
```diff
@@ -14,7 +14,13 @@
...
     "autoload": {
         "psr-4": {
-          "Google\\GAX\\":"src/"
+          "Google\\GAX\\":"src/",
+          "google\\protobuf\\":"src/generated/",
+          "google\\api\\":"src/generated/",
+          "google\\rpc\\":"src/generated/",
+          "google\\type\\":"src/generated/",
+          "google\\longrunning\\":"src/generated/",
+          "google\\logging\\":"src/generated/"
```
garrettjonesgoogle: Does this actually work with autoloading? The filenames aren't named the same as the classes as per PSR-4.

michaelbausor: You are right, this won't work for all of the classes that don't match the file names. Have switched to use classmap instead. This requires running dump-autoload when classes are added to the src/generated directory, but this should be rare.

garrettjonesgoogle:
LGTM

PR #13, opened 2016-06-07
removes ini changes in file

bshaffer:


garrettjonesgoogle:
Could you give more context?

bshaffer:
`php.ini` configuration and error reporting should be left up to the user, and not be defined in the header of a class file.

PR #18, opened 2016-07-12
update auth

dwsupplee:
We are now using auth 0.9 in [gcloud-php](https://github.com/GoogleCloudPlatform/gcloud-php) and would like to update gax to match.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,7 +8,7 @@
     "require": {
         "php": ">=5.5",
         "google/auth": "0.9",
-        "grpc/grpc": "v0.15.0"
+        "grpc/grpc": "v0.15.2"
     },
     "require-dev": {
         "phpunit/phpunit": "4.8.*"
--- a/composer.json
+++ b/composer.json
@@ -7,7 +7,7 @@
     "license": "BSD-3-Clause",
     "require": {
         "php": ">=5.5",
-        "google/auth": "0.7",
+        "google/auth": "0.9",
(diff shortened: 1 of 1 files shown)


michaelbausor:
I re-ran the travis build after Stanley added the grpc v0.15 tag. There is now a new problem, which is that grpc v0.15 requires auth 0.7, so there is a conflict.
https://github.com/grpc/grpc/blob/master/src/php/composer.json#L17

dwsupplee:
Ah, I think I understand what happened here! Sorry about this.

The v0.15.0 tag was dropped from packagist and Stanley just recently re-added it. I'm guessing that when the tag was added again it was pointing at a different commit.

@stanley-cheung Do you think we would be able to correct the v0.15.0 tag to point to the same commit as before and add a v0.16.0 which includes the auth upgrade?

stanley-cheung:
I see. That's a bit tricky. Currently the fix is only on `master`, and the newly cut `v1.0.x` branch. It's not on the `release-0_15` branch. I don't think we can wait for the 1.0 release. Let me get the change back to the `release-0_15` and perhaps make a PHP-specific 0.15.1 tag. Sounds good?

dwsupplee:
Sounds like a great solution. Thanks Stanley!

dwsupplee:
@michaelbausor pushed an update, can you check again please?

michaelbausor:
LGTM

PR #19, opened 2016-07-12
Updated PageStreaming

michaelbausor:
Added PagedListResponse, Page and FixedSizeCollection
Added tests


src/PagedListResponse.php
```diff
@@ -0,0 +1,75 @@
...
+
+class PagedListResponse
+{
+    private $parameters;
+    private $callable;
+    private $pageStreamingDescriptor;
+
+    public function __construct($params, $callable, $pageStreamingDescriptor) {
+        if (empty($params) || !is_object($params[0])) {
+            throw new InvalidArgumentException('First argument must be a request object.');
```
garrettjonesgoogle: In proto, it might be valid for a request type to be a simple type instead of a full request object (but I'm not 100% sure)

michaelbausor: I had a quick look and while every example I found uses messages, I couldn't find a definite answer. When I try creating an rpc method that takes a simple type (e.g. string or int32), I get this error:
java.lang.IllegalArgumentException: proto compilation failed with internal error: proto compilation failed: (...)  library.proto:39:26: Expected message type.

So based on this, I *think* that an object is required, but I'm not 100% sure either.

src/FixedSizeCollection.php
```diff
@@ -0,0 +1,159 @@
...
+ * collection will attempt to retrieve a fixed number of elements, and
+ * will make API calls until that fixed number is reached, or there
+ * are no more elements to retrieve.
+ */
+class FixedSizeCollection implements IteratorAggregate
+{
+    private $collectionSize;
+    private $pageList;
+
+    public function __construct($initialPage, $collectionSize) {
```
shinfan: nit: "{" should be in a separate line

michaelbausor: Would it be a good idea to add a formatting check to the Travis tests, using e.g. php-cs-fixer and/or phpcbf? Or will this just be annoying...

src/PagedListResponse.php
```diff
@@ -0,0 +1,158 @@
...
+     * elements to be retrieved.
+     *
+     * NOTE: it is an error to call this method if the optional parameter
+     * 'page_size' has not been set in the original API call. It is also
+     * an error if the collectionSize parameter is less than the
+     * page_size.
+     */
+    public function expandToFixedSizeCollection($collectionSize) {
+        if (!array_key_exists(2, $this->parameters) ||
+                !array_key_exists('page_size', $this->parameters[2])) {
```
shinfan: nit: indentation looks weird

src/Page.php
```diff
@@ -0,0 +1,169 @@
...
+        $resourceField = $this->pageStreamingDescriptor->getResourceField();
+        foreach ($this->getResponseObject()->$resourceField as $element) {
+            yield $element;
+        }
+    }
+
+    /**
+     * Return an iterator over the elements in the response.
+     */
+    public function getIterator() {
```
shinfan: It seems a bit redundant to me to have both getIterator() and iteratePageElements()

src/Page.php
```diff
@@ -0,0 +1,169 @@
...
+     */
+    public function getNextPageToken() {
+        $responsePageTokenField = $this->pageStreamingDescriptor->getResponsePageTokenField();
+        return $this->getResponseObject()->$responsePageTokenField;
+    }
+
+    /**
+     * Retrieves the next Page object using the next page token.
+     */
+    public function getNextPage($pageSize = null) {
```
shinfan: It may sound crazy but since the $pageSize is not functioning as expected, shall we just hide it from the surface? And whenever user wants a fixed page size, he should switch to use FixedSizeCollection.

src/PagedListResponse.php
```diff
@@ -31,44 +31,127 @@
...
+ *
+ * The PagedListResponse object is returned by API methods that implement
+ * pagination, and makes it easier to access multiple pages of results
+ * without having to manually manipulate page tokens. Pages are retrieved
+ * lazily, with additional API calls being made as additional results
+ * are required.
+ *
+ * The list elements can be accessed in the following ways:
+ *  - As a single iterable using the iterateAllElements method
+ *  - As pages fo elements, using the getPage and iteratePages methods
```
garrettjonesgoogle: spelling (fo)

src/PagedListResponse.php
```diff
@@ -31,44 +31,127 @@
...
+        return $this->getPage()->iteratePages();
     }
 
-    public function getFixedSizeCollection($collectionSize) {
-        return new FixedSizeCollection($this->getPage($collectionSize), $collectionSize);
+    /**
+     * Returns a collection of elements with a fixed size set by
+     * the collectionSize parameter. The collection will only contain
+     * fewer than collectionSize elements if there are no more
+     * elements to be retrieved.
```
garrettjonesgoogle: Isn't it if there are no more pages to be retrieved from the server?

src/FixedSizeCollection.php
garrettjonesgoogle: I think it would be a good idea; gcloud-php does it.

src/PagedListResponse.php
michaelbausor: Aren't the two equivalent? But maybe referring to pages is clearer

src/Page.php
michaelbausor: I am not opposed to this idea. The page_size option is available in the original API call as part of the optional parameters, so removing it from here won't lose much. And if the user really needs to retrieve a sequence of pages using different page_size options, they can manually manipulate page tokens and make API calls

src/Page.php
michaelbausor: Yes, it is a bit silly. The iteratePageElements method is there to mirror the iterate methods in the PagedListResponse, while the getIterator is there to implement the IteratorAggregator interface. We could remove the iteratePageElements method and just rely on iteration (for this and for the FixedSizeCollection)?

michaelbausor:
Addressed PR comments, PTAL

src/Page.php
michaelbausor: Removed iteratePageElements, and similarly for FixedSizeCollection

src/Page.php
michaelbausor: Removed pageSize object from the getNextPage method. The pageSize parameter can still be used as an option to the original call.

src/PagedListResponse.php
michaelbausor: Fixed

src/PagedListResponse.php
michaelbausor: Fixed

garrettjonesgoogle:
LGTM

PR #23, opened 2016-07-13
Fix flaky test testRetryTimeoutExceeds

michaelbausor:
Fix flaky test testRetryTimeoutExceeds using deterministic function to get elapsed time instead of relying on system clock.

### DIFF
--- a/tests/ApiCallableTest.php
+++ b/tests/ApiCallableTest.php
@@ -175,9 +175,17 @@ class ApiCallableTest extends PHPUnit_Framework_TestCase
             $backoffSettings);
         $callSettings = new CallSettings(['retrySettings' => $retrySettings]);
 
+        // Use time function that simulates 1100ms elapsing with each call to the stub
+        $incrementMillis = 1100;
+        $timeFuncMillis = function() use ($stub, $incrementMillis) {
+            $actualCalls = count($stub->actualCalls);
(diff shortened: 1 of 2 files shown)


michaelbausor:
@garrettjonesgoogle @shinfan PTAL

tests/ApiCallableTest.php
```diff
@@ -175,9 +175,17 @@ public function testRetryTimeoutExceeds()
             $backoffSettings);
         $callSettings = new CallSettings(['retrySettings' => $retrySettings]);
 
+        // Use time function that simulates 1100ms elapsing with each call to the stub
+        $incrementMillis = 1100;
+        $timeFuncMillis = function() use ($stub, $incrementMillis) {
```
shinfan: I haven't tried it but instead of passing a function, you could consider overriding the system timer:
http://de.php.net/manual/en/function.override-function.php

michaelbausor: Just had a look, I don't really like the look of it... It accepts the code for the new function as a string (why!?), so it can't bind to the stub, it would (I think) need to access the stub from a global variable.

shinfan: Yea, I did not notice it only accepts the code...weird...

shinfan:
LGTM

Issue #28, opened 2016-07-18
Timestamp type appears to be missing


dwsupplee:
When attempting to decode a message received from pubsub I'm getting the following error:

`PHP Fatal error:  Class '\google\protobuf\Timestamp' not found in /Users/dave/dev/gcloud-php/vendor/stanley-cheung/protobuf-php/library/DrSlump/Protobuf/Codec/Binary.php on line 267`

I looked through `src/generated` and wasn't able to find this type. 

Please let me know if there is anything I can do to help.

michaelbausor:
This fixed the issue in my tests, please let me know if you are still seeing a  problem

PR #29, opened 2016-07-19
Regenerate common APIs and well known types

michaelbausor:
This PR separates the "well known types" (which are defined by proto files in https://github.com/google/protobuf/tree/master/src/google/protobuf) from the common api types (which are defined by proto files in (https://github.com/googleapis/googleapis/tree/master/google). In the long term, the proto implementation for PHP will be responsible for generating the well known types, as happens in other languages, so we will be able to remove the well known types and leave the common api types.

Fixes https://github.com/googleapis/gax-php/issues/28


michaelbausor:
cc @dwsupplee 
This should fix the issue with timestamp, I will let you know when merged

dwsupplee:
Awesome, thanks for the quick fix!

garrettjonesgoogle:
LGTM

PR #33, opened 2016-07-27
DO NOT MERGE - test of using keyFile for credentials

michaelbausor:


michaelbausor:
Note: This is a demonstration of how a credentials keyFile can be used - this is not intended to be merged as is.

There is a hacked-together caching mechanism added as well, to see what happens when caching is used. It is only used by the keyFile credentials, because when it is used by Application Default Credentials, there is a segmentation fault when they are reused.

I have replaced the call to updateMetadata with direct use of fetchAuthToken. This should do exactly the same thing as previously, but it is clearer that the $context object is completely unused (as it was previously by updateMetadata)

michaelbausor:
@jdpedrie @dwsupplee @bshaffer 
Regarding auth and caching - is it correct to create and reuse a single CredentialsLoader object? This is what I was testing in this PR, and that seems to be how the caching implementation (https://github.com/google/google-auth-library-php/pull/123) will work, because it will wrap a CredentialsLoader or anything implementing FetchAuthTokenInterface.

I ask because while this works for a CredentialsLoader created from a key file with CredentialsLoader::makeCredentials($scopes, $jsonData), I am getting a segmentation fault in gRPC when I try to reuse the CredentialsLoader object returned from ApplicationDefaultCredentials::getCredentials($scopes). I wanted to confirm I wasn't making a mistake with the Auth library before trying to investigate what might be the cause in gRPC.

dwsupplee:
Reusing a single `CredentialsLoader` instance seems to make sense to me.

I believe the segmentation fault may be due to an issue in the auth library. It appears reusing the same `CredentialsLoader` instance with ADC for repeated calls to `fetchAuthToken` results in the following error:

```php
use Google\Auth\ApplicationDefaultCredentials;

$creds = ApplicationDefaultCredentials::getCredentials();
$creds->fetchAuthToken();
$creds->fetchAuthToken(); // exception is thrown
```

```js
{
 "error": "invalid_request",
 "error_description": "Required parameter is missing: grant_type"
}
```

This seems like something we should look into, but we should be able to circumvent the issue for now by creating a new `CredentialsLoader` instance on every call like done previously. The important part will be making sure we are only using one instance of the cache implementation.

src/GrpcBootstrap.php
```diff
@@ -32,32 +32,55 @@
...
     /**
-     * Provides a default instance of GrpcBootstrap.
+     * Accept an optional keyFile argument, which can be used to load credentials
+     * instead of using ApplicationDefaultCredentials
      */
-    public static function defaultInstance()
+    public function __construct($scopes, $options = [])
     {
-        return new GrpcBootstrap();
+        if (array_key_exists('keyFile', $options) && !is_null($options['keyFile'])) {
```
shinfan: Can we use isset($options, 'keyFile')?

src/GrpcBootstrap.php
```diff
@@ -32,32 +32,55 @@
...
     /**
      * Creates the callback function to be passed to gRPC for providing the credentials
      * for a call.
-     *
-     * @param scopes The scopes for Oauth2 credentials.
      */
-    public function createCallCredentialsCallback($scopes)
+    public function createCallCredentialsCallback()
     {
-        $authCredentials = $this->getADCCredentials($scopes);
```
shinfan: We can remove getADCCredentials if it is not used.

src/GrpcBootstrap.php
```diff
@@ -32,32 +32,55 @@
...
+    public function __construct($scopes, $options = [])
     {
-        return new GrpcBootstrap();
+        if (array_key_exists('keyFile', $options) && !is_null($options['keyFile'])) {
+            # Get credentials from the keyFile provided
+            $jsonData = json_decode(file_get_contents($options['keyFile']), true);
+            $this->authCredentials = CredentialsLoader::makeCredentials($scopes, $jsonData);
+        } else {
+            # Get the Application Default Credentials
+            $this->authCredentials = ApplicationDefaultCredentials::getCredentials($scopes);
```
shinfan: I would probably use  getADCCredentials() here so it is overridable in the unit test. or we can remove getADCCredentials since it is no longer used.

michaelbausor:
Closing this PR, changes here: https://github.com/googleapis/gax-php/pull/34

PR #37, opened 2016-08-12
Update travis

michaelbausor:
- Use stable version of grpc pecl package
- Force travis to use PHP 5.5.38


garrettjonesgoogle:
LGTM

PR #39, opened 2016-08-23
Support pageSize parameter

michaelbausor:
Add support for pageSize parameter to getPage method in Page object, as required by FixedSizeCollection

Add requestPageSizeField support for PageStreamingDescriptor

### DIFF
--- a/src/PageStreamingDescriptor.php
+++ b/src/PageStreamingDescriptor.php
@@ -39,7 +39,6 @@ use InvalidArgumentException;
 class PageStreamingDescriptor
 {
     private $requestPageTokenField;
-    private $requestHasPageSizeField;
     private $requestPageSizeField;
     private $responsePageTokenField;
     private $resourceField;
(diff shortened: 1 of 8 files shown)


src/PageStreamingDescriptor.php
```diff
@@ -55,6 +57,13 @@ public function __construct($descriptor)
     {
         self::validate($descriptor);
         $this->requestPageTokenField = $descriptor['requestPageTokenField'];
+        if (isset($descriptor['requestPageSizeField'])) {
+            $this->requestHasPageSizeField = true;
+            $this->requestPageSizeField = $descriptor['requestPageSizeField'];
+        } else {
+            $this->requestHasPageSizeField = false;
+            $this->requestPageSizeField = null;
```
garrettjonesgoogle: Couldn't you just get by with only requestPageSizeField, and distinguish its presence/absence using null vs not null?

michaelbausor: Done

michaelbausor:
PTAL

garrettjonesgoogle:
LGTM after Travis passes

PR #40, opened 2016-10-14
Update agent header

michaelbausor:
Update the agent header to x-goog-api-client

Update the metadata processing so that x-goog-api-client is used (previously, "headers" was appearing in the metadata)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: php
 php:
   - '5.5.38'
-  - '5.6'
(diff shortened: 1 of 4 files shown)


michaelbausor:
Fixed build failure with changes to travis yaml, PTAL. Also filed this issue to follow up on: https://github.com/googleapis/gax-php/issues/41

garrettjonesgoogle:
LGTM

PR #42, opened 2016-10-31
Add LRO support

michaelbausor:
This is a possible OperationResponse object for use with LRO in PHP. Some points to note:

- I'm not sure if we want to support promises, so I have commented them as optional. We can add support for them later in a non-breaking way
- Calling getResult() returns null when the operation is not complete, and throws when it has completed with an error. Maybe it should return null when completed with an error?
- A consequence of getResult() throwing is that pollUntilComplete() does not have a return value (it can't return getResult() because that might throw), which complicates the simple case


jdpedrie:
@dwsupplee and I agree that if an LRO completes with an error, it should not throw an exception. Instead, the response should be returned, and left to the implementor to deal with.

If there is an error in any of the requests, (i.e. an invalid argument or a server error), those exceptions should certainly still be thrown.

michaelbausor:
OK no worries. If we are returning either the response or error, do you prefer:
- Just return either the unpacked response, or the Status object containing the error, and let the user figure it out
- Return an array, which would look like [true, response] for success or [false, status] for error
- Return an array, which would look like [Status.code.OK, response] for success or [Status.code, status] for error

The reason to prefer either the second or third option over the first is to provide an easy way for people to check for success or failure, instead of having to interrogate the return type. Alternatively, we could provide helpers on the OperationResponse object, but this feels more awkward to me.

Thanks for having a look so quickly!

jdpedrie:
I should have been more careful in my previous comment. When I said "response", I was thinking in HTTP terms. I'd like to have the entire `Operation` available to manage as I see fit in a given situation. So I could check whether `Operation.response` or `Operation.error` were set and handle things as needed in either case. Apologies!

michaelbausor:
No worries, you are right that we should make the proto object available. I had intended to do that but missed that method :-( There is now an added method to get the operation proto object.

That still leaves the question of what to return from the getResult() method, which I think is a useful helper so that it isn't necessary to get the operation proto in most cases.

src/OperationResponse.php
```diff
@@ -0,0 +1,133 @@
+<?php
+
+// Simple case
+$op = $sampleApi->longRunningRpc();
+// ... do stuff ...
+$op->pollUntilComplete();
```
shinfan: ok I saw the implementation of the operation. nvm.

src/OperationResponse.php
```diff
@@ -0,0 +1,133 @@
...
+        return $this->operationProto->getName();
+    }
+
+    public function pollUntilComplete($pollSettings = [])
+    {
+        while (!$this->isDone()) {
+            // TODO: use poll settings
+            sleep(1);
+            $this->refresh();
+        }
```
shinfan: should this method return success/error?

jdpedrie:
@michaelbausor WDYT about having `getResult()` and `getError()` methods, both of which are null when `done` is false. Once complete, on success, `getResult()` will return the response and `getError()` will remain null. On error, vice versa.

I think explicitly choosing the result or the error makes it easier to prevent errors in mishandling an unexpected result (i.e. if you code around what a success should look like and instead get an error, will your code break?).

```php
$op = $sampleApi->longRunningRpc();
$op->pollUntilComplete();

if (!is_null($op->getError())) {
    // handle errors
}

return $op->getResult();
```

src/OperationResponse.php
michaelbausor: Very possibly! Originally it was tricky because I had getResult() throwing an exception on errors. If we go with the suggestion from @jdpedrie below, then it could return getResult(), which will be null in the case of errors. It could also return a bool indicating success or error.

michaelbausor:
I think that sounds sensible. That makes the OperationResponse object very similar in structure to the proto object, will helpers for refresh(), cancel(), pollUntilComplete() and using promises.

dwsupplee:
@michaelbausor do we have the ability to make async calls with gRPC yet? If yes, it would be great if we could set some time aside to come up with a good plan for how to handle promises. For example, I have a WIP branch right now that introduces `Async` versions of all our veneer methods that trigger network requests by utilizing [guzzle's promises library](https://github.com/guzzle/promises). If we could match a similar style in the generated code that would be really awesome.

michaelbausor:
Not right now through Veneer toolkit, but I am looking into what would be required to add that support. Definitely it would be great to pick a time to discuss how we want to handle promises more generally - how about Thursday or Friday this week some time?

dwsupplee:
Friday should work for me!

michaelbausor:
Following up re status - it seems that the latest is your open issue in gRPC (https://github.com/grpc/grpc/issues/6654), which hasn't been given a priority yet. So we aren't able to support async over gRPC at this time. But it would still be worth having the discussion so that we are on the same page. Friday sounds good, I have created an event at 10:30 PST as a placeholder, please edit or ping me if that doesn't work for you.

michaelbausor:
I have updated the proposed LRO surface, PTAL when you have a chance. It uses getResult() and getError() as @jdpedrie suggested above.

pollUntilComplete() accepts an optional callback, but I am not sure if this is adding very much value - perhaps it should be removed.

cc @bshaffer

michaelbausor:
@garrettjonesgoogle @shinfan PTAL for review before merging

src/Longrunning/OperationsClient.php
```diff
@@ -0,0 +1,481 @@
...
+ * with these names, this class includes a format method for each type of name, and additionally
+ * a parse method to extract the individual identifiers contained within names that are
+ * returned.
+ */
+class OperationsClient
+{
+    /**
+     * The default address of the service.
+     */
+    const SERVICE_ADDRESS = 'longrunning.googleapis.com';
```
garrettjonesgoogle: In Toolkit, add logic to not populate SERVICE_ADDRESS and $defaultScopes in PHP. This is configured for longrunning like this:

```
  required_constructor_params:
  - service_address
  - scopes
```

michaelbausor: Fixed, PTAL

src/Longrunning/OperationsClient.php
```diff
@@ -121,17 +117,16 @@ private static function getPageStreamingDescriptors()
      * Constructor.
      *
      * @param array $options {
-     *                       Optional. Options for configuring the service API wrapper.
+     *                       Required. Options for configuring the service API wrapper. Those options
+     *                       that must be provided are marked as Required.
      *
-     *     @type string $serviceAddress The domain name of the API remote host.
-     *                                  Default 'longrunning.googleapis.com'.
+     *     @type string $serviceAddress Required. The domain name of the API remote host.
```
garrettjonesgoogle: If this is required, should it be a proper parameter instead of an array parameter? (I'll let others decide, but I wanted to ask)

michaelbausor: I'm not fully sure, but my thinking was that it is more intuitive to update the documentation to indicate that the options array is now required, rather than changing the signature of the constructor. compared to other <X>Client objects.

src/Longrunning/OperationsClient.php
```diff
@@ -0,0 +1,478 @@
...
+use google\longrunning\DeleteOperationRequest;
+use google\longrunning\GetOperationRequest;
+use google\longrunning\ListOperationsRequest;
+use google\longrunning\OperationsGrpcClient;
+
+/**
+ * Service Description: Manages long-running operations with an API service.
+ *
+ * When an API method normally takes long time to complete, it can be designed
+ * to return [Operation][google.longrunning.Operation] to the client, and the client can use this
```
dwsupplee: I've noticed a few instances of `[Thing][namespace.Thing]` - is there special tooling in place to handle parsing this format? I don't think it is valid markdown.

src/Longrunning/OperationsClient.php
```diff
@@ -0,0 +1,478 @@
...
+     *     }
+     * } finally {
+     *     if (isset($operationsClient)) {
+     *         $operationsClient->close();
+     *     }
+     * }
+     * ```
+     *
+     * @param string $name         The name of the operation collection.
+     * @param string $filter       The standard list filter.
```
dwsupplee: Is there no way to list operations without a filter? Where would the best place to find documentation on how the filter should be formatted?

src/OperationResponse.php
```diff
@@ -0,0 +1,230 @@
...
+     *
+     * The $pollSettings optional argument can be used to control the polling loop.
+     *
+     * @param array $options {
+     *                       Options for configuring the polling behaviour.
+     *
+     *     @type float $pollingIntervalSeconds The polling interval to use, in seconds.
+     *                                          Default: 1.0
+     * }
+     * @return boolean Indicates if the operation completed successfully.
```
dwsupplee: Personally, I tend to prefer the short version of variable names in docblocks. Since the short version was used in `isDone()` above WDYT about the same here?

src/OperationResponse.php
```diff
@@ -0,0 +1,230 @@
...
+        return $this->operationName;
+    }
+
+    /**
+     * Poll the server in a loop until the operation is complete.
+     *
+     * Return true if the operation completed successfully, otherwise return
+     * false.
+     *
+     * The $pollSettings optional argument can be used to control the polling loop.
```
dwsupplee: This seems a little redundant since the description for the param is just below, but if we do keep it should this be `The $options optional argument`?

src/OperationResponse.php
```diff
@@ -0,0 +1,230 @@
...
+    public function getError()
+    {
+        if (!$this->isDone() || !$this->lastProtoResponse->hasError()) {
+            return null;
+        }
+        return $this->lastProtoResponse->getError();
+    }
+
+    /**
+     * @return \google\longrunning\Operation The last Operation object received from the server.
```
dwsupplee: Can this also be `null`?

src/OperationResponse.php
```diff
@@ -0,0 +1,230 @@
...
+    public function getOperationsClient()
+    {
+        return $this->operationsClient;
+    }
+
+    /**
+     * Starts asynchronous cancellation on a long-running operation. The server
+     * makes a best effort to cancel the operation, but success is not
+     * guaranteed. If the server doesn't support this method, it will throw an
+     * ApiException with code \google\rpc\Code::UNIMPLEMENTED. Clients can continue
```
dwsupplee: Since we are documenting in the description, would it make sense to also add an `@throws` tag?

src/OperationResponse.php
```diff
@@ -0,0 +1,230 @@
...
+     */
+    public function getResult()
+    {
+        if (!$this->isDone() || !$this->lastProtoResponse->hasResponse()) {
+            return null;
+        }
+
+        $anyResponse = $this->lastProtoResponse->getResponse();
+        if (is_null($this->operationReturnType)) {
+            return $anyResponse;
```
dwsupplee: If `$this->operationReturnType` is defined, does it share the same way to interface with the code as `$anyResponse`?

src/OperationResponse.php
```diff
@@ -0,0 +1,230 @@
...
+     *
+     * Return true if the operation completed successfully, otherwise return
+     * false.
+     *
+     * The $pollSettings optional argument can be used to control the polling loop.
+     *
+     * @param array $options {
+     *                       Options for configuring the polling behaviour.
+     *
+     *     @type float $pollingIntervalSeconds The polling interval to use, in seconds.
```
dwsupplee: Do you think it would be useful to have a max duration for the polling?

src/OperationResponse.php
michaelbausor: Yes, fixed

src/OperationResponse.php
michaelbausor: Added to all the methods which make API calls

src/Longrunning/OperationsClient.php
michaelbausor: This is an open issue for our generator: https://github.com/googleapis/toolkit/issues/331

src/Longrunning/OperationsClient.php
```diff
@@ -0,0 +1,478 @@
...
+
+/**
+ * Service Description: Manages long-running operations with an API service.
+ *
+ * When an API method normally takes long time to complete, it can be designed
+ * to return [Operation][google.longrunning.Operation] to the client, and the client can use this
+ * interface to receive the real response asynchronously by polling the
+ * operation resource, or pass the operation resource to another API (such as
+ * Google Cloud Pub/Sub API) to receive the response.  Any API service that
+ * returns long-running operations should implement the `Operations` interface
```
michaelbausor: https://github.com/googleapis/googleapis/blob/master/google/longrunning/operations.proto#L40

Perhaps there should be a reference in the proto documentation that would propagate here?

src/Longrunning/OperationsClient.php
```diff
@@ -0,0 +1,478 @@
...
+ *     $response = $operationsClient->getOperation($name);
+ * } finally {
+ *     if (isset($operationsClient)) {
+ *         $operationsClient->close();
+ *     }
+ * }
+ * ```
+ *
+ * Many parameters require resource names to be formatted in a particular way. To assist
+ * with these names, this class includes a format method for each type of name, and additionally
```
michaelbausor: Created an issue to track this: https://github.com/googleapis/toolkit/issues/883

src/Longrunning/OperationsClient.php
michaelbausor: Hmmm looks to me like this should not be required, but it is marked as such in our gapic config (https://github.com/googleapis/googleapis/blame/master/google/longrunning/longrunning_gapic.yaml#L59). @garrettjonesgoogle what do you think?

src/OperationResponse.php
michaelbausor: Added, PTAL

composer.json
```diff
@@ -17,7 +17,8 @@
     },
     "autoload": {
         "psr-4": {
-          "Google\\GAX\\":"src/"
+          "Google\\GAX\\":"src/",
+          "Google\\Longrunning\\":"src/Longrunning/"
```
dwsupplee: Using the same root namespace seems like the way to go to me

src/Longrunning/OperationsClient.php
dwsupplee: That could definitely be helpful. My initial confusion was due to the fact I thought this was referencing a PHP interface.

dwsupplee:
The updates look great, nice work!

composer.json
bshaffer: Second for using the same root namespace.

src/Longrunning/OperationsClient.php
michaelbausor: Actually, this seems like it might be a case that justifies manually editing the comments in the generated class, rather than the proto file comments. This is usually not a sustainable way to handle the autogen comments, but in this case:
- The operations interface serves a quite different purpose from other APIs, so the comments in the proto make less sense than usual when attached to the OperationsClient object
- The underlying proto is likely to be very stable
- The generated code is being kept in a different repo from other generated clients

@garrettjonesgoogle what are your thoughts on applying a manual edit to the OperationsClient documentation in this case?

composer.json
bshaffer: This can now be removed, thanks to c84d412

src/OperationResponse.php
```diff
@@ -0,0 +1,265 @@
...
+    private $operationReturnType;
+    private $metadataReturnType;
+    private $lastProtoResponse;
+    private $deleted = false;
+
+    /**
+     * OperationResponse constructor.
+     *
+     * @param string $operationName
+     * @param \Google\Longrunning\OperationsClient $operationsClient
```
bshaffer: This needs to be updated for c84d412

src/OperationResponse.php
```diff
@@ -0,0 +1,265 @@
...
+    /**
+     * @return \google\longrunning\Operation|null The last Operation object received from the server.
+     */
+    public function getLastProtoResponse()
+    {
+        return $this->lastProtoResponse;
+    }
+
+    /**
+     * @return \Google\Longrunning\OperationsClient The OperationsClient object used to make
```
bshaffer: This needs to be updated for c84d412

src/Longrunning/OperationsClient.php
```diff
@@ -0,0 +1,478 @@
...
+ * This file was generated from the file
+ * https://github.com/google/googleapis/blob/master/google/longrunning/operations.proto
+ * and updates to that file get reflected here through a refresh process.
+ *
+ * EXPERIMENTAL: this client library class has not yet been declared beta. This class may change
+ * more frequently than those which have been declared beta or 1.0, including changes which break
+ * backwards compatibility.
+ */
+
+namespace Google\GAX\Longrunning;
```
bshaffer: nit: Since this word is hyphenated, I suggest it be `LongRunning` for camelcasing.

src/Longrunning/OperationsClient.php
garrettjonesgoogle: This is a hard one... Ideally I'd like to have the original comment in the proto reworded given that this will clash with other languages too. Proto comments are written for the proto context, and writers aren't taking into account the target language context. I'm fine if you want to work around it for now, though.

src/OperationResponse.php
michaelbausor: Done

src/OperationResponse.php
michaelbausor: Done

src/Longrunning/OperationsClient.php
michaelbausor: Done

composer.json
michaelbausor: Removed

src/Longrunning/OperationsClient.php
michaelbausor: Update the documentation with a manual edit.

michaelbausor:
PTAL

PR #46, opened 2016-11-28
Fix auth version constraint

jdpedrie:
@michaelbausor, we goofed on the auth fix earlier. The version should be `^0.11` to match [`google/apiclient`](https://github.com/google/google-api-php-client/blob/master/composer.json#L10), not `0.11`.

Can you merge and tag this as v0.4.1?

Thanks


michaelbausor:
LGTM

PR #47, opened 2016-12-01
Update headers

michaelbausor:
- Fix capitalization
- Add gax version
- Change to space-separated

### DIFF
--- a/src/AgentHeaderDescriptor.php
+++ b/src/AgentHeaderDescriptor.php
@@ -84,7 +84,8 @@ class AgentHeaderDescriptor
     /**
      * Returns the version string for GAX.
      */
-    public static function getGaxVersion() {
+    public static function getGaxVersion()
+    {
         return self::GAX_VERSION;
     }
 
--- a/src/AgentHeaderDescriptor.php
+++ b/src/AgentHeaderDescriptor.php
@@ -40,6 +40,7 @@ use InvalidArgumentException;
 class AgentHeaderDescriptor
 {
     const AGENT_HEADER_KEY = 'x-goog-api-client';
+    const GAX_VERSION = '0.5.0-dev';
 
     private $clientName;
     private $clientVersion;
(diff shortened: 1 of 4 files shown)


src/AgentHeaderDescriptor.php
```diff
@@ -40,6 +40,7 @@
 class AgentHeaderDescriptor
 {
     const AGENT_HEADER_KEY = 'x-goog-api-client';
+    const GAX_VERSION = '0.5.0-dev';
```
shinfan: Probably add a TODO and created an issue here (similar to Java) that replace this hardcoded data with automatic detection.

src/AgentHeaderDescriptor.php
```diff
@@ -75,11 +76,19 @@ public function __construct($headerInfo)
...
+            "$this->codeGenName/$this->codeGenVersion gax/$this->gaxVersion ".
             "php/$this->phpVersion"]];
     }
 
+    /**
+     * Returns the version string for GAX.
+     */
+    public static function getGaxVersion()
+    {
+        return self::GAX_VERSION;
```
garrettjonesgoogle: this isn't connected to the getHeader call - how are they stitched together?

michaelbausor: The update to toolkit calls this method in the client

src/AgentHeaderDescriptor.php
michaelbausor: Added. Unfortunately there isn't an easy way to do this for PHP, because the version number is defined in the git tag and nowhere else. But we can make a release process that bumps this version

jdpedrie: Google cloud PHP has a tool to semi automate the version bump. If you like I can help you with something similar for GAX.

shinfan:
LGTM

src/AgentHeaderDescriptor.php
jdpedrie: It changes the value in the file.

michaelbausor: That would be great, I will check out what google cloud PHP uses.

PR #49, opened 2016-12-02
Bump version to 0.5.0 for release

michaelbausor:


garrettjonesgoogle:
I noticed that composer.json doesn't have this version number in it - is that standard for php packages, to not have the version number anywhere in the source tree?

michaelbausor:
Yes, my understanding is that this is normal (see e.g. [google cloud php](https://github.com/GoogleCloudPlatform/google-cloud-php/blob/master/composer.json) or [symfony](https://github.com/symfony/symfony/blob/master/composer.json) ) which don't have version numbers. The version number is fully controlled by the git tag.

garrettjonesgoogle:
Ok. LGTM.

PR #51, opened 2016-12-13
gRPC streaming response classes

michaelbausor:
This is a first pass at streaming response classes for PHP. These classes are fairly thin wrappers around the gRPC classes, providing some helper methods and translating status codes into ApiExceptions as appropriate.

See here for the gRPC streaming objects: https://github.com/grpc/grpc/tree/master/src/php/lib/Grpc

cc @dwsupplee @jdpedrie @bshaffer

### DIFF
--- a/src/ApiCallable.php
+++ b/src/ApiCallable.php
@@ -134,6 +134,44 @@ class ApiCallable
         return $inner;
     }
 
+    private static function callAndWriteData($callable, $params)
+    {
+        $initParams = [[], []];
+        if (array_key_exists(self::GRPC_CALLABLE_METADATA_INDEX, $params)) {
+            $initParams[0] = $params[self::GRPC_CALLABLE_METADATA_INDEX];
+        }
+        if (array_key_exists(self::GRPC_CALLABLE_OPTION_INDEX, $params)) {
+            $initParams[1] = $params[self::GRPC_CALLABLE_OPTION_INDEX];
+        }
+        $response = call_user_func_array($callable, $initParams);
+        $response->write($params[0]);
+        return $response;
+    }
+
+    private static function setGrpcStreaming($callable, $grpcStreamingDescriptor)
+    {
+        $inner = function () use ($callable, $grpcStreamingDescriptor) {
+            switch ($grpcStreamingDescriptor['grpcStreamingType']) {
+                case 'grpcClientStreaming':
+                    $response = ApiCallable::callAndWriteData($callable, func_get_args());
+                    $wrappedResponse = new ClientStreamingResponse($response);
+                    break;
+                case 'grpcServerStreaming':
+                    $response = call_user_func_array($callable, func_get_args());
+                    $wrappedResponse = new ServerStreamingResponse($response);
+                    break;
+                case 'grpcBidiStreaming':
+                    $response = ApiCallable::callAndWriteData($callable, func_get_args());
+                    $wrappedResponse = new BidiStreamingResponse($response);
+                    break;
+                default:
+                    throw new ValidationException('Unexpected gRPC streaming type: ' . $grpcStreamingDescriptor['grpcStreamingType']);
+            }
+            return $wrappedResponse;
+        };
+        return $inner;
+    }
+
     private static function setCustomHeader($callable, $headerDescriptor)
     {
         $inner = function () use ($callable, $headerDescriptor) {
@@ -194,6 +232,10 @@ class ApiCallable
 
         if (array_key_exists('longRunningDescriptor', $options)) {
             $apiCall = self::setLongRunnning($apiCall, $options['longRunningDescriptor']);
+        }
+
+        if (array_key_exists('grpcStreamingDescriptor', $options)) {
+            $apiCall = self::setGrpcStreaming($apiCall, $options['grpcStreamingDescriptor']);
         }
 
         if (array_key_exists('headerDescriptor', $options)) {
--- a/src/ApiCallable.php
+++ b/src/ApiCallable.php
@@ -134,26 +134,12 @@ class ApiCallable
         return $inner;
     }
 
-    private static function callAndWriteData($callable, $params)
-    {
-        $initParams = [[], []];
-        if (array_key_exists(self::GRPC_CALLABLE_METADATA_INDEX, $params)) {
-            $initParams[0] = $params[self::GRPC_CALLABLE_METADATA_INDEX];
-        }
-        if (array_key_exists(self::GRPC_CALLABLE_OPTION_INDEX, $params)) {
-            $initParams[1] = $params[self::GRPC_CALLABLE_OPTION_INDEX];
-        }
-        $response = call_user_func_array($callable, $initParams);
-        $response->write($params[0]);
-        return $response;
-    }
-
     private static function createGrpcStreamingApiCall($callable, $grpcStreamingDescriptor)
     {
         switch ($grpcStreamingDescriptor['grpcStreamingType']) {
             case 'ClientStreaming':
                 $apiCall = function () use ($callable) {
-                    $response = ApiCallable::callAndWriteData($callable, func_get_args());
+                    $response = call_user_func_array($callable, func_get_args());
                     return new ClientStreamingResponse($response);
                 };
                 break;
@@ -165,7 +151,7 @@ class ApiCallable
                 break;
             case 'BidiStreaming':
                 $apiCall = function () use ($callable) {
-                    $response = ApiCallable::callAndWriteData($callable, func_get_args());
+                    $response = call_user_func_array($callable, func_get_args());
                     return new BidiStreamingResponse($response);
                 };
                 break;
--- a/src/ApiCallable.php
+++ b/src/ApiCallable.php
@@ -134,12 +134,18 @@ class ApiCallable
         return $inner;
     }
 
+    private static function callWithoutRequest($callable, $params)
+    {
+        array_shift($params);
+        return call_user_func_array($callable, $params);
+    }
+
     private static function createGrpcStreamingApiCall($callable, $grpcStreamingDescriptor)
     {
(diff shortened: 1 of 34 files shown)


michaelbausor:
@garrettjonesgoogle @shinfan PTAL

src/ServerStreamingResponse.php
```diff
@@ -0,0 +1,92 @@
...
+    public function readAll()
+    {
+        $resourcesField = $this->resourcesField;
+        foreach ($this->call->responses() as $response) {
+            if (!is_null($resourcesField)) {
+                foreach ($response->$resourcesField() as $resource) {
+                    yield $resource;
+                }
+            } else {
+                yield $response;
```
garrettjonesgoogle: Would it be surprising to anyone to receive a mixture of the resource type and the response type?

src/ServerStreamingResponse.php
```diff
@@ -0,0 +1,92 @@
...
+            }
+        }
+        $status = $this->call->getStatus();
+        if (!($status->code == Grpc\STATUS_OK)) {
+            throw new ApiException($status->details, $status->code);
+        }
```
garrettjonesgoogle: Is there a particular reason I'm missing that the status check occurs after the loop to read the resources?

src/BidiStreamingResponse.php
```diff
@@ -0,0 +1,174 @@
...
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+namespace Google\GAX;
+
+use Grpc;
+use Google\GAX\ApiException;
+use Google\GAX\ValidationException;
+
+/**
+ * ClientStreamingResponse is the response object from a gRPC client streaming API call.
```
garrettjonesgoogle: Wrong name in the doc here

src/BidiStreamingResponse.php
```diff
@@ -0,0 +1,174 @@
...
+
+    /**
+     * Call closeWrite(), and read all responses from the server, until the streaming call is
+     * completed. Throws an ApiException if the streaming call failed.
+     *
+     * @return \Generator|mixed[]
+     * @throws ValidationException
+     * @throws ApiException
+     */
+    public function closeAndReadAll()
```
garrettjonesgoogle: This name is confusing because it doesn't make sense that you can read anything after closing. I see that it is actually closing the write specifically. Why do we need this combination - can't the use call `closeWrite()` and then call a `readAll()` function after?

src/ServerStreamingResponse.php
michaelbausor: They won't ever receive a mixture - if $resourcesField is set, it will always be set and they will get all resources. This could be restructured so that it is clearer

src/ServerStreamingResponse.php
michaelbausor: When the loop is complete, it means that either all responses have been returned, or something failed. It is necessary to check the status to determine which one.

src/BidiStreamingResponse.php
michaelbausor: Yes they can. The reason to combine them is that readAll() doesn't really make sense if you *haven't* called closeWrite(). For example with speech, if you write audio and then call read() in a loop without calling closeWrite(), the call to read() will block until the streaming call times out and it returns null. So calling read() before closeWrite() has the potential block and timeout if the server is waiting on more calls to write() before sending anything back, while calling readAll() before closeWrite() is *guaranteed* to timeout (or block forever).

Maybe rename to closeWriteAndReadAll()?

src/BidiStreamingResponse.php
michaelbausor: Fixed

src/BidiStreamingResponse.php
garrettjonesgoogle: yeah, `closeWriteAndReadAll()` would be much clearer.

michaelbausor:
Added ApiCallable tests for streaming - still need to add unit tests for new streaming classes, but this was quite a large change so may be worth reviewing now and reviewing unit tests separately

michaelbausor:
Added unit tests and renamed grpc stream classes, PTAL

michaelbausor:
Just added a change to ApiException to support metadata property. Added to this PR because it touches the streaming classes.

src/ApiCallable.php
```diff
@@ -134,6 +134,40 @@ private static function setLongRunnning($callable, $longRunningDescriptor)
...
+                $apiCall = function () use ($callable, $grpcStreamingDescriptor) {
+                    $response = call_user_func_array($callable, func_get_args());
+                    return new ServerStream($response, $grpcStreamingDescriptor);
+                };
+                break;
+            case 'BidiStreaming':
+                $apiCall = function () use ($callable, $grpcStreamingDescriptor) {
+                    $response = ApiCallable::callWithoutRequest($callable, func_get_args());
+                    return new BidiStream($response, $grpcStreamingDescriptor);
+                };
```
garrettjonesgoogle: So you have ClientStream/ServerStream/BiditStream classes, but also here you have logic outside of them that are specific to their cases. Maybe the logic could be moved inside each?

```
case 'BidiStreaming':
    return BidiStream::makeApiCall($callable, $grpcStreamingDescriptor);
```

Maybe my idea is a bad design choice because it puts the code for issue calls in multiple places - let me know what you think.

src/ApiCallable.php
```diff
@@ -189,10 +238,20 @@ public static function createApiCall($stub, $methodName, CallSettings $settings,
...
+            }
             $apiCall = self::setPageStreaming($apiCall, $options['pageStreamingDescriptor']);
         }
 
         if (array_key_exists('longRunningDescriptor', $options)) {
+            if ($isGrpcStreaming) {
+                throw new ValidationException(
+                    'grpcStreamingDescriptor not compatible with longRunningDescriptor'
+                );
+            }
```
garrettjonesgoogle: There is a lot of mixing of validation and core logic; maybe the validation could be consolidated at the top to make the core flow cleaner?

tests/BidiStreamingResponseTest.php
```diff
@@ -0,0 +1,319 @@
...
+use Google\GAX\ApiException;
+use Google\GAX\BidiStream;
+use Google\GAX\Testing\MockBidiStreamingCall;
+use Google\GAX\UnitTests\Mocks\MockPageStreamingResponse;
+use Google\GAX\UnitTests\Mocks\MockStatus;
+use google\rpc\Status;
+use Grpc;
+use PHPUnit_Framework_TestCase;
+
+class BidiStreamingResponseTest extends PHPUnit_Framework_TestCase
```
garrettjonesgoogle: `BidiStreamTest`?

tests/ClientStreamingResponseTest.php
```diff
@@ -0,0 +1,173 @@
...
+use Google\GAX\ServerStream;
+use Google\GAX\Testing\MockClientStreamingCall;
+use Google\GAX\Testing\MockServerStreamingCall;
+use Google\GAX\UnitTests\Mocks\MockPageStreamingResponse;
+use Google\GAX\UnitTests\Mocks\MockStatus;
+use google\rpc\Status;
+use Grpc;
+use PHPUnit_Framework_TestCase;
+
+class ClientStreamingResponseTest extends PHPUnit_Framework_TestCase
```
garrettjonesgoogle: `ClientStreamTest`?

tests/ServerStreamingResponseTest.php
```diff
@@ -0,0 +1,208 @@
...
+use Google\GAX\OperationResponse;
+use Google\GAX\ServerStream;
+use Google\GAX\Testing\MockServerStreamingCall;
+use Google\GAX\UnitTests\Mocks\MockPageStreamingResponse;
+use Google\GAX\UnitTests\Mocks\MockStatus;
+use google\rpc\Status;
+use Grpc;
+use PHPUnit_Framework_TestCase;
+
+class ServerStreamingResponseTest extends PHPUnit_Framework_TestCase
```
garrettjonesgoogle: `ServerStreamTest`?

src/Testing/MockBidiStreamingCall.php
```diff
@@ -0,0 +1,141 @@
...
+ */
+class MockBidiStreamingCall
+{
+    private $responses;
+    private $deserialize;
+    private $status;
+    private $writesDone = false;
+    private $receivedWrites = [];
+
+    public function __construct($responses, $deserialize = null, $status = null)
```
garrettjonesgoogle: maybe either document that `$deserialize` is a function, or name it accordingly (e.g. `$deserializeFn`)

src/ApiCallable.php
michaelbausor: No I think this is good - the Page class is already responsible for making calls, so this is reasonable

michaelbausor: Done

src/ApiCallable.php
michaelbausor: Done

tests/BidiStreamingResponseTest.php
michaelbausor: Done

tests/ClientStreamingResponseTest.php
michaelbausor: Done

tests/ServerStreamingResponseTest.php
michaelbausor: Done

src/Testing/MockBidiStreamingCall.php
michaelbausor: Added documentation

michaelbausor:
PTAL

garrettjonesgoogle:
LGTM

PR #56, opened 2017-01-12
Add OperationResponse getReturnTypeOptions

michaelbausor:
cc @jdpedrie @dwsupplee

### DIFF
--- a/src/OperationResponse.php
+++ b/src/OperationResponse.php
@@ -220,6 +220,25 @@ class OperationResponse
     }
 
     /**
+     * Get an array containing the values of 'operationReturnType' and/or 'metadataReturnType' if
+     * they are set. The array can be passed as the $options argument to the constructor when
+     * creating another OperationResponse object.
+     *
+     * @return array
+     */
+    public function getReturnTypeOptions()
+    {
+        $options = [];
+        if (isset($this->operationReturnType)) {
+            $options['operationReturnType'] = $this->operationReturnType;
+        }
+        if (isset($this->metadataReturnType)) {
+            $options['metadataReturnType'] = $this->metadataReturnType;
+        }
+        return $options;
+    }
+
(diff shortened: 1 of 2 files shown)


src/OperationResponse.php
```diff
@@ -220,6 +220,25 @@ public function getError()
...
+     * Get an array containing the values of 'operationReturnType' and/or 'metadataReturnType' if
+     * they are set. The array can be passed as the $options argument to the constructor when
+     * creating another OperationResponse object.
+     *
+     * @return array
+     */
+    public function getReturnTypeOptions()
+    {
+        $options = [];
+        if (isset($this->operationReturnType)) {
```
jdpedrie: You don't need the isset checks. The properties are declared up at the top. If they're never set, they'll just be null. I'd prefer to have the keys always present with a possible null value than to have to check if the key is there or not. :)

michaelbausor: Fair enough - updated to use null values instead of missing keys

tests/OperationResponseTest.php
```diff
@@ -116,6 +124,10 @@ public function testWithOptions()
         $this->assertNull($op->getResult());
         $this->assertNull($op->getError());
         $this->assertNull($op->getMetadata());
+        $this->assertEquals([
+            'operationReturnType' => '\google\rpc\Status',
+            'metadataReturnType' => '\google\rpc\Status',
```
garrettjonesgoogle: Having both of these set to the same thing means your test might overlook assignment mixups in the implementation; try to have your unit test inputs use distinct values if possible.

michaelbausor: Fixed

michaelbausor:
PTAL

garrettjonesgoogle:
LGTM

PR #57, opened 2017-01-14
Update to use ReceivedRequest object

michaelbausor:


src/Testing/ReceivedRequest.php
```diff
@@ -0,0 +1,58 @@
...
+    public function getFuncCall()
+    {
+        return $this->funcCall;
+    }
+
+    public function getRequestObject()
+    {
+        return $this->requestObject;
+    }
+}
```
shinfan: nit: newline EOF

PR #58, opened 2017-01-16
Add MockOperationsImpl

michaelbausor:
Required by https://github.com/googleapis/toolkit/pull/928 to test LRO methods

### DIFF
--- a/src/Testing/LongRunning/MockOperationsImpl.php
+++ b/src/Testing/LongRunning/MockOperationsImpl.php
@@ -1,0 +1,46 @@
+<?php
+/*
+ * Copyright 2016, Google Inc. All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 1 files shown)


garrettjonesgoogle:
I can't tell quite what this does, it seems magical...

michaelbausor:
I can see why it looks that way...

It inherits from the Grpc stub (https://github.com/googleapis/gax-php/blob/master/src/generated/operations.php#L830) and used the MockStubTrait (https://github.com/googleapis/gax-php/blob/master/src/Testing/MockStubTrait.php) to override _simpleRequest, meaning that the stub will just record the request object and return the prespecified response object.

This is part of the standard unit test generation, but for longrunning this will be checked in to GAX so that it can be used in the generated unit tests of other APIs that have longrunning methods.

garrettjonesgoogle:
okeydokey, LGTM

PR #60, opened 2017-02-17
Update agent header

michaelbausor:
Toolkit PR: https://github.com/googleapis/toolkit/pull/1042

### DIFF
--- a/src/AgentHeaderDescriptor.php
+++ b/src/AgentHeaderDescriptor.php
@@ -98,7 +98,7 @@ class AgentHeaderDescriptor
 
         $grpcVersion = isset($headerInfo['grpcVersion'])
             ? $headerInfo['grpcVersion']
-            : AgentHeaderDescriptor::UNKNOWN_HEADER;
+            : phpversion('grpc');
         $metricsHeaders['grpc'] = $grpcVersion;
 
         if (isset($headerInfo['additionalMetrics'])) {
--- a/src/AgentHeaderDescriptor.php
+++ b/src/AgentHeaderDescriptor.php
@@ -50,8 +50,8 @@ class AgentHeaderDescriptor
      *     Optional.
      *
      *     @type string $phpVersion the PHP version.
-     *     @type string $clientName the name of the client application.
-     *     @type string $clientVersion the version of the client application.
+     *     @type string $libName the name of the client application.
+     *     @type string $libVersion the version of the client application.
      *     @type string $codeGenName the code generator name of the client library.
      *     @type string $codeGenVersion the code generator version of the client library.
      *     @type string $gaxVersion the GAX version.
@@ -77,11 +77,11 @@ class AgentHeaderDescriptor
             : phpversion();
         $metricsHeaders['gl-php'] = $phpVersion;
 
-        if (isset($headerInfo['clientName'])) {
-            $clientVersion = isset($headerInfo['clientVersion'])
-                ? $headerInfo['clientVersion']
+        if (isset($headerInfo['libName'])) {
+            $clientVersion = isset($headerInfo['libVersion'])
+                ? $headerInfo['libVersion']
                 : AgentHeaderDescriptor::UNKNOWN_HEADER;
-            $metricsHeaders[$headerInfo['clientName']] = $clientVersion;
+            $metricsHeaders[$headerInfo['libName']] = $clientVersion;
         }
 
         if (isset($headerInfo['codeGenName'])) {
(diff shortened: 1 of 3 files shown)


michaelbausor:
Updated with appName->libName change, PTAL

src/AgentHeaderDescriptor.php
```diff
@@ -43,44 +41,87 @@ class AgentHeaderDescriptor
...
-     *     @type string $clientVersion the version of the client application.
+     *     @type string $phpVersion the PHP version.
+     *     @type string $libName the name of the client application.
+     *     @type string $libVersion the version of the client application.
      *     @type string $codeGenName the code generator name of the client library.
      *     @type string $codeGenVersion the code generator version of the client library.
      *     @type string $gaxVersion the GAX version.
-     *     @type string $phpVersion the PHP version.
+     *     @type string $grpcVersion the gRPC version.
+     *     @type string $additionalMetrics a dictionary of additional metrics for the header.
```
bjwatson: Scratch `$additionalMetrics`. The service control team wants tight control over what we pass. The only thing we should leave somewhat open is `$libName`.

src/AgentHeaderDescriptor.php
```diff
@@ -43,44 +41,87 @@ class AgentHeaderDescriptor
...
+        $metricsHeaders['gl-php'] = $phpVersion;
+
+        if (isset($headerInfo['libName'])) {
+            $clientVersion = isset($headerInfo['libVersion'])
+                ? $headerInfo['libVersion']
+                : AgentHeaderDescriptor::UNKNOWN_HEADER;
+            $metricsHeaders[$headerInfo['libName']] = $clientVersion;
+        }
+
+        if (isset($headerInfo['codeGenName'])) {
```
bjwatson: This should always be set, and should always be `gapic`. We're not worried about the use case of anyone else using GAX directly.

tests/AgentHeaderDescriptorTest.php
```diff
@@ -0,0 +1,123 @@
...
+        $agentHeaderDescriptor = new AgentHeaderDescriptor([]);
+        $header = $agentHeaderDescriptor->getHeader();
+
+        $this->assertSame($expectedHeader, $header);
+    }
+
+    public function testWithInput()
+    {
+        $expectedHeader = [AgentHeaderDescriptor::AGENT_HEADER_KEY => [
+            'gl-php/php-9.9.9 gccl/gccl-9.9.9 gapic/gapic-9.9.9 gax/gax-9.9.9 grpc/grpc-9.9.9' .
```
bjwatson: These versions are misleading. They should just be `9.9.9`. Our spec for the custom header field says these _should_ be semver (i.e. they don't have to be, but they ought to be if possible).

src/AgentHeaderDescriptor.php
michaelbausor: Done

src/AgentHeaderDescriptor.php
michaelbausor: Done. Removed codeGenName and renamed codeGenVersion to gapicVersion

tests/AgentHeaderDescriptorTest.php
michaelbausor: Done. Updated to different numbers to test assignment.

src/AgentHeaderDescriptor.php
```diff
@@ -52,11 +52,9 @@ class AgentHeaderDescriptor
      *     @type string $phpVersion the PHP version.
      *     @type string $libName the name of the client application.
      *     @type string $libVersion the version of the client application.
-     *     @type string $codeGenName the code generator name of the client library.
-     *     @type string $codeGenVersion the code generator version of the client library.
+     *     @type string $gapicVersion the code generator version of the client library.
```
bjwatson: Nit: I'd specify that this is "...the code generator version of the _GAPIC_ library."

src/AgentHeaderDescriptor.php
```diff
@@ -43,44 +41,74 @@ class AgentHeaderDescriptor
     // TODO(michaelbausor): include bumping this version in a streamlined
     // release process. Issue: https://github.com/googleapis/gax-php/issues/48
     const GAX_VERSION = '0.7.0';
+    const UNKNOWN_HEADER = '';
```
bjwatson: Let's rename this to `UNKNOWN_VERSION`. Sorry I didn't catch this in my first review.

michaelbausor: Done

src/AgentHeaderDescriptor.php
michaelbausor: Done

PR #66, opened 2017-05-18
Migrate to new protobuf implementation

michaelbausor:
Significant changes:
- Remove common types
- Protobuf fields are private, so we use getters and setters in paging descriptor instead of field access
- Repeated and map fields use different getters and setters

This PR should not be merged until changes in toolkit and google-cloud-php are also ready (see https://github.com/googleapis/toolkit/pull/1281)

@shinfan @garrettjonesgoogle If it would be useful, I can break this PR into multiple changes.


src/LongRunning/OperationsClient.php
```diff
@@ -435,7 +478,11 @@ public function cancelOperation($name, $optionalArgs = [])
      * Sample code:
      * ```
      * try {
-     *     $operationsClient = new OperationsClient();
+     *     $options = [
+     *         'serviceAddress' => 'my-service-address',
+     *         'scopes' => ['my-service-scope'],
+     *     ];
+     *     $operationsClient = new OperationsClient($options);
```
garrettjonesgoogle: LRO doesn't have a service address - isn't there special logic in toolkit for that case?

src/LongRunning/OperationsClient.php
```diff
@@ -49,11 +49,11 @@
...
-use google\longrunning\CancelOperationRequest;
-use google\longrunning\DeleteOperationRequest;
-use google\longrunning\GetOperationRequest;
-use google\longrunning\ListOperationsRequest;
-use google\longrunning\OperationsGrpcClient;
+use Google\Longrunning\CancelOperationRequest;
+use Google\Longrunning\DeleteOperationRequest;
+use Google\Longrunning\GetOperationRequest;
+use Google\Longrunning\ListOperationsRequest;
+use Google\Longrunning\OperationsGrpcClient;
```
garrettjonesgoogle: I can't remember, have we fixed-up upper-camel casing for other services? (the ideal here would be `LongRunning`)

src/LongRunning/OperationsClient.php
michaelbausor: Yes - this refresh of the longrunning client is updating the sample code to reflect that you need to pass in the serviceAddress and scopes manually if constructing an OperationsClient (usually you would not construct an OperationsClient yourself - one is created and held by API clients with longrunning operations)

This is independent of the update to the protobuf implementation - I will refresh the operations client separately.

src/LongRunning/OperationsClient.php
michaelbausor: Because these are protobuf-generated, we aren't able to fix this at the moment. Potentially we can have the PHP protoc implementation accept a namespace option, similar to C#s (https://github.com/googleapis/googleapis/blob/master/google/longrunning/operations.proto#L24)

src/LongRunning/OperationsClient.php
michaelbausor: This has now been updated separately

src/LongRunning/OperationsClient.php
michaelbausor: I have filed an issue here: https://github.com/google/protobuf/issues/3122

michaelbausor:
This PR just got a lot harder to review due to the new proto3 common types. We can discuss how to break this into sub-PRs for review.

michaelbausor:
Replaced by https://github.com/googleapis/gax-php/pull/76

PR #68, opened 2017-05-22
Handle insecure credentials

michaelbausor:
Insecure credentials are represented as null, so we need to accept a null value. See https://github.com/grpc/grpc/blob/master/src/php/ext/grpc/channel_credentials.c#L198


garrettjonesgoogle:
LGTM

PR #70, opened 2017-05-25
Include metadata in ApiException toString method

michaelbausor:
An uncaught ApiException will now also display the metadata information, which can be critical to debugging.

### DIFF
--- a/src/ApiException.php
+++ b/src/ApiException.php
@@ -63,18 +63,27 @@ class ApiException extends Exception
         return $this->metadata;
     }
 
-    // custom string representation of object
+    /**
+     * @return string
+     */
+    public function getMetadataAsString()
+    {
+        $decodedMetadata = Serializer::decodeMetadata($this->metadata);
+        return json_encode($decodedMetadata);
+    }
+
+    /**
+     * String representation of ApiException
+     * @return string
+     */
     public function __toString()
     {
         $str = __CLASS__ . ": [{$this->code}]: {$this->message}\n";
         if (isset($this->metadata) && count($this->metadata) > 0) {
-            $str .= "Metadata:\n" . json_encode($this->metadata);
-            /*
-            foreach ($this->metadata as $key => $value) {
-                $str .= "\t$key: $value\n";
-            }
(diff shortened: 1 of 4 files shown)


michaelbausor:
@dwsupplee @jdpedrie This partially addresses https://github.com/GoogleCloudPlatform/google-cloud-php/issues/509 as it relates to the autogen layer. Please let me know if you think we should take a different approach

src/ApiException.php
```diff
@@ -48,13 +49,31 @@ public function __construct($message, $code, \Exception $previous = null)
...
+            'code' => $code,
+            'status' => GrpcConstants::getStatusNameFromCode($status->code),
+            'details' => Serializer::decodeMetadata($metadata)
+        ];
+
+        $message = json_encode($messageData, JSON_PRETTY_PRINT);
+
+        $ex = new ApiException($message, $code);
+        $ex->basicMessage = $basicMessage;
+        $ex->metadata = $metadata;
```
garrettjonesgoogle: Is this a typical pattern in PHP? (Public constructor allows a subset of the fields to be provided, static factory methods set more of the fields than the constructor does)

src/GrpcConstants.php
```diff
@@ -32,22 +32,25 @@
...
 use Grpc;
+use InvalidArgumentException;
 
 /**
  * Holds constants necessary for interacting with gRPC.
  */
 class GrpcConstants
 {
     private static $statusCodeNames;
+    private static $statusCodeNamesInverted;
```
garrettjonesgoogle: Maybe it would be clearer to call this `$statusCodesToNames`

src/ApiException.php
michaelbausor: No, and I think this is not a great pattern. I have added the `metadata` and `basicMessage` fields as optional params in the constructor.

src/GrpcConstants.php
michaelbausor: Updated

garrettjonesgoogle:
LGTM

PR #72, opened 2017-05-31
Update to allow access to call objects

michaelbausor:
Supports https://github.com/googleapis/toolkit/pull/1310

### DIFF
--- a/src/Testing/MockStubTrait.php
+++ b/src/Testing/MockStubTrait.php
@@ -210,7 +210,7 @@ trait MockStubTrait
      *
      * @return ReceivedRequest[] An array of received requests
      */
-    public function getReceivedCalls()
+    public function popReceivedCalls()
     {
         $receivedFuncCallsTemp = $this->receivedFuncCalls;
         $this->receivedFuncCalls = [];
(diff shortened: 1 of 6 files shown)


src/Testing/MockStubTrait.php
```diff
@@ -72,7 +73,9 @@ public function _simpleRequest(
             throw new UnderflowException("ran out of responses");
         }
         list($response, $status) = array_shift($this->responses);
-        return new MockUnaryCall($response, $deserialize, $status);
+        $call = new MockUnaryCall($response, $deserialize, $status);
+        $this->callObjects[] = $call;
```
garrettjonesgoogle: I'm not familiar with this syntax, what does it do?

michaelbausor: It appends `$call` to the end of the array (it is preferred over array_push: http://php.net/manual/en/function.array-push.php)

src/Testing/MockStubTrait.php
```diff
@@ -217,6 +226,16 @@ public function getReceivedCallCount()
     }
 
     /**
+     * @return mixed[] The call objects created by calls to the stub
+     */
+    public function getCallObjects()
```
garrettjonesgoogle: this is a mutating call - maybe call it `popCallObjects`

michaelbausor: Updated, along with other instances of mutating calls that used `get`

garrettjonesgoogle:
LGTM

michaelbausor:
Toolkit PR: https://github.com/googleapis/toolkit/pull/1310

PR #73, opened 2017-06-12
Migrate to new proto3 implementation

michaelbausor:
This PR contains all the changes from https://github.com/googleapis/gax-php/pull/66 except for generated protobuf types. It should be significantly easier to review, PTAL.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -8,7 +8,8 @@
     "require": {
         "php": ">=5.5",
         "google/auth": "^0.11",
-        "grpc/grpc": "v1.0.*"
+        "grpc/grpc": "dev-master",
+        "google/protobuf": "dev-php-bug"
     },
     "require-dev": {
         "phpunit/phpunit": "4.8.*",
@@ -18,8 +19,13 @@
     "autoload": {
         "psr-4": {
           "Google\\GAX\\":"src/",
-          "Google\\GAX\\UnitTests\\":"tests/"
-        },
-        "classmap": ["src/generated", "src/generated/well-known-types"]
-    }
+          "Google\\GAX\\UnitTests\\":"tests/",
+          "Google\\":"src/generated/Google/",
+          "GPBMetadata\\":"src/generated/GPBMetadata/"
+        }
+    },
+    "repositories": [{
+        "type": "vcs",
+        "url": "/usr/local/google/home/michaelbausor/dev/protobuf/protobuf"
+    }]
 }
--- a/composer.json
+++ b/composer.json
@@ -8,8 +8,8 @@
     "require": {
         "php": ">=5.5",
         "google/auth": "^0.11",
-        "grpc/grpc": "dev-master",
-        "google/protobuf": "dev-php-bug"
+        "grpc/grpc": "^1.3",
+        "google/protobuf": "dev-master"
     },
     "require-dev": {
         "phpunit/phpunit": "4.8.*",
@@ -23,9 +23,5 @@
           "Google\\":"src/generated/Google/",
           "GPBMetadata\\":"src/generated/GPBMetadata/"
         }
-    },
-    "repositories": [{
-        "type": "vcs",
-        "url": "/usr/local/google/home/michaelbausor/dev/protobuf/protobuf"
-    }]
+    }
 }
--- a/src/PageStreamingDescriptor.php
+++ b/src/PageStreamingDescriptor.php
@@ -40,12 +40,37 @@ class PageStreamingDescriptor
 {
     private $descriptor;
 
+    /**
+     * @param array $descriptor {
+     *     Required.
+     *     @type string $requestPageTokenGetMethod the get method for the page token in the request object.
+     *     @type string $requestPageTokenSetMethod the set method for the page token in the request object.
+     *     @type string $responsePageTokenGetMethod the get method for the page token in the response object.
+     *     @type string resourcesGetMethod the get method for the resources in the response object.
+     *
+     *     Optional.
+     *     @type string $requestPageSizeGetMethod the get method for the page size in the request object.
+     * }
+     */
     public function __construct($descriptor)
     {
         self::validate($descriptor);
         $this->descriptor = $descriptor;
     }
 
+    /**
+     * @param array $fields {
+     *     Required.
+     *
+     *     @type string $requestPageTokenField the page token field in the request object.
+     *     @type string $responsePageTokenField the page token field in the response object.
+     *     @type string $resourceField the resource field in the response object.
+     *
+     *     Optional.
+     *     @type string $requestPageSizeField the page size field in the request object.
+     * }
+     * @return PageStreamingDescriptor
+     */
     public static function createFromFields($fields)
     {
         $requestPageToken = $fields['requestPageTokenField'];
--- a/src/PageStreamingDescriptor.php
+++ b/src/PageStreamingDescriptor.php
@@ -38,61 +38,89 @@ use InvalidArgumentException;
  */
 class PageStreamingDescriptor
 {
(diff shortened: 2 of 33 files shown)


src/Testing/MockClientStreamingCall.php
```diff
@@ -74,16 +73,18 @@ public function wait()
 
     /**
      * Save the request object, to be retrieved via getReceivedCalls()
-     * @param $request The request object
+     * @param \Google\Protobuf\Internal\Message $request The request object
```
garrettjonesgoogle: It seems like requests don't have to be this type (looking at the implementation)

src/Serializer.php
```diff
@@ -118,8 +133,172 @@ public static function decodeMetadata($metadata)
...
+        }
+        return $result;
+    }
+
+    private function encodeMessageImpl(Message $message, Descriptor $messageType)
+    {
+        $data = [];
+
+        foreach ($messageType->getField() as $field) {
+            /** @var FieldDescriptor $field */
```
garrettjonesgoogle: Does this comment affect something?

src/Serializer.php
```diff
@@ -118,8 +133,172 @@ public static function decodeMetadata($metadata)
...
+                $oneof = $messageType->getOneofDecl()[$field->getOneofIndex()];
+                $oneofName = $oneof->getName();
+                $oneofGetter = 'get' . ucfirst(self::toCamelCase($oneofName));
+                if ($message->$oneofGetter() !== $field->getName()) {
+                    continue;
+                }
+            }
+
+            if ($field->isMap()) {
+                //throw new RuntimeException("Map field not supported: $key");
```
garrettjonesgoogle: dead code? (there are other commented-out lines in this file too)

src/Serializer.php
```diff
@@ -118,8 +133,172 @@ public static function decodeMetadata($metadata)
         return $result;
     }
 
+    private function encodeElement(FieldDescriptor $field, $data)
```
garrettjonesgoogle: Why do we need to do all of this encoding/decoding work - why doesn't protobuf have this already?

src/PageStreamingDescriptor.php
```diff
@@ -38,61 +38,89 @@
...
     {
-        return $this->requestPageTokenField;
+        return 'set' . ucfirst($field);
     }
 
-    public function getRequestPageSizeField()
+    public function getRequestPageTokenGetMethod()
     {
-        return $this->requestPageSizeField;
+        return $this->descriptor['requestPageTokenGetMethod'];
```
garrettjonesgoogle: What's the advantage of having a key/value array instead of storing this data as member variables?

composer.json
```diff
@@ -18,8 +19,13 @@
...
-        "classmap": ["src/generated", "src/generated/well-known-types"]
-    }
+          "Google\\GAX\\UnitTests\\":"tests/",
+          "Google\\":"src/generated/Google/",
+          "GPBMetadata\\":"src/generated/GPBMetadata/"
+        }
+    },
+    "repositories": [{
+        "type": "vcs",
+        "url": "/usr/local/google/home/michaelbausor/dev/protobuf/protobuf"
```
garrettjonesgoogle: I assume you'll remove this before checking in the code.

src/PageStreamingDescriptor.php
michaelbausor: There is not a strong advantage, but it seemed to simplify the implementation (that may be subjective).

As a more general point, classes such as this one and RetrySettings, whose primary purpose is as value containers, should probably be replaced by arrays whenever that is sensible, so this is a partial step in that direction.

I have added additional documentation to the construction methods.

src/Serializer.php
michaelbausor: This is for encoding/decoding to and from PHP arrays, rather than a wire format, which protobuf doesn't support. This is necessary for APIs like Pubsub, Logging and Spanner with manual veneers. Previously this was being handled in google-cloud-php, but it is general functionality that I wanted to pull into GAX because it could be useful to GAPICs in future.

src/Serializer.php
michaelbausor: It doesn't affect behaviour, but it is useful for IDE autocompletion. I can remove this and other comments if necessary.

src/Serializer.php
michaelbausor: Removed dead code

src/Testing/MockClientStreamingCall.php
michaelbausor: You are correct, updated with `|mixed`.

src/Testing/SerializationTrait.php
```diff
@@ -0,0 +1,62 @@
+<?php
+/*
+ * Copyright 2016, Google Inc.
```
michaelbausor: Fixed

src/Testing/SerializationTrait.php
```diff
@@ -0,0 +1,62 @@
...
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+namespace Google\GAX\Testing;
+
+trait SerializationTrait
+{
+    protected function deserializeResponse($value, $deserialize)
```
michaelbausor: You are right - updated to `deserializeMessage`.

tests/GeneratedTestTest.php
```diff
@@ -0,0 +1,122 @@
...
+            $this->assertRepeatedFieldEquals($expected, $actual);
+        } catch (PHPUnit_Framework_ExpectationFailedException $ex) {
+            // As expected the assertion failed, silently return
+            return;
+        }
+        // The assertion did not fail, make the test fail
+        $this->fail('This test did not fail as expected');
+    }
+
+    public function successCases()
```
michaelbausor: Updated

tests/SerializerTest.php
```diff
@@ -0,0 +1,178 @@
...
+use Google\Protobuf\Struct;
+use Google\Protobuf\Value;
+use Google\Rpc\Status;
+
+/**
+ * @group core
+ */
+class SerializerTest extends \PHPUnit_Framework_TestCase
+{
+    private function backAndForth($message, $arrayStructure)
```
michaelbausor: Updated

tests/SerializerTest.php
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    public function testHttpRule()
+    {
+        $message = new HttpRule();
+        $this->backAndForth($message, [
+            'selector' => '',
+            'body' => '',
+            'additionalBindings' => [],
+        ]);
```
michaelbausor: The array is what we expect `$message` to encode to, and vice versa. I can make this more explicit

src/Serializer.php
garrettjonesgoogle: If it's useful for something, then feel free to keep it - I figured that was the case, but I'm not familiar enough with PHP development to know :-)

tests/SerializerTest.php
michaelbausor: Added comments and variables to clarify these tests.

michaelbausor:
PTAL

garrettjonesgoogle:
LGTM

PR #78, opened 2017-06-23
Bump version to 0.20.0 for release

michaelbausor:
Bumping from 0.10.0 to 0.20.0 to indicate the major nature of the change of protobuf implementations


garrettjonesgoogle:
Interesting strategy :-) I think it makes sense, LGTM.

PR #94, opened 2017-08-16
Fix DEFAULT keyword in generated protobuf class

michaelbausor:
- Add a test that instantiates all of the generated protobuf classes in src/generated
- Rename instance of invalid DEFAULT keyword

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ php:
 sudo: required
 dist: trusty
 install:
-  - pecl install grpc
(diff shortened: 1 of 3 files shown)


garrettjonesgoogle:
PHP, you are weird.

This looks reasonable to me, but since I don't have much familiarity with the things you are calling, let someone else look too.

michaelbausor:
@shinfan @neozwu Can either of you take a look? If possible I would like to merge this today. Thanks!

PR #96, opened 2017-08-19
Test protobuf c ext

michaelbausor:
cc @dwsupplee @jdpedrie Once https://github.com/GoogleCloudPlatform/google-cloud-php/pull/600 is merged we should create a similar PR for google-cloud-php

Note that this doesn't test all combinations of platforms and c extension presence/absence - this is only to be a good citizen for Travis, as the extra testing combinations are (IMHO) unlikely to find issues.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ sudo: required
 dist: trusty
 install:
   - travis_retry pecl install grpc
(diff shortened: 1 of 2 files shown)


michaelbausor:
Gentle ping on this

PR #101, opened 2017-09-14
DO NOT MERGE: Example restructuring

michaelbausor:
This is still very incomplete. I tried to partially implement the Operations transport interface and gRPC implementation, as I think once we know what we want/need to put into it we know what we need to cut from ApiCallable.

PTAL and let me know what you think!

### DIFF
--- a/src/ApiCallable.php
+++ b/src/ApiCallable.php
@@ -40,22 +40,22 @@ use InvalidArgumentException;
  */
 class ApiCallable
 {
-    const GRPC_CALLABLE_PARAM_COUNT = 3;
-    const GRPC_CALLABLE_METADATA_INDEX = 1;
-    const GRPC_CALLABLE_OPTION_INDEX = 2;
-    const GRPC_RESPONSE_STATUS_INDEX = 1;
+
+    const TRANSPORT_METHOD_PARAM_COUNT = 2;
+    const TRANSPORT_METHOD_OPTIONS_INDEX = 1;
 
     private static function setTimeout($apiCall, $timeoutMillis)
     {
         $inner = function () use ($apiCall, $timeoutMillis) {
             $params = func_get_args();
-            if (count($params) != self::GRPC_CALLABLE_PARAM_COUNT ||
-                !is_array($params[self::GRPC_CALLABLE_OPTION_INDEX])) {
+            if (count($params) != self::TRANSPORT_METHOD_PARAM_COUNT ||
+                !is_array($params[self::TRANSPORT_METHOD_OPTIONS_INDEX])
+            ) {
                 throw new InvalidArgumentException('Options argument is not found.');
             } else {
-                $timeoutMicros = $timeoutMillis * 1000;
-                $params[self::GRPC_CALLABLE_OPTION_INDEX]['timeout'] = $timeoutMicros;
-            }
+                    $timeoutMicros = $timeoutMillis * 1000;
+                    $params[self::TRANSPORT_METHOD_OPTIONS_INDEX]['timeout'] = $timeoutMicros;
+                }
             return call_user_func_array($apiCall, $params);
         };
         return $inner;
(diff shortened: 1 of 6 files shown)


michaelbausor:
(Also not that I still have things like the Grpc status codes - I just didn't get to that tonight, we will need to remove them and replace, probably with values from the Google\Rpc\Code proto).

src/Grpc/GrpcUnaryCall.php
```diff
@@ -0,0 +1,56 @@
...
+
+    public function __construct(\Grpc\UnaryCall $innerUnaryCall)
+    {
+        $this->innerUnaryCall = $innerUnaryCall;
+    }
+
+    public function wait()
+    {
+        list($response, $status) = $this->innerUnaryCall->wait();
+        if ($status->code == \Google\Rpc\Code::OK) {
```
bshaffer: 👍

src/Grpc/GrpcUnaryCall.php
```diff
@@ -0,0 +1,56 @@
...
+            return $response;
+        } else {
+            throw ApiException::createFromStdClass($status);
```
bshaffer: Here is where we would define how the exception is formatted then. I dig it.

src/LongRunning/OperationsGrpcTransport.php
```diff
@@ -0,0 +1,130 @@
...
+     *     @type \Google\Auth\CredentialsLoader $credentialsLoader
+     * }
+     *
+     * @return \Google\GAX\UnaryCall with wait() returning \Google\Longrunning\Operation
+     * @throws \Google\GAX\ApiException if the remote call fails
+     */
+    public function GetOperation(GetOperationRequest $request, $optionalArgs = [])
+    {
+        list($metadata, $options) = $this->constructGrpcArgs($optionalArgs);
+        $innerUnaryCall = $this->grpcStub->GetOperation($request, $metadata, $options);
```
bshaffer: I assume the `grpcStub` is passed in as a constructor argument or created by default, i.e.

```php
public function __construct(array $options = [])
{
    if (isset($options['operationsStub'])) {
        $this->grpcStub = $options['operationsStub'];
    } else {
        $this->grpcStub = new OperationsGrpcClient();
    }
}
```

michaelbausor: Yes - I was imagining it would be passed into a constructor defined on `GrpcTransportTrait`, so shared by all.

michaelbausor:
Re callback creation logic: it has moved here: https://github.com/googleapis/gax-php/pull/101/files#diff-7f453da75824436820c26d837b7f1303R52

So should be no user-visible BC problems (i.e. in the gapic surface).

src/Grpc/GrpcUnaryCall.php
```diff
@@ -0,0 +1,56 @@
...
+class GrpcUnaryCall implements UnaryCall
+{
+    private $innerUnaryCall;
+
+    public function __construct(\Grpc\UnaryCall $innerUnaryCall)
+    {
+        $this->innerUnaryCall = $innerUnaryCall;
+    }
+
+    public function wait()
```
bshaffer: Correct me if I'm wrong, but since [OperationsGapicClient](https://github.com/googleapis/gax-php/blob/e09f219ed50ea1f6162d1945156d2ea6af6cd1f9/src/LongRunning/Gapic/OperationsGapicClient.php#L311) expects a callable, if we want to keep the interface, then in `OperationsGapicClient::getOperation`, will need to call `$wait`:

```php
    public function getOperation($name, $optionalArgs = [])
    {
        $request = new GetOperationRequest();
        $request->setName($name);

        $mergedSettings = $this->defaultCallSettings['getOperation']->merge(
            new CallSettings($optionalArgs)
        );
        $unaryCall = ApiCallable::createApiCall(
            $this->operationsStub,
            'GetOperation',
            $mergedSettings,
            $this->descriptors['getOperation']
        );
        return $unaryCall->wait();
    }
```

We could make the unary call into a callable (i.e. implement `__invoke`), but then we lose the ability to do async later on.

src/LongRunning/OperationsGrpcTransport.php
bshaffer: I am thinking we can move the `createCredentialsCallback` function out of `OperationsGapicClient`, and handle it in the constructor of `GrpcTransportTrait`  as well.

bshaffer: Thinking about this more, we may want to remove `GrpcCredentialsHelper` and do all that logic in `GrpcTransportTrait`

src/Grpc/GrpcUnaryCall.php
michaelbausor: `piCallable::createApiCall` currently returns a `callable` - I was imagining it would still return a `callable`. This could either (a) be a callable that is used exactly as it is now, or (b) be a callable that returns a `UnaryCall` object, in which case yes we would need to call wait.

I think I prefer (a), but not 100% sure

bshaffer: I agree, I kept it a callable and then return wait on the callable in the new PR.

bshaffer: There is an interesting side-effect to this, which is that now the retry logic does not work, as an exception is not thrown until _after_ wait is called.

michaelbausor: This is a great point. This seems problematic in general for any sort of scheme to combine async calls and retries, because we have no access to the result of the initial call without calling wait() on the grpc return object.

Perhaps what is required is that the <API>TransportInterface methods still returns a GAX\UnaryCall interface object, but that we call wait inside the apicallable, much like existing code here: https://github.com/googleapis/gax-php/blob/master/src/ApiCallable.php#L143 (although the logic to convert from a status to an exception will already be handled.)

That would mean we wouldn't need to call wait() in the method implementation in the Gapic client, and then retries should function as expected.

bshaffer: It seems to me that we may be able to have a "call" object which owns a call stack, and the call stack can either be called synchronously or asynchronously. 

```php
$call = $transport->createCall($method);
$call->setCallStack(ApiCallable::createCallStack($options));
```

src/ApiCallable.php
```diff
@@ -134,63 +134,39 @@ private static function setLongRunnning($callable, $longRunningDescriptor)
...
-                call_user_func_array($callable, func_get_args())->wait();
-            if ($status->code == Grpc\STATUS_OK) {
-                return $response;
-            } else {
-                throw ApiException::createFromStdClass($status);
-            }
-        };
-    }
-
-    private static function createGrpcStreamingApiCall($callable, $grpcStreamingDescriptor)
```
bshaffer: @michaelbausor How do you see us replacing this logic? I see two options:

1. Rename "grpcStreamingDescriptor" to "streamingDescriptor" and pass this to both Rest and GRPC calls as an option using `ApiCallable`
1. Remove "grpcStreamingDescriptor" all together and have the GAPIC client return the proper call object from the transport layer

michaelbausor: I like 2. The return types on the transport interface can be `BidiStream` (or `BidiStreamInterface`), then the details of constructing the `BidiStream` object can be handled in the grpc transport implementation (same for Server and Client streams).

PR #99, opened 2017-09-11
WIP: Retry settings

michaelbausor:
WIP PR to update retry settings. Toolkit PR: https://github.com/googleapis/toolkit/pull/1552

See comments on RetrySettings class for example usage.

cc @vam-google @dwsupplee @jdpedrie @bshaffer

PR #102, opened 2017-09-14
Retry settings changes

michaelbausor:
This PR replaces https://github.com/googleapis/gax-php/pull/99. Same toolkit PR: googleapis/toolkit#1552

cc @bshaffer @vam-google @dwsupplee @jdpedrie


src/GrpcConstants.php
```diff
@@ -1,129 +0,0 @@
...
-namespace Google\GAX;
-
-use Exception;
-use Grpc;
-use InvalidArgumentException;
-
-/**
- * Holds constants necessary for interacting with gRPC.
- */
-class GrpcConstants
```
bshaffer: I really like this change.

src/ApiCallable.php
```diff
@@ -139,7 +141,7 @@ private static function createUnaryApiCall($callable)
         return function () use ($callable) {
             list($response, $status) =
                 call_user_func_array($callable, func_get_args())->wait();
-            if ($status->code == Grpc\STATUS_OK) {
+            if ($status->code == \Google\Rpc\Code::OK) {
```
bshaffer: Where is `\Google\Rpc\Code` defined?

src/RpcStatus.php
```diff
@@ -0,0 +1,121 @@
...
+    const OUT_OF_RANGE = 'OUT_OF_RANGE';
+    const UNIMPLEMENTED = 'UNIMPLEMENTED';
+    const INTERNAL = 'INTERNAL';
+    const UNAVAILABLE = 'UNAVAILABLE';
+    const DATA_LOSS = 'DATA_LOSS';
+    const UNAUTHENTICATED = 'UNAUTHENTICATED';
+
+    const UNRECOGNIZED_STATUS = 'UNRECOGNIZED_STATUS';
+
+    private static $knownStatusMap = [
```
bshaffer: For consistency shouldn't this be called `rpcStatusToCodeMap`?

src/RpcStatus.php
```diff
@@ -0,0 +1,121 @@
...
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+namespace Google\GAX;
+
+use Google\Rpc\Code;
+
+class RpcStatus
```
bshaffer: Since these constants will be used for Http/JSON as well, it may be better to call this `ApiStatus` or something along those lines...

michaelbausor: I feel like `Rpc` is also transport agnostic, and at least a little bit mirrors \Google\Rpc\Code. But I also don't object to ApiStatus either.

src/RpcStatus.php
michaelbausor: Yes, will update

src/RpcStatus.php
bshaffer: Well, since Rpc and REST are considered different paradigms, it's not precisely correct.

dwsupplee:
Just a heads up, I've set some time aside this afternoon to review. At first glance I really like the direction this is going! 👍

michaelbausor:
@dwsupplee SG!

michaelbausor:
@dwsupplee Re your concern about complexity, it is a good point. There IS a way to update these settings without knowing anything - on all of the methods, we accept an array in place of a RetrySettings object, see  the [documentation here](https://github.com/googleapis/gax-php/pull/102/files#diff-f2a433713cb7ae1163b87f090713672aR92). There is also the `with` method, which lets you change things on an existing RetrySettings object.

But currently, there is no simple way to get a RetrySettings object from scratch without setting everything. One possibility would be to create a `RetrySettings::default()` static method, which I seriously considered. The thing that stopped me was the `retryableCodes` setting: what should be a sensible default? It isn't clear to me that there is one - having it be empty could trick people into thinking they had enabled retries, when in fact nothing would be retried. It could perhaps be ['DEADLINE_EXCEEDED', 'UNAVAILABLE'] as that is a safe-ish set. Do you think this is worth adding?

dwsupplee:
> There IS a way to update these settings without knowing anything 

Good to know! I had a feeling I was missing something :). I did see that documentation, but I think I was thrown by the fact those are the optional arguments that were used there. Given that, I'm not quite as concerned about creating a RetrySettings object from scratch.

Quick question, is this the same retry process we are using for GAPIC cross language?

michaelbausor:
The collection of settings is the same across languages. But the implementation of a RetrySettings class, or something else, and how those settings are passed by the user, varies by language.

src/RetrySettings.php
```diff
@@ -32,44 +32,271 @@
...
     }
 
     /**
-     * Construct an instance.
+     * Creates a new instance of RetrySettings that updates the settings in the existing instance
+     * with the settings specified in the $settings parameter.
      *
-     * @param int[] $retryableCodes Status codes to retry
-     * @param BackoffSettings $backoffSettings Backoff settings
+     * @param array $settings {
```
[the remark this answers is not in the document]
michaelbausor: Done

michaelbausor:
PTAL

src/CallSettings.php
```diff
@@ -191,16 +145,36 @@ public function __construct($settings = [])
...
+     * @type array $userHeaders
+     *          An array of headers to be included in the request.
+     * }
+     * @return CallSettings
+     */
+    public function with(array $settings)
     {
-        return $this->timeoutMillis;
+        $existingSettings = $this->toArray();
+        return new CallSettings($settings + $existingSettings);
```
vam-google: I'm not sure how exactly this works in PHP, but won't `$existingSettings` override items (with same key) in `$settings` (because it comes second in the expression)? Is it the case? If no, then please discard this comment. If it is, Is it the desired behavior (according to the documentation of this method it is not)?

src/CallSettings.php
```diff
@@ -217,30 +191,34 @@ public function getUserHeaders()
...
+        }
+        return $arr;
+    }
+
+    private static function convertArrayFromSnakeCase($settings)
+    {
+        $camelCaseSettings = [];
+        foreach ($settings as $key => $value) {
+            $wordsKey = str_replace('_', ' ', $key);
+            $camelCaseKey = str_replace(' ', '', ucwords($wordsKey));
```
vam-google: It seems like you can just specify `'_'` as a delimiter in [ucwords](http://php.net/manual/en/function.ucwords.php), so the line 218 will become redundant.

src/CallSettings.php
```diff
@@ -217,30 +191,34 @@ public function getUserHeaders()
...
+        if (!is_null($this->getRetrySettings())) {
+            $arr['retrySettings'] = $this->getRetrySettings();
+        }
+        if (!is_null($this->getUserHeaders())) {
+            $arr['userHeaders'] = $this->getUserHeaders();
+        }
+        return $arr;
+    }
+
+    private static function convertArrayFromSnakeCase($settings)
```
vam-google: It seems like a common operation. Shouldn't it be in some utility class instead?

src/RetrySettings.php
```diff
@@ -32,44 +32,258 @@
...
+ * ```
+ * $retrySettings = new RetrySettings([
+ *     'initialRetryDelayMillis' => 100,
+ *     'retryDelayMultiplier' => 1.3,
+ *     'maxRetryDelayMillis' => 60000,
+ *     'initialRpcTimeoutMillis' => 20000,
+ *     'rpcTimeoutMultiplier' => 1.0,
+ *     'maxRpcTimeoutMillis' => 20000,
+ *     'totalTimeoutMillis' => 600000,
+ *     'retryableCodes' => [ApiStatus::DEADLINE_EXCEEDED, ApiStatus::UNAVAILABLE],
```
vam-google: I know it exists in this form in other places (in Java too), but what is the correct way to spell this word (probably none, because it does not exist in dictionaries): "retryable" or "`retriable`"? At least when I try to translate both of these in google translate from English to Russian, only the second one gets translated.

src/CallSettings.php
jdpedrie: https://secure.php.net/manual/en/language.operators.array.php

> The + operator returns the right-hand array appended to the left-hand array; for keys that exist in both arrays, the elements from the left-hand array will be used, and the matching elements from the right-hand array will be ignored.

src/CallSettings.php
bshaffer: This will work:

```php
str_replace('_', '', ucwords($key, '_'))
```

src/RetrySettings.php
bshaffer: from [this source](https://english.stackexchange.com/questions/305273/retriable-or-retryable): 

> it's impossible to say that one spelling is correct and the other is incorrect

> Retryable is the more correct spelling, triable in the sense of able to be tested, is rare. Triable has a different meaning.

Dictionary.com does not have either one, while wiktionary only has a definition for [retriable](https://en.wiktionary.org/wiki/retriable). It looks like either one is appropriate, and the more important thing is consistency.

src/CallSettings.php
michaelbausor: I would rather not expose this method publicly if possible, as it is only used here. Moving it to a different class would require it to be public.

If we find a broader use for it, we can move it into a utility class at that point.

src/RetrySettings.php
bshaffer: The results are in! And very inconclusive, but more libraries use `retryable` than `retriable`

- Ruby uses just [retryable](https://github.com/googleapis/gax-ruby/search?utf8=%E2%9C%93&q=retryable&type=)
- Python uses just [retryable](https://github.com/googleapis/gax-python/search?utf8=%E2%9C%93&q=retryable&type=)
- NodeJS uses just [retryable](https://github.com/googleapis/gax-nodejs/search?utf8=%E2%9C%93&q=retryable&type=)
- Go uses just [retriable](https://github.com/googleapis/gax-go/search?utf8=%E2%9C%93&q=retriable&type=)
- Java uses both [retryable](https://github.com/googleapis/gax-java/search?utf8=%E2%9C%93&q=retryble&type=) and [retriable](https://github.com/googleapis/gax-java/search?utf8=%E2%9C%93&q=retriable&type=)
- DotNet uses neither

michaelbausor: I will leave as-is for now.

PR #104, opened 2017-09-18
DO NOT MERGE: Another Http/json RFC

bshaffer:
Important Changes:

 - Removed `GrpcCredentialsHelper` and refactored that logic into `GrpcTransportTrait`
 - Added `GapicClientTrait` for eventually determining the default transport.
 - The GrpcTransport and GrpcClient class names are generated by GAPIC


src/GapicClientTrait.php
```diff
@@ -0,0 +1,44 @@
...
+    {
+        if (isset($options['transport'])) {
+            $validTransports = ['grpc'];
+            if (!in_array($options['transport'], $validTransports)) {
+                throw new Exception(sprintf(
+                    'Invalid transport provided: %s',
+                    $options['transport']
+                ));
+            }
+            return $options['transport'];
```
michaelbausor: This is probably also a good place to check preconditions on transport, e.g. check for `grpc` extension

src/ApiCallable.php
```diff
@@ -39,22 +39,22 @@
...
+            if (count($params) != self::TRANSPORT_METHOD_PARAM_COUNT ||
+                !is_array($params[self::TRANSPORT_METHOD_OPTIONS_INDEX])
+            ) {
                 throw new InvalidArgumentException('Options argument is not found.');
             } else {
-                $timeoutMicros = $timeoutMillis * 1000;
-                $params[self::GRPC_CALLABLE_OPTION_INDEX]['timeout'] = $timeoutMicros;
-            }
+                    $timeoutMicros = $timeoutMillis * 1000;
+                    $params[self::TRANSPORT_METHOD_OPTIONS_INDEX]['timeout'] = $timeoutMicros;
```
michaelbausor: I got this slightly wrong, this should be 'timeoutMillis', because it is going to be passed to https://github.com/googleapis/gax-php/pull/104/files#diff-b729a08173cf725b2e82dbee0d1f03e5R67

The conversion to timeoutMicros happens here: https://github.com/googleapis/gax-php/pull/104/files#diff-7f453da75824436820c26d837b7f1303R160

src/Grpc/GrpcTransportTrait.php
```diff
@@ -29,22 +29,25 @@
...
-class GrpcCredentialsHelper
+trait GrpcTransportTrait
 {
     use ValidationTrait;
 
-    private $args;
+    private $grpcStub;
+    private $credentialsCallback;
+
+    // This value must be filled in by implementing classes
```
michaelbausor: This seems to be implementing inheritance, but in a trait. Why not make this a base class instead of a trait?

src/LongRunning/Gapic/OperationsGapicClient.php
```diff
@@ -159,17 +160,8 @@ private static function getGapicVersion()
      *
      *     @type string $serviceAddress Required. The domain name of the API remote host.
      *     @type mixed $port The port on which to connect to the remote host. Default 443.
-     *     @type \Grpc\Channel $channel
```
michaelbausor: Do we want to remove all these args? Or will we still support them, but document them as only being available with grpc transport?

src/LongRunning/Gapic/OperationsGapicClient.php
```diff
@@ -240,19 +232,17 @@ public function __construct($options = [])
...
-            $createStubOptions['sslCreds'] = $options['sslCreds'];
-        }
-        $this->grpcCredentialsHelper = new GrpcCredentialsHelper($options);
-
-        $createOperationsStubFunction = function ($hostname, $opts, $channel) {
-            return new OperationsGrpcClient($hostname, $opts, $channel);
-        };
-        if (array_key_exists('createOperationsStubFunction', $options)) {
-            $createOperationsStubFunction = $options['createOperationsStubFunction'];
+        if (empty($options['createOperationsStubFunction'])) {
```
michaelbausor: We should rename this option to `createOperationsTransportFunction`

src/LongRunning/Gapic/OperationsGapicClient.php
```diff
@@ -308,10 +298,7 @@ public function getOperation($name, $optionalArgs = [])
             $this->descriptors['getOperation']
         );
 
-        return $callable(
-            $request,
-            [],
-            ['call_credentials_callback' => $this->createCredentialsCallback()]);
+        return $callable($request)->wait();
```
michaelbausor: I think the $callable will need a second arg:
`return $callable($request, [])->wait();`

src/GapicClientTrait.php
bshaffer: Good call!

src/LongRunning/Gapic/OperationsGapicClient.php
bshaffer: Then we should also rename `$operationsStub` to `$operationsTransport`

src/LongRunning/Gapic/OperationsGapicClient.php
bshaffer: I was envisioning documenting them in the GrpcTransport, but we can leave them documented here as well.

src/Grpc/GrpcTransportTrait.php
bshaffer: As it uses a constructor and protected properties, it is a better fit for a parent class than a trait.

src/ApiCallable.php
bshaffer: So it should just be this, correct?

```php
$params[self::TRANSPORT_METHOD_OPTIONS_INDEX]['timeout'] = $timeoutMillis;
```

src/Grpc/GrpcTransportTrait.php
```diff
@@ -173,7 +145,30 @@ public function createStub($generatedCreateStub, $args = [])
...
+    {
+        $metadata = [];
+        $options = [];
+        if (array_key_exists('timeoutMillis', $optionalArgs)) {
+            $options['timeout'] = $optionalArgs['timeoutMillis'] * 1000;
+        }
+        if (array_key_exists('headers', $optionalArgs)) {
+            $metadata = $optionalArgs['headers'];
+        }
+        $options['call_credentials_callback'] = $this->credentialsCallback;
```
bshaffer: This should be overridable

bshaffer:
@michaelbausor @dwsupplee I've refactored the tests and I feel like this is getting pretty close. PTAL!

src/ApiCallable.php
```diff
@@ -206,18 +180,24 @@ private static function setCustomHeader($callable, $headerDescriptor, $userHeade
...
-            );
-        } else {
-            $apiCall = ApiCallable::createUnaryApiCall($callable);
+        $apiCall = function() use ($transport, $methodName) {
+            $callable = [$transport, $methodName];
+            return call_user_func_array($callable, func_get_args());
+        };
+
+        // Call the sync method "wait" if this is not a gRPC call
+        if (!array_key_exists('grpcStreamingDescriptor', $options)) {
```
bshaffer: I opted to use the `wait` function only when the gRPC streaming descriptors aren't set. We could make this generic, but they _are_ gRPC specific, so I'm not overly concerned with it. I am definitely open to better ways to go about this!

src/Grpc/GrpcTransportTrait.php
```diff
@@ -173,7 +145,46 @@ public function createStub($generatedCreateStub, $args = [])
         if (isset($args['forceNewChannel']) && $args['forceNewChannel']) {
             $stubOpts['force_new'] = true;
         }
-        return $generatedCreateStub($fullAddress, $stubOpts, $channel);
+
+        if (empty($args['createGrpcStubFunction'])) {
```
bshaffer: I added this to help with unit tests, but I am not sold on it. Also the whole `$grpcStubClassName` is pretty weak. Also it may be best to make this an abstract base class rather than a trait.

src/ApiCallable.php
bshaffer: **Note, this has been refactored out in #109 and looks way better**

bshaffer:
This is now tracked in the branch `json-transport` and #113

PR #109, opened 2017-09-22
Middleware RFC

bshaffer:
Paying special attention to 0d8dbf8, I've added middleware functionality. PTAL @michaelbausor and @dwsupplee


bshaffer:
Also see cabc2ac for removing `ApiCallable` all together 👍  I'm really liking this, I hope you agree!

dwsupplee:
👏 this is looking awesome. Will have some time to do a full review soon!

src/CallStackTrait.php
```diff
@@ -0,0 +1,94 @@
...
+     *     Optional.
+     *     @type \Google\GAX\PageStreamingDescriptor $pageStreamingDescriptor
+     *           the descriptor used for page-streaming.
+     *     @type \Google\GAX\AgentHeaderDescriptor $headerDescriptor
+     *           the descriptor used for creating GAPIC header.
+     * }
+     *
+     * @return callable
+     */
+    private static function createCallStack(
```
michaelbausor: Nit: what about a function name with `wrap` as the verb, rather than `create`?

src/Grpc/GrpcClientStream.php
```diff
@@ -84,12 +70,7 @@ public function write($request)
      */
     public function readResponse()
     {
-        list($response, $status) = $this->call->wait();
-        if ($status->code == Grpc\STATUS_OK) {
-            return $response;
-        } else {
-            throw ApiException::createFromStdClass($status);
-        }
+        return $this->call->wait();
```
michaelbausor: This is a bit academic since there are no examples of client streaming in any APIs, but I don't think this is correct. The `$this->call` object is still a [ClientStreamingCall](https://github.com/grpc/grpc-php/blob/master/src/lib/ClientStreamingCall.php) object, so `wait` will still return `[$response, $status]`.  We should reuse the implementation of checking the status and throwing rather than repeating it though.

src/Grpc/GrpcUnaryCall.php
```diff
@@ -0,0 +1,56 @@
+<?php
+/*
+ * Copyright 2016, Google Inc.
```
michaelbausor: Nit: 2017

src/Testing/MockUnaryCall.php
```diff
@@ -72,7 +74,9 @@ public function __construct($response, $deserialize = null, $status = null)
      */
     public function wait()
     {
-        $obj = $this->deserializeMessage($this->response, $this->deserialize);
-        return [$obj, $this->status];
+        if ($this->status->code != \Google\Rpc\Code::OK) {
```
michaelbausor: This change makes sense so long as this class is going to be used to mock the response from the TransportInterface. If it is going to be used to mock the response from the gRPC stub (which is how it is used currently) then I don't think you need this change.

tests/CallStackTraitTest.php
```diff
@@ -56,7 +58,7 @@
 use PHPUnit_Framework_TestCase;
 use Grpc;
 
-class ApiCallableTest extends PHPUnit_Framework_TestCase
+class CallStackTraitTest extends PHPUnit_Framework_TestCase
```
michaelbausor: This may already be covered by the tests you have here, but it would be good to test using a "real" grpc transport object together with a mock grpc stub. I put real in quotes because it may not be that meaningful to call a transport "real" or "mock" if it is created specifically for testing. But what I mean is that we should test with a transport object that is as close as possible to the expected grpc transport class that will be generated, but with only the grpc stub mocked out.

src/Grpc/GrpcClientStream.php
bshaffer: I think this happened because I was mocking the call wrong. Will look into it and fix.

src/LongRunning/OperationsGrpcTransport.php
```diff
@@ -0,0 +1,146 @@
...
+use Google\GAX\PageStreamingDescriptor;
+use Google\GAX\ValidationException;
+use Google\Longrunning\CancelOperationRequest;
+use Google\Longrunning\DeleteOperationRequest;
+use Google\Longrunning\GetOperationRequest;
+use Google\Longrunning\ListOperationsRequest;
+use Google\Longrunning\ListOperationsResponse;
+use Google\Longrunning\OperationsGrpcClient;
+
+
```
dwsupplee: Extra whitespace here

src/LongRunning/OperationsTransportInterface.php
```diff
@@ -0,0 +1,122 @@
...
+use Google\GAX\PageStreamingDescriptor;
+use Google\GAX\ValidationException;
+use Google\Longrunning\CancelOperationRequest;
+use Google\Longrunning\DeleteOperationRequest;
+use Google\Longrunning\GetOperationRequest;
+use Google\Longrunning\ListOperationsRequest;
+use Google\Longrunning\ListOperationsResponse;
+use Google\Longrunning\OperationsGrpcClient;
+
+
```
dwsupplee: Extra whitespace here

src/LongRunning/OperationsGrpcTransport.php
```diff
@@ -0,0 +1,146 @@
...
+{
+    use GrpcTransportTrait;
+
+    private static $grpcStubClassName = 'Google\LongRunning\OperationsGrpcClient';
+
+    /**
+     * @param GetOperationRequest $request
+     * @param array  $optionalArgs {
+     *                             Optional.
+     *     @type integer $timeoutMillis
```
dwsupplee: could we use int here and in the other definitions?

src/CallStackTrait.php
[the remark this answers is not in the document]
bshaffer: I'm interested in this, tell me more! I assume this would mean incorporating a `CallStack` class which would allow splicing/incerting/removing, as Guzzle does it?

src/ClientStreamInterface.php
```diff
@@ -0,0 +1,59 @@
...
+    public function readResponse();
+
+    /**
+     * Write all data and read the response from the server, completing the streaming
+     * call.
+     *
+     * @param mixed[] $requests An iterator of request objects to write to the server
+     * @return mixed The response object from the server
+     */
+    public function writeAllAndReadResponse($requests);
```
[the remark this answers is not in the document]
bshaffer: I do not really want to make this an optional parameter, since if they wanted to just read instead of write, they could call `readResponse`. Also it's `mixed` so we can't typehint!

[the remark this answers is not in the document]
dwsupplee: That's fair! Would it then make sense to update `BidiStreamInterface::writeAll()` to match, or is this a reason it needs to have the optional argument that I have missed?

Doesn't this take an array of mixed types? What other types than an array can be passed in?

src/CallStackTrait.php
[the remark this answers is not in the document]
dwsupplee: Yes, something similar to that would be reasonable :). I wouldn't necessarily say this is something we need to support at the moment, but if it is a feature we would like to add (I think it could add some nice value) it may be worth considering how to allow for it in the future as we touch these files now.

bshaffer:
This is now tracked in the branch `json-transport` and #113

PR #111, opened 2017-09-27
make credentials faster

pongad:
The google/auth package provides a general-purpose caching
with FetchAuthTokenCache. The implementation is aimed to be compatible
with PSR-6.

This commit introduces a faster cache, usable in conjuction with
FetchAuthTokenCache.
It queries for the last received token and reuses it if the token has
not expired.
In this way, we avoid sanitizing cache keys and performing table
lookups.

In my testing, this increases our throughput by ~10%.

### DIFF
--- a/src/GrpcCredentialsHelper.php
+++ b/src/GrpcCredentialsHelper.php
@@ -130,7 +130,13 @@ class GrpcCredentialsHelper
     {
         $credentialsLoader = $this->args['credentialsLoader'];
         $callback = function () use ($credentialsLoader) {
-            $token = $credentialsLoader->fetchAuthToken();
+            $token = null;
+            if ($this->args['enableCaching']) {
+              $token = $credentialsLoader->getLastReceivedToken();
+            }
+            if (is_null($token) || time() > $token['expires_at']) {
+              $token = $credentialsLoader->fetchAuthToken();
+            }
             return ['authorization' => array('Bearer ' . $token['access_token'])];
         };
         return $callback;
--- a/src/GrpcCredentialsHelper.php
+++ b/src/GrpcCredentialsHelper.php
@@ -132,10 +132,10 @@ class GrpcCredentialsHelper
         $callback = function () use ($credentialsLoader) {
             $token = null;
             if ($this->args['enableCaching']) {
-              $token = $credentialsLoader->getLastReceivedToken();
+                $token = $credentialsLoader->getLastReceivedToken();
             }
             if (is_null($token) || time() > $token['expires_at']) {
-              $token = $credentialsLoader->fetchAuthToken();
+                $token = $credentialsLoader->fetchAuthToken();
             }
             return ['authorization' => array('Bearer ' . $token['access_token'])];
         };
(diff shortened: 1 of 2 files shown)


pongad:
@bshaffer A problem with the current implementation is that this level of cache does not honor the FetchAuthTokenCache's lifetime config. AFAICT, items in FATC expires `lifetime` seconds into the future, while this reuse-last-cache expires whenever the token reports itself to be expired.

I don't think this will be a big problem in practice, but it sounds a little scary. Maybe we use gate this behind a different config? WDYT?

michaelbausor:
These changes seem fine, and a 10% increase in throughput is pretty huge. But what is the justification for making these changes here, instead of improving the caching in the auth library? It is very surprising to me that we can get that much of a speedup just by avoiding hitting the existing caching solution, and suggests that that might be a better place for the changes. Does that make sense?

pongad:
@michaelbausor Your thoughts absolutely makes sense. At a high level, I think it's the same reason that your CPU has multi-layered cache.

More specifically, the existing caching solution is more general, and the interface allows for an implementation where multiple PHP processes share the cache.
This is immensely useful. PHP processes are short-lived, serving only one request, so it'd be inefficient for every process to fetch credentials over network.

There is a cost to this generality though. To conform to standard and avoid surprises, the implementation needs to deal with long keys, multiple key-value pairs, etc. To continue with the CPU analogy, this implementation is like an L2 cache.

The new layer implemented in this PR doesn't conform to the standard and only caches within-process. Instead, it uses this flexibility to go faster: a bit like an L1 cache.

The existing implementation isn't "bad". However, if you need to look things up in a map, it's just going to be slower than accessing a variable.

michaelbausor:
@pongad that does make sense - that was a good explanation :-) 

I think that this change is something we should do, but that we should also file an issue, either here or in the auth library, to do a larger review/rework. I had a quick look at the way this is done in gax, and I think that we can make it a lot better. For example, the `enableCaching` flag is used to wrap a `FetchAuthTokenInterface` in a cache, but provides no way to use any PSR caching implementation other than the default `MemoryCacheItemPool`. So someone who wanted to use caching across requests would need to set `enableCaching` to false, and pass in a `FetchAuthTokenCache` object they had instantiated themselves.

It also seems to me that we want the behaviour you have implemented to always apply, regardless of whether or not the `enableCaching` flag is set. Is there ever anyone who wants to make network requests to retrieve a token when they already have a valid token? I don't think so, and I don't think we need/want to provide any way for users to do that.

@bshaffer has a big PR open at the moment that definitely affects this class, and many others in Gax. @bshaffer -does it make sense to merge this now? Or to wait, and refactor after your PR goes in?

bshaffer:
@pongad @michaelbausor The big PR (#118) has now been merged, so to answer your question super late, we should now refactor this PR.

michaelbausor:
Please see https://github.com/googleapis/gax-php/pull/145

michaelbausor:
Closing this out, as it is implemented in #145

PR #113, opened 2017-10-11
Removed the generated transport interfaces and classes

bshaffer:
Makes GRPC Transport generic, removes TransportInterface and implementing classes.

cc @dwsupplee @michaelbausor @vam-google


dwsupplee:
👍 looking nice!

PR #114, opened 2017-10-24
Revert to using ucwords() without the delimiter parameter

vam-google:
To support PHP5.5 prior 5.5.16, in which ucwords() does not have the second parameter.


vam-google:
@dwsupplee, @jdpedrie please take a look.

PR #117, opened 2017-10-30
adds call class constants

bshaffer:
Is this an improvement?


dwsupplee:
It does seem cleaner to have the strings as constants, but also, what about using [class resolution via ::class](http://php.net/manual/en/migration55.new-features.php#migration55.new-features.class-name)?

```php
$callClass = GrpcUnaryCall::class;
```

PR #118, opened 2017-10-31
Add JSON Transport

bshaffer:


michaelbausor:
I made some comments. I will hopefully have some time tomorrow, and I was not sure whereabouts in the PR it would be most productive for me to review - if you want me to look at anything in particular, please ping me and I will try to take a look!

src/ApiCore/GrpcTransport.php
```diff
@@ -0,0 +1,242 @@
...
+            default:
+                throw new ValidationException("Unexpected gRPC streaming type: $streamingType");
+        }
+    }
+
+    private function getOptions(CallSettings $settings)
+    {
+        $retrySettings = $settings->getRetrySettings();
+        $options = $settings->getGrpcOptions() ?: []
+            + ['call_credentials_callback' => $this->credentialsCallback];
```
michaelbausor: Add brackets around `$settings->getGrpcOptions() ?: []` for clarity, even if that is not strictly necessary. Otherwise, it is unclear whether this is always adding `[]+ ['call_credentials_callback' => $this->credentialsCallback]`, or adding `['call_credentials_callback' => $this->credentialsCallback]` to the result of the expression.

src/ApiCore/CallSettings.php
```diff
@@ -222,14 +244,12 @@ public function merge(CallSettings $otherSettings = null)
...
-        }
-        if (!is_null($this->getUserHeaders())) {
-            $arr['userHeaders'] = $this->getUserHeaders();
-        }
-        return $arr;
+        return array_filter([
+            'retrySettings' => $this->getRetrySettings(),
+            'userHeaders' => $this->getUserHeaders(),
+            'grpcOptions' => $this->getGrpcOptions(),
+            'grpcOptions' => $this->getRestOptions()
```
bshaffer: ITS A BUG

src/ApiCore/CallSettings.php
```diff
@@ -149,33 +151,37 @@ private static function constructRetry(
...
-     *          Retry settings to use for this method.
-     *    @type array $userHeaders
-     *          An array of headers to be included in the request.
+     *     Optional.
+     *
+     *     @type \Google\ApiCore\RetrySettings $retrySettings
+     *           Retry settings to use for this method.
+     *     @type array $userHeaders
+     *           An array of headers to be included in the request.
+     *     @type array $grpcOptions
```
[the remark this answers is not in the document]
bshaffer: Would this be an improvement? 

```php
$callSettings = new CallSettings([
    'transportOptions' => [
         'rest' => [ $guzzleOption => $guzzleVal ],
         'grpc' => [ $grpcOption => $grpcVal ],
   ]
]);
```

And then in the transport classes:

```php
$transportOptions = $callSettings->getTransportOptions();
$options = $transportOptions['grpc'];
```

So transports handle their own settings, and the `CallSettings` object does not have anything transport-specific in it. The user would then interact with it like this:

```php
$client->requestThing($request, [
    'retrySettings' => [],
    'transportOptions' => [
         'grpc' => []
    ]
]);
```

What do you think?

src/ApiCore/ApiTransportTrait.php
```diff
@@ -0,0 +1,141 @@
...
+use Google\Auth\FetchAuthTokenCache;
+use Google\Auth\HttpHandler\HttpHandlerFactory;
+
+trait ApiTransportTrait
+{
+    use ArrayTrait;
+
+    private $agentHeaderDescriptor;
+
+    private function setCommonDefaults(array $options)
```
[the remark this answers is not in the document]
dwsupplee: I'm planning to rework this a bit.

src/ApiCore/ApiTransportTrait.php
```diff
@@ -0,0 +1,141 @@
...
+    {
+        $options += [
+            'enableCaching' => true,
+            'authCache' => new MemoryCacheItemPool(),
```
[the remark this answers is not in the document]
bshaffer: There's no constructor logic for `MemoryCacheItemPool` so instantiating should be pretty lightweight. But I'd be ok moving it for best practice.

src/ApiCore/Call.php
```diff
@@ -30,35 +30,50 @@
...
-use Google\Rpc\Code;
+use Google\Protobuf\Internal\Message;
 
-/**
- * Class MockDeadlineExceededUnaryCall simulates a unary call returning DEADLINE_EXCEEDED.
- *
- * If $timeoutMicros is set, the call to wait() will sleep before returning.
- */
-class MockDeadlineExceededUnaryCall
+class Call
```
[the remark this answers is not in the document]
bshaffer: The decision to make it a class was based off of the typehinting we're doing for Pages, callables, etc, which makes it a lot clearer what's going on.

src/ApiCore/CallSettings.php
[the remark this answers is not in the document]
michaelbausor: That looks good!

src/ApiCore/LongRunning/Gapic/OperationsGapicClient.php
```diff
@@ -243,19 +247,21 @@ public function __construct($options = [])
...
 
-        $createStubOptions = [];
-        if (array_key_exists('sslCreds', $options)) {
-            $createStubOptions['sslCreds'] = $options['sslCreds'];
+        if (empty($options['createTransportFunction'])) {
+            $options['createTransportFunction'] = function ($options, $transport = null) {
+                switch ($transport) {
+                    case 'grpc':
+                        if (empty($options['createGrpcStubFunction'])) {
+                            $options['createGrpcStubFunction'] = function ($fullAddress, $stubOpts, $channel) {
```
michaelbausor: If we move setting the default value of `$options['createGrpcStubFunction']` up to where the rest of the default options are set, then I think all of the rest of this function (including the switch case) can be pushed into the call to `$this->getTransport($options)`. Does that work?

dwsupplee: This is outdated, and will be updated once I have regenerated the GAPICs.

src/ApiCore/GapicClientTrait.php
```diff
@@ -0,0 +1,157 @@
...
+
+use Exception;
+use InvalidArgumentException;
+use Google\ApiCore\LongRunning\OperationsClient;
+use Google\Cloud\Version;
+
+/**
+ * Common functions used to work with various clients.
+ */
+trait GapicClientTrait
```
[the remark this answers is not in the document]
dwsupplee: Traits lend themselves to a more compositional approach, which I tend to prefer over inheritance. Given our use case, I think either approach would end up working well - so I think it just boils down to a code style preference.

src/ApiCore/LongRunning/Gapic/OperationsGapicClient.php
bshaffer: we have removed this all together. We're just waiting on the generation of the GAPIC clients to remove this.

composer.json
```diff
@@ -23,7 +24,7 @@
     },
     "autoload-dev": {
         "psr-4": {
-            "Google\\": "tests"
+            "Google\\ApiCore\\UnitTests\\": "tests"
```
michaelbausor: As part of the most recent restructuring (moving to ApiCore) I switched the test directory structure/namespace to use `Tests/Unit`, to mirror what we use in google-cloud-php. Is the change back to using `UnitTests` intentional? I am happy to go back to this if we think it is better.

src/ApiCore/ApiStatus.php
```diff
@@ -114,4 +114,16 @@ public static function statusFromRpcCode($code)
...
+    /**
+     * @param string $status
+     * @return int
+     */
+    public static function rpcCodeFromStatus($status)
+    {
+        if (array_key_exists($status, self::$apiStatusToCodeMap)) {
+            return self::$apiStatusToCodeMap[$status];
+        }
+        return Code::UNKNOWN;
```
michaelbausor: I am not sure that this should return UNKNOWN. That is a specific error code that corresponds to a HTTP 500 response (see https://github.com/googleapis/googleapis/blob/master/google/rpc/code.proto#L51). I would be tempted to either return `-1` or throw an exception.

src/ApiCore/ApiTransportTrait.php
```diff
@@ -0,0 +1,144 @@
...
+        $retrySettings = $settings->getRetrySettings();
+        if ($retrySettings && $retrySettings->retriesEnabled()) {
+            $callable = new Middleware\RetryMiddleware($callable);
+        }
+
+        return $callable;
+    }
+
+    /**
+     * Gets credentials from ADC. This exists to allow overriding in unit tests.
```
michaelbausor: Is this comment still true? If not - can this method be removed completely?

src/ApiCore/ArrayTrait.php
```diff
@@ -0,0 +1,113 @@
...
+     * @param array $arr
+     * @param bool $isRequired
+     * @return string|null
+     * @throws \InvalidArgumentException
+     */
+    private function pluck($key, array &$arr, $isRequired = true)
+    {
+        if (!array_key_exists($key, $arr)) {
+            if ($isRequired) {
+                throw new \InvalidArgumentException(
```
michaelbausor: Should this be `ValidationException`?

In general, should all exceptions we throw be ones that we have defined, either `ValidationException` or `ApiException`? This seems like a question of what is standard practice in PHP, so I am happy to stick with `InvalidArgumentException` if that is the standard.

src/ApiCore/BidiStream.php
```diff
@@ -1,6 +1,6 @@
 <?php
 /*
- * Copyright 2016, Google Inc.
+ * Copyright 2017, Google Inc.
```
michaelbausor: I don't think we should update copyright years on existing source files, if all we do if modify them. This can be tricky because of the way git works - sometimes it thinks a new file is a modified old file, even though they are completely different. But in general if we make changes to an existing file, the copyright should not change.

src/ApiCore/ArrayTrait.php
dwsupplee: I've seen projects that like to extend `InvalidArgumentException` under their own namespace, and others which just use the original SPL exception. I would say the standard exists at the codebase level.

For a little more detail, this trait comes from [google-cloud-php](https://github.com/GoogleCloudPlatform/google-cloud-php/blob/master/src/Core/ArrayTrait.php). I ported it here as I found a few use cases where it helped to keep the code a bit more succinct. The plan is to remove the guts of the code in google-cloud-php and just extend it from here. Given that, it would be nice to retain `InvalidArgumentException` for google-cloud-php so there are no breaking changes - but we could just convert the `ValidationException` over in google-cloud-php as a way retain BC.

Personally, I like the granularity of using the standard SPL exceptions instead of just a blanket `ValidationException` for anything that goes wrong - but would find it acceptable either way. Which do you prefer?

composer.json
dwsupplee: I like `Tests\Unit`, I must have goofed with `UnitTests` on the rebase. Apologies.

src/ApiCore/ApiTransportTrait.php
dwsupplee: It should still hold true, although the current state of the tests may not currently reflect it.

src/ApiCore/ApiStatus.php
dwsupplee: I like `-1` for parity with `statusFromRpcCode`

src/ApiCore/RestTransport.php
```diff
@@ -0,0 +1,171 @@
...
+                    $call->getMethod(),
+                    $call->getMessage(),
+                    $settings->getUserHeaders()
+                ),
+                $this->getOptions($settings)
+            )->then(
+                function (ResponseInterface $response) use ($call) {
+                    $decodeType = $call->getDecodeType();
+
+                    return (new Serializer)
```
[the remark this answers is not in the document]
dwsupplee: Definitely just missed the fact the new protos had method. Should we just remove the Serializer?

src/ApiCore/ArrayTrait.php
michaelbausor: I am happy to stick with `InvalidArgumentException`

src/ApiCore/RestTransport.php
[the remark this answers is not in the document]
michaelbausor: We need the serializer for pubsub and logging, when converting between manual-layer json args and protobuf, which uses slightly different capitalization and callable hooks to simplify particular types. But for this use case when the json is for wire transmission only, we should use the protobuf methods.

src/ApiCore/ApiTransportInterface.php
```diff
@@ -47,16 +47,28 @@ public function startCall(Call $call, CallSettings $settings);
     /**
      * @param Call $call
      * @param CallSettings $settings The call settings to use for this call.
-     * @param string $streamingType
-     * @param string $resourcesGetMethod
+     * @param array $descriptor
+     *
+     * @return PromiseInterface
+     */
+    public function startOperationsCall(Call $call, CallSettings $settings, array $descriptor);
```
michaelbausor: Why do we want `startOperationsCall` and `getPagedListResponse` to be in the transport interface? I would expect their implementations to be the same between transports (and in fact that is the case, because they are implemented in the shared `ApiTransportTrait`).

dwsupplee: The thought is that if a user would like to implement their own transport, they know exactly which methods will be required in order to do so.

michaelbausor: Right. But, they **don't** need to implement those methods, because those methods are the same between transports. So if those methods are elsewhere in a transport-agnostic place, then they will be called correctly regardless of transport, right?

composer.json
bshaffer: Wouldn't this structure make more sense:

```
tests/Unit/ApiCore/LongRunning/OperationsClientTest.php
tests/Unit/ApiCore/Middleware/RetryMiddlewareTest.php
tests/Unit/ApiCore/testdata/...
tests/Unit/ApiCore/*Test.php
```

```json
"autoload-dev": {
    "psr-4": {
        "Google\\Tests\\": "tests"
    }
}
```

with the namespaces then being  `Google\Tests\Unit\ApiCore` and `Google\Tests\Unit\ApiCore\LongRunning`? This would be more consistent with `google-cloud-php`

michaelbausor: Would that be more consistent? Google Cloud PHP puts tests under the `Google\Cloud` namespace, so I would have thought the corresponding namespace would be `Google\ApiCore`. (See here: https://github.com/GoogleCloudPlatform/google-cloud-php/blob/master/tests/unit/Datastore/BlobTest.php#L18)

bshaffer: Ahh I see, yes so we would want `Google\ApiCore\Tests\Unit`. Hmm, this is okay but then why do we have `Google\ApiCore\LongRunning\Tests\Unit` instead of `Google\ApiCore\Tests\Unit\\LongRunning` ?

michaelbausor: Good point. I think we should change to `Google\ApiCore\Tests\Unit\LongRunning`

src/ApiCore/CallSettings.php
```diff
@@ -41,8 +41,8 @@ class CallSettings
 
     private $retrySettings;
     private $userHeaders;
-    private $grpcOptions;
-    private $restOptions;
+    private $transportOptions;
+    private $timeoutMillis;
```
michaelbausor: Looking at the changes, I am not sure what adding `$timeoutMillis` here is gaining us.

I understand that we want to provide an easy way for users to set the timeout in the call, so that is an argument for adding a `timeoutMillis` optional arg to the gapic methods. But once we have that, why don't we just take that user-provided arg and use it to set the `initialRpcTimoutMillis` and `noRetriesTimeoutMillis`, similar to what we were already doing in the `configureCallSettings()` method? That means we don't need to extra field in CallSettings and don't need the extra logic in the Retry middleware.

bshaffer: The main reason is that the transport layers shouldn't need to look into retry settings. There is duplicate logic in handling the timeout in the transport layers.  If we use `initialRpcTimoutMillis` and `noRetriesTimeoutMillis` instead, then each transport has to look into the fields and do this, or something like it:

```php
$timeout = null;
if ($settings->retryEnabled()) {
    $timeout = $settings->getRetrySettings()->getInitialRpcTimeoutMillis();
} else {
    $timeout = $settings->getRetrySettings()->getNoRetriesTimeoutMillis();
}
```
Right now the transport does this:
```php
$timeout = $settings->getTimeoutMillis();
```

Additionally, `initialRpcTimeoutMillis` intuitively wouldn't change each retry.

michaelbausor: Got it, that explains why I am confused. I think my actual disagreement with what we are doing is focused somewhere else: we shouldn't have `CallSettings` be part of the signature of the callable returned here.

michaelbausor: Got it, that explains why I am confused. I think my actual disagreement with what we are doing is focused somewhere else: we shouldn't have `CallSettings` be part of the signature of the callable returned from here: https://github.com/googleapis/gax-php/pull/118/files#diff-499ff24bbcb2ebc2894a1138637631a5R193

The `CallSettings` object should be used to *construct* the callable, not be passed as an argument to the callable. There should be an options parameter of some sort (it could be an array, or a dedicated class called `TransportCallOptions` or something) but not `CallSettings`

Then the Retrying middleware would accept the `CallSettings` object in its constructor here: https://github.com/googleapis/gax-php/pull/118/files#diff-fe4a9f1b4b922c01aa282bd750c22646R149

And the action that the Retrying middleware would take is to modify the timeout in `TransportCallOptions` (or the array of options) and pass it down the stack.

Does that make sense?

bshaffer: Yes it makes sense. What if we just separate `CallSettings` and `RetrySettings`?

```php
public function getCallable(RetrySettings $retrySettings)
{
    $callable = function (Call $call, CallSettings $settings) {
        // ...
    }
    // other middlewares added here
    return Middleware\RetryMiddleware($callable, $retrySettings);
}
```

And then in the GAPIC clients:

```php
public function startCall(Call $call, CallSettings $settings, RetrySettings, $retrySettings)
{
    $callable = $this->transport->getCallable($retrySettings);
    return $callable(
        $call,
        $settings
    );
}
```

michaelbausor: Yes, that works. But then should we pass `RetrySettings` directly, or should it be part of an $options array, or another class, perhaps `CallConstructionSettings`?

The reason I am pushing for that is because if we add a new setting/option that is required at callable construction time, we will need to pass it into `startCall`, but it doesn't make sense to add it to `CallSettings` if we are using `CallSettings` as the settings for *making* the call rather than *constructing* the call.

```
public function startCall(Call $call, CallSettings $settings, array $callConstructionOptions)
{
    // $callConstructionOptions contains 'retrySettings' => RetrySettings()'
    $callable = $this->transport->getCallable($callConstructionOptions);
    return $callable(
        $call,
        $settings
    );
}
```

dwsupplee: The delineation between settings for making the call and constructing the call make a lot of sense. Since the `Call` object already contains some information for constructing the call, what do you think about utilizing it for construction options like retries?

bshaffer: I'm not a super fan of making it part of the `Call` object, personally. The `$callConstructionOptions` method above sounds good to me.

michaelbausor: I don't have strong feelings about the `Call` object, but it looks like the `Call` object is used when *making* the call as well. I think that whatever we choose, we should have a clean separation between settings used for construction and settings used for making the call.

I also think there is value in keeping the *fixed* values separate from the *changeable* values. Right now the `Call` object contains things that are totally fixed, so (1) perhaps the `Call` object should only be used at construction time, and (2) we should keep those fixed settings separate from the variable settings used at construction time, such as Retry settings.

dwsupplee: How are we defining making the call vs. constructing the call? This is what I have been thinking, please correct me if this is wrong:

[constructing the call](https://github.com/googleapis/gax-php/blob/b3bedf376d58c03ae4528877db9d2b99eb07b963/src/ApiCore/RestTransport.php#L111-L115)
[settings for making (or delivering?) the call](https://github.com/googleapis/gax-php/blob/b3bedf376d58c03ae4528877db9d2b99eb07b963/src/ApiCore/RestTransport.php#L116)

michaelbausor: I was categorizing somewhat differently. Things needed for constructing the call is everything in this method signature: https://github.com/googleapis/gax-php/blob/b3bedf376d58c03ae4528877db9d2b99eb07b963/src/ApiCore/RestTransport.php#L105
Things needed for making the call is everything in the callable signature: https://github.com/googleapis/gax-php/blob/b3bedf376d58c03ae4528877db9d2b99eb07b963/src/ApiCore/RestTransport.php#L107

dwsupplee:
I've pushed up the regenerated GAPICs, please take a look :)

https://github.com/GoogleCloudPlatform/google-cloud-php/tree/transport

src/GapicClientTrait.php
```diff
@@ -0,0 +1,68 @@
...
+        // @todo add rest support
+        // $defaultTransport = $isGrpcExtensionLoaded ? 'grpc' : 'rest';
+        $defaultTransport = 'grpc';
+        $transport = isset($config['transport'])
+            ? strtolower($config['transport'])
+            : $defaultTransport;
+
+        if ($transport === 'grpc') {
+            if (!$isGrpcExtensionLoaded) {
+                throw new Exception(
```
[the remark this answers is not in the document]
dwsupplee: Addressed: https://github.com/googleapis/gax-php/pull/118/commits/d0b24e5db5cd878779dd129d836963e97e0b5d8a

src/GapicClientTrait.php
```diff
@@ -0,0 +1,152 @@
...
+    {
+        $serviceAddress = $this->pluck('serviceAddress', $config);
+        $port = $this->pluck('port', $config);
+        $isGrpcExtensionLoaded = $this->getGrpcDependencyStatus();
+        $defaultTransport = $isGrpcExtensionLoaded
+            ? 'grpc'
+            : 'rest';
+        $transport = isset($config['transport'])
+            ? strtolower($config['transport'])
+            : $defaultTransport;
```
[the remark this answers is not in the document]
dwsupplee: Addressed: https://github.com/googleapis/gax-php/pull/118/commits/d0b24e5db5cd878779dd129d836963e97e0b5d8a

src/GapicClientTrait.php
```diff
@@ -0,0 +1,152 @@
...
+
+        if ($transport === 'grpc' && !$isGrpcExtensionLoaded) {
+            throw new Exception(
+                'gRPC support has been requested but required dependencies ' .
+                'have not been found. Please make sure to run the following ' .
+                'from the command line: pecl install grpc'
```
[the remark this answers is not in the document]
dwsupplee: https://github.com/googleapis/gax-php/pull/118/commits/d0b24e5db5cd878779dd129d836963e97e0b5d8a

src/GrpcTransport.php
```diff
@@ -0,0 +1,259 @@
...
+     * @return PromiseInterface
+     */
+    public function startCall($method, Message $message, $decodeTo, CallSettings $settings)
+    {
+        // @todo move descriptors to call settings?
+        // @todo get middleware working properly
+        // @todo sign request middleware?
+        $handler = function () use ($method, $message, $decodeTo) {
+            $call = $this->_simpleRequest(
+                "/$method",
```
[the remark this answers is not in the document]
dwsupplee: $method here does contain the service name, I borrowed the naming from [here](https://github.com/grpc/grpc-php/blob/master/src/lib/BaseStub.php#L223).

src/GapicClientTrait.php
```diff
@@ -0,0 +1,152 @@
...
+                self::$gapicVersion = trim(file_get_contents(__DIR__.'/../VERSION'));
+            } elseif (class_exists(Version::class)) {
+                self::$gapicVersion = Version::VERSION;
+            }
+        }
+
+        return self::$gapicVersion;
+    }
+
+    private function configureOptions(array $options)
```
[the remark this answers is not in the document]
dwsupplee: I believe it is okay, and yes this is being used in the GAPIC constructors (although has been refactored a bit since this comment)

src/ApiCore/ApiTransportInterface.php
```diff
@@ -29,34 +29,34 @@
...
-    public static function createAny($value)
-    {
-        $any = new Any();
-        $any->setValue($value->serializeToString());
-        return $any;
-    }
+    public function startStreamingCall(
+        Call $call,
+        CallSettings $callSettings,
+        $streamingType,
```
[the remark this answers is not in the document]
dwsupplee: This is done

src/ApiCore/ApiTransportTrait.php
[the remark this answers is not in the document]
dwsupplee: Refactored this out into a factory: https://github.com/googleapis/gax-php/pull/118/commits/aa93ff29d9374202ea6ed0e0ec75452ab9335cc4

src/ApiCore/ApiTransportTrait.php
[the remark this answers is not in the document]
dwsupplee: This is done

src/ApiCore/ApiTransportTrait.php
```diff
@@ -0,0 +1,141 @@
...
+
+    private $agentHeaderDescriptor;
+
+    private function setCommonDefaults(array $options)
+    {
+        $options += [
+            'enableCaching' => true,
+            'authCache' => new MemoryCacheItemPool(),
+            'authCacheOptions' => null,
+            'authHttpHandler' => HttpHandlerFactory::build()
```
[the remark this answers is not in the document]
dwsupplee: All set

src/ApiCore/GrpcTransport.php
dwsupplee: All set

src/ApiCore/ApiStatus.php
dwsupplee: All set: https://github.com/googleapis/gax-php/pull/118/commits/366ca4b610bdd7e0a128941b7426814243db1916

src/ApiCore/BidiStream.php
dwsupplee: All set: https://github.com/googleapis/gax-php/pull/118/commits/30f9995614ee1481f21bb25c034fd5877721a701

src/ApiCore/RestTransport.php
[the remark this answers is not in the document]
dwsupplee: We are using the built in methods now

src/ApiCore/ApiTransportInterface.php
dwsupplee: After a discussion on this, reworked the transport interface a bit: https://github.com/googleapis/gax-php/pull/118/commits/aa93ff29d9374202ea6ed0e0ec75452ab9335cc4

src/ApiCore/CallSettings.php
dwsupplee: Thanks for the explanation. 

> Right now the Call object contains things that are totally fixed, so (1) perhaps the Call object should only be used at construction time

There is one exception to this in [Page.php](https://github.com/googleapis/gax-php/blob/30f9995614ee1481f21bb25c034fd5877721a701/src/ApiCore/Page.php#L123-L138), the message needs to be updated with the new page token. That bit right there is what actually inspired the decision to have the `Call` class in the first place.

dwsupplee: Additionally, I like the idea of the `Call` being something that could eventually be modified by users when they are able to inject their own middleware layers.  I could just be overestimating the usefulness of that, however.

michaelbausor: Hmm, I think I have been a bit confused (and confusing) about these distinctions. What about a different, clearer set of distinctions between the inputs:
  (1) Things that will ALWAYS be the same between calls (like $method, $decodeType in Call object)
  (2) Things that can change, but are NOT used by the transport layer (retries)
  (3) Things that can change, and ARE used by the transport layer ($message in Call object, timeout, headers).

I think that (1) and (2) should be passed to `getCallable`, and (3) should be in the signature of the callable.

Note also that the function of middlewares is to modify the things in bullet 3 before passing them on to the transport, e.g. retry middleware adjusts the timeout based on the multiplier, header middleware augments the headers. The Page object is not strictly middleware, but it does a similar thing by modifying the page token in the request. (Should it be middleware??)

Right now the `Call` object spans (1) and (3). I don't know what the right restructuring is, but I think we should separate those into different objects/args. I also like the idea of custom middleware. I think that bundling the request object, timeout and headers into a single object (`Call` or whatever we call it) without the fixed $method and $decodeType makes it easy to see what your middleware should be modifying. WDYT?

dwsupplee: >  I think that bundling the request object, timeout and headers into a single object (Call or whatever we call it) without the fixed $method and $decodeType makes it easy to see what your middleware should be modifying. WDYT?

I'm on board with that 👍

dwsupplee: Been thinking about this a bit. While items like `$method` and `$decodeType` may not need to be modified, is there harm in exposing them for conditional logic in middleware?

As a hypothetical (and not very practical) example,

What if someone would like to add a special header in the case of `$decodeType === SomeExpectedType::class`?

src/ApiCore/GapicClientTrait.php
```diff
@@ -118,56 +119,209 @@ private function setClientOptions(array $options)
...
+     * }
+     * @param array $callConstructionOptions {
+     *     Call Construction Options
+     *
+     *     @type RetrySettings $retrySettings [optional] A retry settings override
+     *           For the call.
+     * }
+     * @return Promise
+     */
+    private function startCall(Call $call, array $options, array $callConstructionOptions)
```
michaelbausor: Should this be `startUnaryCall`?

src/ApiCore/GapicClientTrait.php
```diff
@@ -118,56 +119,209 @@ private function setClientOptions(array $options)
...
+     * }
      * @return callable
      */
-    private function getServerStreamingCallable()
+    private function createCallStack(callable $callable, array $callConstructionOptions)
+    {
+        return new RetryMiddleware(
+            new AgentHeaderMiddleware(
+                $callable,
+                $this->agentHeaderDescriptor
```
michaelbausor: Could this be passed through `$callConstructionOptions` instead of accessed through `$this`?

src/ApiCore/GapicClientTrait.php
```diff
@@ -197,12 +354,14 @@ private function startOperationsCall(
      *
      * @return PagedListResponse
      */
-    private function getPagedListResponse(Call $call, array $options, array $descriptor)
+    private function getPagedListResponse(Call $call, array $optionalArgs, array $descriptor)
```
michaelbausor: Should this be `startPagedUnaryCall`? But it isn't starting a call, it is finishing it... maybe `makePagedUnaryCall`? I'm not sure

src/ApiCore/Call.php
```diff
@@ -0,0 +1,128 @@
...
+ */
+class Call
+{
+    const UNARY_CALL = 0;
+    const BIDI_STREAMING_CALL = 1;
+    const CLIENT_STREAMING_CALL = 2;
+    const SERVER_STREAMING_CALL = 3;
+
+    private $method;
+    private $type;
```
michaelbausor: Maybe this could have a more descriptive name, like `$callType`?

src/ApiCore/Call.php
```diff
@@ -0,0 +1,128 @@
...
+     * @param string $decodeType
+     * @param Message $message
+     * @param array $descriptor
+     */
+    public function __construct(
+        $method,
+        $decodeType,
+        Message $message = null,
+        $descriptor = [],
+        $type = Call::UNARY_CALL
```
michaelbausor: Nit: ordering of fields and docs in constructor should match ordering of constructor args.

src/ApiCore/GapicClientTrait.php
```diff
@@ -0,0 +1,316 @@
...
+            }
+        }
+
+        return self::$gapicVersion;
+    }
+
+    /**
+     * Configures the GAPIC client based on an array of options.
+     *
+     * @param array $options {
```
michaelbausor: Ae these docs out of date? Looks like we check for a lot of options not listed

tests/ApiCore/Tests/Mocks/mocks.proto
```diff
@@ -0,0 +1,19 @@
+syntax = "proto3";
+
+package google.apiCore.tests.mocks;
```
michaelbausor: I like making this a proto!

I don't think it is normal to camelcase in the proto package option. To get the correct camelcased PHP namespace, you can use:
`php_namespace = "Google\\ApiCore\\Tests\Mocks";`

src/ApiCore/GapicClientTrait.php
bshaffer: Fixed by David.

src/ApiCore/RequestBuilder.php
```diff
@@ -0,0 +1,234 @@
+<?php
+/*
+ * Copyright 2017, Google Inc.
```
michaelbausor: Should this be 2018?

dwsupplee:
Updated copyright years, PTAL.

PR #120, opened 2017-11-01
Pass through the auth http handler as an option

RyanGordon:
First step in fixing https://github.com/GoogleCloudPlatform/google-cloud-php/issues/725


bshaffer:
The `GrpcCredentialsHelper` class has been removed, and has been replaced with the [`TransportFactory`](https://github.com/googleapis/gax-php/blob/master/src/ApiCore/TransportFactory.php#L128) which should be propagating `$authHttpHandler` as expected.

Thank you for your contribution, and please use [`0.30`](https://github.com/googleapis/gax-php/releases/tag/0.30.0) for the latest changes and the fix.

PR #125, opened 2017-11-17
Restructure and update namespace

michaelbausor:
- Change namespace to ApiCore
- Move LongRunning generated client into LongRunning namespace
- Move Google Common Protos into separate folder
- Update some docs on streaming classes

In addition to general review, please let me know what you think about google-common-protos folder. We had some discussion of this already. Should this live inside the `src` directory? Have a different name?


michaelbausor:
@dwsupplee Re tests: good point, I will change it. I had done this as part of creating the separate google-common-protos, google-api-core folders, but since that has been changed back it also makes sense to put the tests at the top level again.

Re LongRunning, I think we have two choices for the namespace of the LongRunning generated client:
1. Google\ApiCore\LongRunning - this is what you are suggesting, and makes sense for the reasons you say. The downside is that the protobuf types that are used in the surface (GetOperationRequest, Operation, etc.) live in Google\LongRunning, so there is a difference between the client namespace and the namespace of request/response types used.*
2. Google\LongRunning - this is what I have at the moment. This is less consistent will the rest of ApiCore, but it means that the client and the protobuf types used in the surface are in the same namespace.

I am not opposed to putting the longrunning client in Google\ApiCore\LongRunning - do you think that will be more usable?

* Right now the protobuf types are actually in Google\Longrunning, but we will use the php_namespace option to change this to Google\LongRunning

dwsupplee:
>  The downside is that the protobuf types that are used in the surface (GetOperationRequest, Operation, etc.) live in Google\LongRunning, so there is a difference between the client namespace and the namespace of request/response types used.*

I understand the desire to keep the common generated proto classes namespaced as is, but this strikes me as further evidence to port them into the ApiCore (or common) namespace.

However, assuming we need to keep the generated proto classes where they are, it seems like there are very fair pros/cons for either approach. Perhaps we should bring it to a vote?

Personally I lean towards `Google\ApiCore\LongRunning`

/cc @bshaffer @jdpedrie - thoughts?

dwsupplee:
>  I don't think it's a hard rule that everything outside of ApiCore is strictly protos

Should we decide on a hard rule, though? If we go the route of having `LongRunning` be in its own namespace, why don't we just move everything in ApiCore up (e.g. `Google\CallSettings`)? What happens if down the line we introduce more classes that happen to rely on namespaces introduced by the protos? It seems we would be watering down the package of `ApiCore`.

I guess I just don't follow why we wouldn't want to contain as much of the code as possible in a package (ApiCore) under our vendor name (Google).

michaelbausor:
@dwsupplee I don't think this is a slippery slope. We can have our code always live in `ApiCore`, unless there is a (strong enough) reason to do otherwise. There isn't the expectation that the set of protos in common protos will expand, or that we will add new code, generated or otherwise. That said, if we do, we can make a similar assessment to what we are doing now.

I think we can assess this on its merits - what is the benefit of consistency between LongRunning protos and generated client, vs. putting as much as possible into ApiCore?

I don't feel that strongly, but I would also lean towards `Google\LongRunning`.

@jdpedrie do you want to weigh in?

jdpedrie:
In order to more closely follow conventions widely accepted in the PHP community, I am in favor of maintaining a single vendor/package root namespace in a package. Therefore, my vote is towards `Google\ApiCore\LongRunning`. To maintain the consistency you're after, Mike, I'd also vote for adding the php_namespace to the LongRunning protos and namespacing them within `Google\ApiCore\LongRunning`.

dwsupplee:
Hm. Well, should we let this be decided by a dance off?

michaelbausor:
@dwsupplee and @jdpedrie, you guys seem to feel more strongly on the issue than me and @bshaffer - barring a strong objection, I will move it into ApiCore.

bshaffer:
+1 Sounds good to me!

PR #129, opened 2017-12-12
RFC: Retry refactor for JSON transport

bshaffer:
An example PR of what it would look like to remove `noRetriesRpcTimeoutMillis` all together.


bshaffer:
Closing this as we agreed on a different solution!

PR #131, opened 2017-12-19
Refactor gapic/transport call stack

bshaffer:
Not tested, but should work as a good proof of concept. 

This change removes almost twice as much code as it adds. That's pretty cool.


src/ApiCore/GapicClientTrait.php
```diff
@@ -83,57 +100,87 @@ private function setClientOptions(array $options)
...
             : TransportFactory::build($options);
     }
 
-    private function configureCallSettings($method, array $optionalArgs)
+    /**
+     * @param string $method
+     *
+     * @return callable
+     */
+    private function getUnaryCallable($method)
```
michaelbausor: Should this perhaps have a more general signature, like accepting an options array? That would support future additions.

bshaffer: I'm not crazy about having an undefined $options array. It would be more confusing, since it has the same name as the callable $options argument. We could just pass `Call` in, since $method is a part of that, but I felt like that was confusing as well. At least this way what the method needs is explicitly defined. Also, it's private so we can change this signature at any time.

Another option would be to remove `$method` and send the entire retrySettings config array to RetryMiddleware. But this seems a little hackier to me than what we have here.

Issue #130, opened 2017-12-13
Cryptic error message when passing empty strings to path template


michaelbausor:
Copied from: https://github.com/GoogleCloudPlatform/google-cloud-php/issues/789
cc @bshaffer 

I get the following error when running [this Monitoring Sample](https://github.com/GoogleCloudPlatform/php-docs-samples/blob/master/appengine/standard/grpc/monitoring.php) and providing empty strings from Project ID and Instance ID:

```
Fatal error: Uncaught Exception: Parse error on line 1: ^ Expecting FORWARD_SLASH, WILDCARD, PATH_WILDCARD, LITERAL, LEFT_BRACE, got 'NOTHING' in vendor/google/gax/src/Jison/Parser.php:499 Stack trace:
#0 vendor/google/gax/src/Jison/Parser.php(561): Google\GAX\Jison\Parser->parseError('Parse error on ...', Object(Google\GAX\Jison\ParserError))
#1 vendor/google/gax/src/Parser.php(66): Google\GAX\Jison\Parser->parse(false)
#2 vendor/google/gax/src/PathTemplate.php(59): Google\GAX\Parser->parse(false)
#3 vendor/google/gax/src/PathTemplate.php(104): Google\GAX\PathTemplate->__construct(false)
#4 vendor/google/cloud-monitoring/V3/Gapic/MetricServiceGapicClient.php(236): Google\GAX\PathTemplate->render(Array)
#5 monitoring.php(25): Google\Cloud\Monitoring\V3\Gapic\MetricServiceGapicClient::projectName(false)
#6 {main} Next Google\GAX\ValidationException: Exception in parser in vendor/google/gax/src/Parser.php:84
Stack trace:
#0 vendor/google/gax/src/PathTemplate.php(59): Google\GAX in vendor/google/gax/src/Parser.php on line 84
```

Here is a full executable script (using `google/cloud-monitoring:^0.6`). I assume the issue is actually in `google/gax`, but wanted to log it here with context:

```php
require_once __DIR__ . '/vendor/autoload.php';

use Google\Api\Metric;
use Google\Api\MonitoredResource;
use Google\Cloud\Monitoring\V3\MetricServiceClient;
use Google\Monitoring\V3\Point;
use Google\Monitoring\V3\TimeInterval;
use Google\Monitoring\V3\TimeSeries;
use Google\Monitoring\V3\TypedValue;
use Google\Protobuf\Timestamp;

// These variables are set by the App Engine environment. To test locally,
[...]

PR #140, opened 2017-12-27
Fix poor errors from path template for empty input

michaelbausor:
Fixes https://github.com/googleapis/gax-php/issues/130

### DIFF
--- a/src/ApiCore/Parser.php
+++ b/src/ApiCore/Parser.php
@@ -92,7 +92,6 @@ class Parser
             }
         }
         return $segments;
-
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/ApiCore/Parser.php
```diff
@@ -60,29 +60,38 @@ public function __construct()
      */
     public function parse($data)
     {
+        if (is_null($data)) {
```
vam-google: Does [empty()](http://php.net/manual/en/function.empty.php) (available starting from php4) allow to combine these two if statements in one? Youl would have to have one error message instead of two more specific ones (something like `'Cannot parse empty $data parameter'`, but it seems ok for this case). Same applies to similar check in `PathTemplate.__constructor()`.

michaelbausor: Fixed

PR #141, opened 2018-01-02
adds support for :action in RequestBuilder

bshaffer:
We had some issues regarding the path template parser when colons were used in the URL for some actions. The issue was for URLs in the Spanner API like this:

```
/v1/projects/{name=projects/*/instances/*/operations/*}:cancel
```

The existing parser did not support this. To fix this, I added support for colons after wildcards. The updates in PathTemplate are a bit strange, but essentially the logic says "colons are allowed in wildcard names and also in path templates". The feature is URLs like this are supported. 

```
/v1/*:action
/v1/*:action/path
/v1/{name=*}:action
/v1/{name=*}:action/path
```

A potential negative is that for URLs *without* a colon, if the wildcard name was appended with a colon (in this case, ":cancel") the other URL would be used, e.g. `/v1/*` could potentially render `/v1/*:action`. This isn't optimal, but it was the only implementation that allowed the current tests to pass.

### DIFF
--- a/src/ApiCore/Jison/Parser.jison
+++ b/src/ApiCore/Jison/Parser.jison
@@ -46,6 +46,7 @@
 "{"                     return 'LEFT_BRACE'
 "}"                     return 'RIGHT_BRACE'
 "="                     return 'EQUALS'
+":"                     return 'COLON'
 "**"                    return 'PATH_WILDCARD'
 "*"                     return 'WILDCARD'
 [^*=}{}\/]+             return 'LITERAL'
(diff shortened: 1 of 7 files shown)


michaelbausor:
The changes look fine, although it seems that some of the tests are failing.

Can you also add a test that uses a colon in the user input field? For example, a path template string `"projects/{project}"`, and then `$pathTemplate->format(['project' => 'google.com:my-project'])`. This is a valid format, so we should check that this is still supported after this change (I think it should be, but am not completely sure).

PR #142, opened 2018-01-10
Allow turning off of REST performance warning

bshaffer:
```php
Google\ApiCore\TransportFactory::disableRestPerformanceNotice();
```


bshaffer:
Added to #118 directly

PR #145, opened 2018-01-31
Update auth handling

michaelbausor:
- Applies changes from https://github.com/googleapis/gax-php/pull/111 after merge of http transport PR.
- Mild refactoring to add AuthWrapper class

cc @pongad

### DIFF
--- a/src/ApiCore/Transport/GrpcTransport.php
+++ b/src/ApiCore/Transport/GrpcTransport.php
@@ -38,14 +38,10 @@ use Google\ApiCore\BidiStream;
 use Google\ApiCore\Call;
 use Google\ApiCore\ClientStream;
 use Google\ApiCore\ServerStream;
-use Google\Auth\FetchAuthTokenInterface;
-use Google\Protobuf\Internal\Message;
 use Google\Rpc\Code;
-use GPBMetadata\Google\Api\Auth;
 use Grpc\BaseStub;
 use Grpc\Channel;
 use GuzzleHttp\Promise\Promise;
-use GuzzleHttp\Promise\PromiseInterface;
 
 /**
  * A gRPC based transport implementation.
--- a/src/ApiCore/Transport/GrpcTransport.php
+++ b/src/ApiCore/Transport/GrpcTransport.php
@@ -33,6 +33,7 @@
 namespace Google\ApiCore\Transport;
 
 use Google\ApiCore\ApiException;
+use Google\ApiCore\AuthWrapper;
 use Google\ApiCore\BidiStream;
 use Google\ApiCore\Call;
 use Google\ApiCore\ClientStream;
@@ -40,6 +41,7 @@ use Google\ApiCore\ServerStream;
 use Google\Auth\FetchAuthTokenInterface;
 use Google\Protobuf\Internal\Message;
 use Google\Rpc\Code;
+use GPBMetadata\Google\Api\Auth;
 use Grpc\BaseStub;
 use Grpc\Channel;
 use GuzzleHttp\Promise\Promise;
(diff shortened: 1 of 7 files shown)


src/ApiCore/Transport/GrpcTransport.php
```diff
@@ -33,13 +33,15 @@
 namespace Google\ApiCore\Transport;
 
 use Google\ApiCore\ApiException;
+use Google\ApiCore\AuthWrapper;
 use Google\ApiCore\BidiStream;
 use Google\ApiCore\Call;
 use Google\ApiCore\ClientStream;
 use Google\ApiCore\ServerStream;
 use Google\Auth\FetchAuthTokenInterface;
```
dwsupplee: This import can be removed now

src/ApiCore/Transport/RestTransport.php
```diff
@@ -33,6 +33,7 @@
 
 use Google\ApiCore\ApiException;
 use Google\ApiCore\ApiStatus;
+use Google\ApiCore\AuthWrapper;
 use Google\ApiCore\Call;
 use Google\ApiCore\RequestBuilder;
 use Google\Auth\FetchAuthTokenInterface;
```
dwsupplee: ditto

src/ApiCore/Transport/GrpcTransport.php
```diff
@@ -33,13 +33,15 @@
...
 use Google\Protobuf\Internal\Message;
 use Google\Rpc\Code;
+use GPBMetadata\Google\Api\Auth;
```
dwsupplee: Is this used?

src/ApiCore/TransportFactory.php
```diff
@@ -40,6 +40,7 @@
 use Google\Auth\CredentialsLoader;
 use Google\Auth\FetchAuthTokenCache;
 use Google\Auth\HttpHandler\HttpHandlerFactory;
+use GPBMetadata\Google\Api\Auth;
```
dwsupplee: ditto

src/ApiCore/TransportFactory.php
```diff
@@ -226,33 +221,35 @@ private static function handleTransport($transport)
...
      */
-    private static function handleCredentialsLoader(array $args)
+    private static function buildAuthWrapper(array $args)
     {
+        $authHttpHandler = $args['authHttpHandler'] ?: HttpHandlerFactory::build();
+
         if (isset($args['credentialsLoader'])) {
-            return $args['credentialsLoader'];
-        }
+            $fetchAuthTokenInterface = $args['credentialsLoader'];
```
dwsupplee: Since this isn't an interface, but an implementation, what do you think about calling it something like `fetcher` or `credentialsLoader`?

bshaffer:
I'm a little confused. Isn't this something we should implement in `google/auth`? 

I believe the idea is, since `scope` is the same per client, each client reuses a token in memory w/o calling a cache pool, which should save us a little bit of performance.

If this is the case, we could look at a flag for keeping the token in memory in the auth class. The only time this would cause an issue is if the same auth class was used for two different API calls (w/ different scopes), which I doubt is a common use case.

michaelbausor:
I am not opposed to pushing logic like this into the auth library when appropriate.

The main change here is to make use of the `getLastReceivedToken()` method in `FetchAuthTokenInterface`, which right now we don't use at all. If we would rather push that logic into the auth lib, then it would be sufficient to add a new method like `getOrFetchAuthToken()` that uses `getLastReceivedToken()` if it is available and valid, and otherwise uses `fetchAuthToken()`.

PR #146, opened 2018-03-06
Update constructor arguments

michaelbausor:
WIP PR to update GAPIC constructor arguments. Still need to add more tests.

cc @jdpedrie - I can't add you as a reviewer?


src/ApiCore/TransportFactory.php
```diff
@@ -55,20 +56,16 @@ class TransportFactory
...
-        'authHttpHandler'   => null,
-        'transport'         => null
+        'authHttpHandler'   => null
     ];
 
     /**
      * Builds a transport given an array of arguments.
      *
-     * @param array $args {
+     * @param string $transport The type of transport to build.
```
dwsupplee: Since it is optional, what is the argument for making `transport` its own parameter?

michaelbausor: Hmm, you are right. I was treating it as required, but that is not correct.

src/ApiCore/TransportFactory.php
```diff
@@ -55,20 +56,16 @@ class TransportFactory
...
+     * @param string $serviceAddress The address of the API remote host.
```
jdpedrie: This should make explicit that ports can be provided as part of the address.

michaelbausor:
PTAL
Toolkit PR: https://github.com/googleapis/toolkit/pull/1882

src/ApiCore/TransportFactory.php
```diff
@@ -35,99 +35,40 @@
...
+     *           Optional. The type of transport to build. Defaults to
+     *           'grpc' when available. Supported values: ['grpc', 'rest'].
+     *     @type string $restClientConfigPath Path to rest client config.
+     *           Required for 'rest' transport only.
      * }
      * @return TransportInterface
+     * @throws ValidationException
      */
-    public static function build(array $args)
+    public static function build($serviceAddress, $authWrapper, array $args)
```
dwsupplee: The original intent is that users may choose to be using this factory to instantiate their transport, and this feels like a bit of a nerf. Prior to this change, if you wanted to use the factory to build your transport it was as straight forward as passing in the service address and port at minimum. Now you have to concern yourself with the concept of the auth wrapper, as well as how to construct one. I may be missing something here, I just don't see the argument for this change.

src/ApiCore/TransportFactory.php
```diff
@@ -136,35 +77,36 @@ public static function build(array $args)
...
-                }
+
+                $stubOpts = ['credentials'=> self::createSslChannelCredentials()];
 
                 return new GrpcTransport(
-                    $host,
+                    $serviceAddress,
                     $authWrapper,
-                    $stubOpts,
-                    $args['channel']
```
dwsupplee: Why are we removing the capability to provide a channel, override the `force_new` option, and to provide ssl credentials?

src/ApiCore/TransportFactory.php
```diff
@@ -201,54 +131,25 @@ protected static function createSslChannelCredentials()
...
 
-    /**
-     * @param string|null $transport
-     * @return string
-     */
-    private static function handleTransport($transport)
+    private static function defaultTransport()
     {
-        if ($transport) {
-            return strtolower($transport);
```
dwsupplee: Why remove this? It just handles cases where users provide an upper case string of 'REST' or 'gRPC'.

src/ApiCore/TransportFactory.php
```diff
@@ -201,54 +131,25 @@ protected static function createSslChannelCredentials()
...
-                $credentialsLoader = new FetchAuthTokenCache(
-                    $credentialsLoader,
-                    $args['authCacheOptions'],
-                    $args['authCache']
-                );
-            }
+        $components = explode(':', $serviceAddress);
+        if (count($components) !== 2) {
+            throw new ValidationException(
+                'Invalid serviceAddress. Expected format "<address>:<port>", got ' . $serviceAddress);
```
dwsupplee: The `);` should be on a new line. Also, if a port is not included does it seem reasonable to default to 443? That is what we do currently and I don't see a reason to change it.

src/ApiCore/TransportFactory.php
```diff
@@ -201,54 +131,25 @@ protected static function createSslChannelCredentials()
...
 
     /**
-     * @param array $args
-     * @return AuthWrapper
+     * @param $serviceAddress
+     * @return array
+     * @throws ValidationException
      */
-    private static function buildAuthWrapper(array $args)
+    private static function validateServiceAddress($serviceAddress)
```
dwsupplee: This name does not ring true to what the method actually does.

src/ApiCore/TransportFactory.php
```diff
@@ -136,35 +77,36 @@ public static function build(array $args)
...
+                }
+
                 return new RestTransport(
-                    new RequestBuilder(
-                        $host,
-                        $args['restClientConfigPath']
-                    ),
+                    $requestBuilder,
                     $authWrapper,
-                    $args['httpHandler'] ?: [HttpHandlerFactory::build(), 'async']
```
dwsupplee: This allowed for supplying a custom function to handle the REST calls. Why is it removed?

src/ApiCore/GapicClientTrait.php
```diff
@@ -96,105 +97,165 @@ private static function getGapicVersion(array $options)
...
-     *           NOTE: This option is only valid when utilizing the gRPC
-     *           transport. Also, if the $channel optional argument is
-     *           specified, then this argument is unused.
-     *     @type bool $forceNewChannel
-     *           If true, this forces gRPC to create a new channel instead of
-     *           using a persistent channel. Defaults to false.
-     *           NOTE: This option is only valid when utilizing the gRPC
-     *           transport. Also, if the $channel optional argument is
-     *           specified, then this option is unused.
-     *     @type CredentialsLoader $credentialsLoader
```
dwsupplee: Is it now the case users cannot provide a `credentialsLoader` instance of their own, unless they create their own `Transport` without using the factory?

src/ApiCore/AuthWrapper.php
```diff
@@ -55,6 +61,54 @@ public function __construct(FetchAuthTokenInterface $fetchAuthTokenInterface, ca
...
+        try {
+            $authHttpHandler = HttpHandlerFactory::build();
+        } catch (\Exception $ex) {
+            throw new ValidationException("Failed to create authHttpHandler", $ex->getCode(), $ex);
+        }
+
+        if (is_null($keyFile)) {
+            $loader = ApplicationDefaultCredentials::getCredentials($scopes, $authHttpHandler);
+        } else {
+            $loader = new ServiceAccountCredentials($scopes, $keyFile);
```
dwsupplee: This looks like we should be using [`CredentialsLoader::makeCredentials()`](https://github.com/google/google-auth-library-php/blob/master/src/CredentialsLoader.php#L115)

src/ApiCore/GapicClientTrait.php
```diff
@@ -96,105 +97,165 @@ private static function getGapicVersion(array $options)
...
+     */
+    private function validateOptions(array $options)
     {
         $this->validateNotNull($options, [
             'serviceName',
+            'serviceAddress',
             'descriptorsConfigPath',
-            'clientConfigPath'
+            'clientConfigPath',
+            'scopes',
```
dwsupplee: My stance is that `scopes` and `serviceAddress` should be validated in the transport factory and not elsewhere. That would be the ideal place to manage that data, as users may use the transport factory on their own right and should get the same errors bubbling up there as they would from this use case.

src/ApiCore/GapicClientTrait.php
```diff
@@ -96,105 +97,165 @@ private static function getGapicVersion(array $options)
...
-            : null;
-        $clientConfig = json_decode(
-            file_get_contents($options['clientConfigPath']),
-            true
-        );
-        $this->serviceName = $options['serviceName'];
-        $this->retrySettings = RetrySettings::load(
-            $this->serviceName,
+
+        if (isset($options['transport'])) {
```
dwsupplee: I believe the same should apply here as well. Does it seem we should worry about validating the transport in the factory, as this just gets passed off there already?

src/ApiCore/GapicClientTrait.php
```diff
@@ -96,105 +97,165 @@ private static function getGapicVersion(array $options)
...
+            $retrySettings = $updatedRetrySettings;
         }
+
+        $this->retrySettings = $retrySettings;
+    }
+
+    /**
+     * @param array $options
+     */
+    private function setAgentHeaderDescriptor($options)
```
dwsupplee: These could all use an array typehint.

src/ApiCore/GapicClientTrait.php
```diff
@@ -96,105 +97,165 @@ private static function getGapicVersion(array $options)
...
+
+    /**
+     * @param array $options
+     */
+    private function setServiceNameAndDescriptors($options)
+    {
+        $serviceName = $options['serviceName'];
+        $descriptors = require($options['descriptorsConfigPath']);
+
+        $this->serviceName = $serviceName;
```
dwsupplee: `$this->serviceName = $options['serviceName'];`?

src/ApiCore/AuthWrapper.php
```diff
@@ -55,6 +61,54 @@ public function __construct(FetchAuthTokenInterface $fetchAuthTokenInterface, ca
...
+            'authCache' => null,
+        ];
+
+        $keyFile = $args['keyFile'] ?: $args['keyFilePath'];
+        $authCache = $args['authCache'] ?: new MemoryCacheItemPool();
+
+        try {
+            $authHttpHandler = HttpHandlerFactory::build();
+        } catch (\Exception $ex) {
+            throw new ValidationException("Failed to create authHttpHandler", $ex->getCode(), $ex);
```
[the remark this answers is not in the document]
michaelbausor: Regarding exception throwing - I don't have overly strong feelings about what it best. We define ApiException for exceptions caused by remote calls, and ValidationException for "local" exceptions. We could throw InvalidArgumentException in place of a lot of the ValidationException calls.

Regarding this case specifically - what is the consensus around throwing and catching the base Exception class? In general, it seems like bad practice, because any catch block will catch everything, even exceptions it can't handle. So I think there is an argument for converting base Exceptions into some more specific type.

src/ApiCore/TransportFactory.php
michaelbausor: I think we want to encourage users to get the casing right - particularly because we now support a $transportConfig array that is keyed to the transport. What if they supply `$transportConfig = ['GRPC' => [...]]`. Should we support that?

src/ApiCore/TransportFactory.php
michaelbausor: Makes sense, will support.

src/ApiCore/AuthWrapper.php
michaelbausor: Done

src/ApiCore/GapicClientTrait.php
michaelbausor: Done

michaelbausor:
Lots of changes based on offline discussion, PTAL. I have tried to reply to a lot of the comments that are still relevant, please re-up any that you think still need addressing.

This PR also now incorporates #164 - it may make review of this PR easier if we get that merged first.

src/ApiCore/GapicClientTrait.php
```diff
@@ -95,117 +99,193 @@ private static function getGapicVersion(array $options)
...
+        $clientConfig = $options['clientConfig'];
+        if (is_string($clientConfig)) {
+            $clientConfig = json_decode(file_get_contents($clientConfig), true);
+        }
         $this->serviceName = $options['serviceName'];
-        $this->retrySettings = RetrySettings::load(
+        $retrySettings = RetrySettings::load(
             $this->serviceName,
             $clientConfig,
-            $this->pluck('retryingOverride', $options, false)
```
dwsupplee: What if we removed [the retryingOverride setting](https://github.com/googleapis/gax-php/blob/master/src/ApiCore/RetrySettings.php#L261) on `RetrySettings::load()` and replaced it with a bool `disableRetries`? We could pass in the value we get at this point and skip needing to reiterate the list again below.

src/ApiCore/GapicClientTrait.php
```diff
@@ -95,117 +99,193 @@ private static function getGapicVersion(array $options)
      * @param array $options {
      *     An array of required and optional arguments.
      *
-     *     @type string $versionFile
```
dwsupplee: I think this is still used: https://github.com/michaelbausor/gax-php/blob/d7cdbdfaedaa4bd74b2306e9322a35b8b8d17e2c/src/ApiCore/GapicClientTrait.php#L84

src/ApiCore/GapicClientTrait.php
```diff
@@ -95,117 +99,193 @@ private static function getGapicVersion(array $options)
...
         $descriptors = require($options['descriptorsConfigPath']);
         $this->descriptors = $descriptors['interfaces'][$this->serviceName];
 
-        if (isset($options['credentialsLoader'])) {
-            $authHttpHandler = isset($options['authHttpHandler']) ? $options['authHttpHandler'] : null;
-            $this->authWrapper = new AuthWrapper($options['credentialsLoader'], $authHttpHandler);
-        } else {
-            $this->authWrapper = AuthWrapper::build($options);
-        }
+        $serviceAddress = $options['serviceAddress'];
```
dwsupplee: Any reason to not just access the array instead of creating new variables? :)

src/ApiCore/GapicClientTrait.php
```diff
@@ -95,117 +99,193 @@ private static function getGapicVersion(array $options)
...
+                "'transport' must be a string, instead got:" .
+                print_r($transport, true)
+            );
+        }
+        $configForSpecifiedTransport = isset($transportConfig[$transport])
+            ? $transportConfig[$transport]
+            : [];
+        switch ($transport) {
+            case 'grpc':
+                self::validateGrpcSupport();
```
dwsupplee: This already happens in [`GrpcTransport::build()`](https://github.com/michaelbausor/gax-php/blob/d7cdbdfaedaa4bd74b2306e9322a35b8b8d17e2c/src/ApiCore/Transport/GrpcTransport.php#L73)

src/ApiCore/Transport/GrpcTransport.php
```diff
@@ -32,38 +32,73 @@
...
+use Google\ApiCore\GapicHelpersTrait;
 use Google\ApiCore\ServerStream;
+use Google\ApiCore\ValidationException;
+use Google\ApiCore\ValidationTrait;
 use Google\Rpc\Code;
 use Grpc\BaseStub;
 use Grpc\Channel;
+use Grpc\ChannelCredentials;
 use GuzzleHttp\Promise\Promise;
+use Throwable;
```
dwsupplee: Unused import

src/ApiCore/Transport/GrpcTransport.php
```diff
@@ -32,38 +32,73 @@
...
-     * @param Channel $channel An already instantiated channel to be used during
-     *        creation of the BaseStub.
+     * Builds a GrpcTransport.
+     *
+     * @param string $serviceAddress
+     *        The address of the API remote host, for example "example.googleapis.com. May also
+     *        include the port, for example "example.googleapis.com:443"
+     * @param array $config
+     *        Config options used to construct the gRPC transport. Supported options are 'stubOpts'
+     *        and 'channel'.
```
dwsupplee: Could we use a style like this for both transports? It makes understanding the expected types easier

```php
    /**
     * @param array $config {
     *    Config options used to construct the gRPC transport.
     *    
     *    @type array $stubOpts Description
     *    @type Channel $channel Description
     * } 
     */
```

src/ApiCore/Transport/GrpcTransport.php
```diff
@@ -32,38 +32,73 @@
...
-        $host,
-        array $stubOpts,
-        Channel $channel = null
-    ) {
-        parent::__construct(
-            $host,
-            $stubOpts,
-            $channel
-        );
+    public static function build($serviceAddress, $config = [])
```
dwsupplee: We could use an array type hint here

src/ApiCore/Transport/RestTransport.php
```diff
@@ -60,6 +68,30 @@ public function __construct(
...
+     *        The address of the API remote host, for example "example.googleapis.com".
+     * @param string $restConfigPath
+     *        Path to rest config file.
+     * @param array $config
+     *        Config options used to construct the gRPC transport. Supported options are
+     *        'httpHandler'.
+     * @return RestTransport
+     * @throws ValidationException
+     */
+    public static function build($serviceAddress, $restConfigPath, $config)
```
dwsupplee: `$config` could use a typehint and should also be optional

src/ApiCore/Transport/RestTransport.php
```diff
@@ -60,6 +68,30 @@ public function __construct(
...
     }
 
+    /**
+     * Builds a RestTransport.
+     *
+     * @param string $serviceAddress
+     *        The address of the API remote host, for example "example.googleapis.com".
+     * @param string $restConfigPath
+     *        Path to rest config file.
+     * @param array $config
```
dwsupplee: I understand if this would be considered outside the scope of this PR, so maybe we can file it as a follow up - but it would be great if we also supported proxying through HTTP client options.

src/ApiCore/AuthWrapper.php
```diff
@@ -114,6 +118,57 @@ public static function build(array $args = [])
...
+    /**
+     * Factory method to create an AuthWrapper from a credentials keyfile and an array of options.
+     *
+     * @param string|array $keyFile
+     *     Credentials to be used. Accepts either a path to a credentials file, or a decoded
+     *     credentials file as a PHP array.
+     * @param array $authConfig {
+     *     @type string[] $scopes The scopes required by this AuthWrapper.
+     *     @type callable $authHttpHandler
+     *           Optional. A handler used to deliver PSR-7 requests specifically
```
dwsupplee: We could probably nuke `optional` from the descriptions, as they all are

src/ApiCore/Transport/GrpcTransport.php
```diff
@@ -32,38 +32,73 @@
...
+        }
+        $channel = $config['channel'];
+        if (!is_null($channel) && !($channel instanceof Channel)) {
+            throw new ValidationException(
+                "Channel argument to GrpcTransport must be of type \Grpc\Channel, " .
+                "instead got: " . print_r($channel, true)
+            );
+        }
+        try {
+            return new GrpcTransport($host, $stubOpts, $channel);
```
dwsupplee: It looks like the constructor may have gone missing :)

src/ApiCore/AuthWrapper.php
```diff
@@ -114,6 +118,57 @@ public static function build(array $args = [])
         return new AuthWrapper($credentialsLoader, $authHttpHandler);
     }
 
+    /**
+     * Factory method to create an AuthWrapper from a credentials keyfile and an array of options.
+     *
+     * @param string|array $keyFile
```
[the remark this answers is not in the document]
michaelbausor: Hmm, not sure. I wanted to create separate methods to try to make the different construction options clearer, but I agree that the duplicated logic is not good and I'm not sure it really achieves the goal of being clearer. WDYT of trying to unify the naming (e.g. `create()`, `createFromKeyFile()`) and making the options identical in each case? (And also trying to reduce the duplication of logic). Or do you think a single method is just better?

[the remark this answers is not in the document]
dwsupplee: I don't want to say a single method is better, it just happens to be my preference here. If we can clean up the logic to reduce duplication, I would be okay with splitting it out. One small note on the naming, though, is that it seems we use `build` for factories elsewhere - maybe it would be helpful to be consistent with that naming here?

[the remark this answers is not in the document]
michaelbausor: I have updated to remove as suggested, on reflection it seemed better for the reasons you suggest - PTAL

PR #149, opened 2018-03-19
Make gapic client trait methods protected

michaelbausor:
This allows a partial veneer to override these methods and modify the behaviour. For example, a partial veneer could override `setClientOptions` to set a default option, or `createCallStack` to insert a new piece of middleware.

### DIFF
--- a/src/ApiCore/GapicClientTrait.php
+++ b/src/ApiCore/GapicClientTrait.php
@@ -161,7 +161,7 @@ trait GapicClientTrait
      *           The code generator version of the GAPIC library.
      * }
      */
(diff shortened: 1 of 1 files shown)


bshaffer:
I understand the justification for `setClientOptions` and `createCallStack`, but what is the use case for the other methods? Is this decision mostly preemptive?

michaelbausor:
Mostly preemptive. I believe that it makes logical sense for the client to be able to override some or all of these methods in certain circumstances. For example, I can imagine an API that has a desire for customized paging or long running operations that we could override here. But you are right that there is no use case for such things now.

PR #151, opened 2018-03-22
Update operation polling

michaelbausor:
@dwsupplee @jdpedrie PTAL when you have a chance. In particular, this PR changes the option keys that are available to configure polling, and renames one method on the OperationResponse class. What are your thoughts on backwards compatibility vs. a breaking change?


michaelbausor:
Also PTAL @bshaffer

src/ApiCore/OperationResponse.php
```diff
@@ -133,40 +164,68 @@ public function getName()
...
      * and therefore will always return true.
      *
      * @param array $options {
      *                       Options for configuring the polling behaviour.
      *
-     *     @type float $pollingIntervalSeconds The polling interval to use, in seconds.
-     *                                          Default: 1.0
-     *     @type float $maxPollingDurationSeconds The maximum amount of time to continue polling.
-     *                                            Default: 0.0 (no maximum)
+     *     @type float $initialPollDelayMillis The initial polling interval to use, in seconds.
```
dwsupplee:
```diff
+ in milliseconds
- in seconds
```

(applies to the other properties too, as well as in the constructor)

src/ApiCore/OperationResponse.php
```diff
@@ -48,26 +52,41 @@
...
      * @param string $operationName
-     * @param \Google\ApiCore\LongRunning\OperationsClient $operationsClient
+     * @param OperationsClient $operationsClient
      * @param array $options {
      *                       Optional. Options for configuring the Operation response object.
      *
      *     @type string $operationReturnType The return type of the longrunning operation.
      *     @type string $metadataReturnType The type of the metadata returned in the Operation response.
-     *     @type \google\longrunning\Operation $lastProtoResponse A response already received from the server.
+     *     @type float $initialPollDelayMillis    The initial polling interval to use, in seconds.
```
dwsupplee: Now that we are using milliseconds, do you think that is enough specificity to use int instead of float?

src/ApiCore/OperationResponse.php
```diff
@@ -133,40 +164,68 @@ public function getName()
...
     public function pollUntilComplete($options = [])
     {
-        $defaultPollSettings = [
-            'pollingIntervalSeconds' => $this::DEFAULT_POLLING_INTERVAL,
-            'maxPollingDurationSeconds' => 0.0,
-        ];
-        $pollSettings = array_merge($defaultPollSettings, $options);
+        $pollSettings = array_merge($this->defaultPollSettings, $options);
+
+        $currentPollDelayMillis = $pollSettings['initialPollDelayMillis'];
```
dwsupplee: These settings seem similar to those used by the retry settings. Do you think it may make sense to have an implementation that accepts these settings and performs the repeated call logic on our behalf? It could be something we use to eventually replace the classes like [`Google\Cloud\Core\ExponentialBackoff`](https://github.com/GoogleCloudPlatform/google-cloud-php/blob/master/Core/src/ExponentialBackoff.php) and could use in our system tests amongst other places.

src/ApiCore/OperationResponse.php
```diff
@@ -133,40 +164,68 @@ public function getName()
...
-        $hasMaxPollingDuration = $maxPollingDuration > 0.0;
-        $endTime = microtime(true) + $maxPollingDuration;
-        while (!$this->isDone() && (!$hasMaxPollingDuration || microtime(true) < $endTime)) {
-            usleep($pollingIntervalMicros);
+        while (true) {
+            if ($this->isDone()) {
+                return true;
+            }
+            if ($hasTotalPollTimeout && $this->getCurrentTimeMillis() > $endTime) {
+                return $this->isDone();
```
dwsupplee: Could this be added as an `elseif` to the above condition and just always return false instead of checking the return of `isDone()` again?

```php
if ($this->isDone()) {
    return true;
} elseif ($hasTotalPollTimeout && $this->getCurrentTimeMillis() > $endTime) {
    return false;
}
```

src/ApiCore/OperationResponse.php
michaelbausor: Fixed

src/ApiCore/OperationResponse.php
michaelbausor: Yep, agreed. Fixed

src/ApiCore/OperationResponse.php
michaelbausor: Fixed

src/ApiCore/OperationResponse.php
michaelbausor: Yes, they are similar. I had a look at trying to combine them, but it is harder than it first seems, because the retrying middleware class needs to retry over promises, so its logic is quite different.

I have moved the polling logic into a trait - let me know if you think this is an improvement or not. I think that this could be used in place of the exponential backoff class - or else, it could supplement it and replace the polling/sleep logic.

michaelbausor:
@dwsupplee merged in master after constructor updates, PTAL

Issue #155, opened 2018-03-27
Improper Parsing of FieldMask for REST requests


bshaffer:
Same as https://github.com/GoogleCloudPlatform/google-cloud-php/issues/958

Field Masks are being improperly parsed as querystrings due to the fact they are handled in a special way by `serializeToJsonString`.  For example, for the following `FieldMask`:

```php
$mask = new FieldMask;
$mask->setPaths(['credentials']);
```

`FieldMask::getQuerystringValue` returns a string, but `RequestBuilder` [expects an array](https://github.com/googleapis/gax-php/blob/df3af58e5f9b29c8126b1191f4b413e13e4eb468/src/ApiCore/RequestBuilder.php#L124). As a result, nothing is rendered, when the correct way to parse it is `update_mask.paths=credentials`. 

When the `FieldMask` is parsed like any other message, this issue goes away. So one possible fix would be for the RequestBuilder to exclude `FieldMask` specifically when calling [`messageToArray`](https://github.com/googleapis/gax-php/blob/master/src/ApiCore/RequestBuilder.php#L222):

```php
if (GPBUtil::hasSpecialJsonMapping($message) && !$message instanceof FieldMask) {
    return json_decode($message->serializeToJsonString(), true);
}
$messageArray = [];
foreach ($this->getAllProperties($message) as $name => $value) {
    $propertyValue = $this->getPrivatePropertyValue($message, $name);
    $messageArray[$name] = $this->getQuerystringValue($propertyValue);
}
return $messageArray;
```

michaelbausor:
I will look at fixing - can you share an example API call you used to find/test this?

bshaffer:
I can submit a fix. I think the best way to do it is to call a local private method instead of using `GPBUtil`, since this is causing https://github.com/GoogleCloudPlatform/google-cloud-php/issues/957

michaelbausor:
Fixed in https://github.com/googleapis/gax-php/pull/156

PR #156, opened 2018-03-27
fixes getQuerystringValue for special JSON cases

bshaffer:
addresses https://github.com/googleapis/gax-php/issues/155


metadata/ApiCore/Testing/Mocks.php
```diff
@@ -1,6 +1,6 @@
 <?php
 # Generated by the protocol buffer compiler.  DO NOT EDIT!
-# source: src/ApiCore/Testing/mocks.proto
```
bshaffer: Couldn't get this change to go away. I generated with the following commands:

```sh
// in root of gax-php repo
$ protoc --php_out=src/ --proto_path=src/ src/ApiCore/Testing/mocks.proto
```

src/ApiCore/RequestBuilder.php
```diff
@@ -120,7 +120,7 @@ public function build($path, Message $message, array $headers = [])
                     }
 
                     $value = $this->getQuerystringValue($propertyValue);
-                    if ($propertyValue instanceof Message) {
+                    if ($propertyValue instanceof Message && is_array($value)) {
```
bshaffer: This is the fix. the special cases reduce `$value` to a scalar, so we can skip the `foreach`

metadata/ApiCore/Testing/Mocks.php
michaelbausor: Odd. I just played around a bit and saw the same thing.

PR #161, opened 2018-04-02
Update auth wrapper to fix segfault

michaelbausor:
This updates the call_credentials_callback function to avoid needing `use ($authWrapper)`, which seems to have been triggering the segfaults.

### DIFF
--- a/src/ApiCore/AuthWrapper.php
+++ b/src/ApiCore/AuthWrapper.php
@@ -84,12 +84,11 @@ class AuthWrapper
         $token = $fetchAuthTokenInterface->getLastReceivedToken();
         if (self::isExpired($token)) {
             $token = $fetchAuthTokenInterface->fetchAuthToken($authHttpHandler);
+            if (!self::isValid($token)) {
+                return '';
+            }
         }
-        if (self::isValid($token)) {
-            return $token['access_token'];
-        } else {
-            return '';
-        }
+        return $token['access_token'];
     }
 
     private static function isValid($token)
--- a/src/ApiCore/AuthWrapper.php
+++ b/src/ApiCore/AuthWrapper.php
@@ -38,8 +38,8 @@ use Google\Auth\FetchAuthTokenInterface;
  */
 class AuthWrapper
 {
-    private $fetchAuthTokenInterface;
-    private $authHttpHandler;
+    public $fetchAuthTokenInterface;
+    public $authHttpHandler;
 
     /**
      * AuthWrapper constructor.
--- a/src/ApiCore/AuthWrapper.php
+++ b/src/ApiCore/AuthWrapper.php
@@ -94,8 +94,7 @@ class AuthWrapper
 
     private static function isValid($token)
     {
-        return !is_null($token)
-            && is_array($token)
+        return is_array($token)
             && array_key_exists('access_token', $token);
     }
 
--- a/src/ApiCore/AuthWrapper.php
+++ b/src/ApiCore/AuthWrapper.php
@@ -63,9 +63,27 @@ class AuthWrapper
     private function getToken()
     {
         $token = $this->fetchAuthTokenInterface->getLastReceivedToken();
-        if (is_null($token) || time() > $token['expires_at']) {
+        if ($this->isExpired($token)) {
             $token = $this->fetchAuthTokenInterface->fetchAuthToken($this->authHttpHandler);
         }
-        return $token['access_token'];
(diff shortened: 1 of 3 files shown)


bshaffer:
I am overjoyed to see this fix to what is almost certainly one of the more bizarre bugs ever seen.

src/ApiCore/AuthWrapper.php
```diff
@@ -57,15 +60,49 @@ public function __construct(FetchAuthTokenInterface $fetchAuthTokenInterface, ca
...
+    public function getAuthorizationHeaderCallback()
+    {
+        $fetchAuthTokenInterface = $this->fetchAuthTokenInterface;
+        $authHttpHandler = $this->authHttpHandler;
+
+        // NOTE: changes to this function should be treated carefully and tested thoroughly. It will
+        // be passed into the gRPC c extension, and changes have the potential to trigger very
+        // difficult-to-diagnose segmentation faults.
+        return function () use ($fetchAuthTokenInterface, $authHttpHandler) {
+            return ['authorization' => ['Bearer ' . self::getToken($fetchAuthTokenInterface, $authHttpHandler)]];
```
dwsupplee: You should be able to access instance variables in an anonymous function, no need for the use statement (or was this part of the reason we were seeing odd behavior? :P).

src/ApiCore/AuthWrapper.php
```diff
@@ -57,15 +60,49 @@ public function __construct(FetchAuthTokenInterface $fetchAuthTokenInterface, ca
...
+        if (self::isValid($token)) {
+            return $token['access_token'];
+        } else {
+            return '';
+        }
+    }
+
+    private static function isValid($token)
+    {
+        return !is_null($token)
```
dwsupplee: I think the `is_array` check will be enough, it should return false if a null value is passed in.

src/ApiCore/AuthWrapper.php
```diff
@@ -57,15 +60,49 @@ public function __construct(FetchAuthTokenInterface $fetchAuthTokenInterface, ca
...
     {
-        $token = $this->fetchAuthTokenInterface->getLastReceivedToken();
-        if (is_null($token) || time() > $token['expires_at']) {
-            $token = $this->fetchAuthTokenInterface->fetchAuthToken($this->authHttpHandler);
+        $token = $fetchAuthTokenInterface->getLastReceivedToken();
+        if (self::isExpired($token)) {
+            $token = $fetchAuthTokenInterface->fetchAuthToken($authHttpHandler);
         }
-        return $token['access_token'];
+        if (self::isValid($token)) {
```
dwsupplee: Since `isExpired` checks validity as well, WDYT about moving this check into the condition above?

```php
$token = $fetchAuthTokenInterface->getLastReceivedToken();
if (self::isExpired($token)) {
    $token = $fetchAuthTokenInterface->fetchAuthToken($authHttpHandler);
    if (!self::isValid($token)) {
         return '';
    }
}

return $token['access_token'];
```

michaelbausor: The case I wanted to try to catch was if `$token` was invalid, then avoid trying to access the array with `$token['access_token']`. In turned out not to be the case, but I was wondering if trying to do an array access on a non-array object in this function (but called from the c extension) might have been causing the segfault.

Rearranging in that way means that if `getLastReceivedToken()` returns an invalid value, we will still try to perform array access.

dwsupplee: I thought `isExpired` checks validity, so in that case wouldn't it fall through and attempt to fetch a fresh token?

michaelbausor: Ah yes, you are totally right. Fixed.

src/ApiCore/AuthWrapper.php
dwsupplee: http://php.net/manual/en/functions.anonymous.php#example-167

I just tested up to PHP 7.2 and it works.

> Closures may also inherit variables from the parent scope. Any such variables must be passed to the use language construct.

The parent scope is the scope in which the anonymous function is defined, which I believe is considered different from the instance scope.

If you'd like to keep it as is because the IDE is complaining, SGTM.

michaelbausor: Ok, I see what you're saying. I thought you were saying that we can access the local variables without use, not using `$this`. I do want to avoid accessing the AuthWrapper object from the function, even as `$this`, because that seemed to be the cause of the problem.

PR #164, opened 2018-04-09
Decouple auth and transport

michaelbausor:
This possible change:
- Removes AuthWrapper from the transports, and makes it an option that can be passed to each call
- Adds middleware to inject an AuthWrapper from the client on each call (but as a side benefit, if the user passes one in it won't overwrite, so users can if they want specify an authWrapper per-call)
- Moves AuthWrapper build function into the AuthWrapper class

If we like the high-level idea here, we can then think about moving AuthWrapper out of gax and into the auth library itself. This change could also be cleaned up and submitted independently of the change to update constructor args.

### DIFF
--- a/src/ApiCore/AuthWrapper.php
+++ b/src/ApiCore/AuthWrapper.php
@@ -31,10 +31,14 @@
  */
 namespace Google\ApiCore;
 
+use DomainException;
+use Exception;
 use Google\Auth\ApplicationDefaultCredentials;
 use Google\Auth\Cache\MemoryCacheItemPool;
 use Google\Auth\FetchAuthTokenCache;
 use Google\Auth\FetchAuthTokenInterface;
+use Google\Auth\HttpHandler\Guzzle5HttpHandler;
+use Google\Auth\HttpHandler\Guzzle6HttpHandler;
 use Google\Auth\HttpHandler\HttpHandlerFactory;
 use Psr\Cache\CacheItemPoolInterface;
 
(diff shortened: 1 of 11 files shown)


dwsupplee:
The concept looks great. 👍

src/ApiCore/AuthWrapper.php
```diff
@@ -55,6 +63,63 @@ public function __construct(FetchAuthTokenInterface $fetchAuthTokenInterface, ca
         $this->authHttpHandler = $authHttpHandler;
     }
 
+    /**
+     * @param array $args {
+     *     An array of required and optional arguments.
+     *
+     *     @type CredentialsLoader $credentialsLoader
```
dwsupplee: Should `credentialsLoader` and `authHttpHandler` have "Optional" in the description like some of the other fields?

src/ApiCore/AuthWrapper.php
```diff
@@ -55,6 +63,63 @@ public function __construct(FetchAuthTokenInterface $fetchAuthTokenInterface, ca
...
+     *           `function (RequestInterface $request, array $options) : ResponseInterface`.
+     *     @type bool $enableCaching
+     *           Optional. Enable caching of access tokens. Defaults to true.
+     *     @type CacheItemPoolInterface $authCache
+     *           Optional. A cache for storing access tokens. Defaults to a simple in memory implementation.
+     *     @type array $authCacheOptions
+     *           Optional. Cache configuration options.
+     * }
+     * @return AuthWrapper
+     * @throws \Exception
```
dwsupplee: I believe this should be a `ValidationException`. Also, `ApplicationDefaultCredentials::getCredentials()` throws a `\DomainException`. We may want to catch that and bubble up a `ValidationException` for consistency, WDYT?

src/ApiCore/GapicClientTrait.php
```diff
@@ -34,8 +34,12 @@
 
 use Google\ApiCore\LongRunning\OperationsClient;
 use Google\ApiCore\Middleware\AgentHeaderMiddleware;
+use Google\ApiCore\Middleware\AuthWrapperMiddleware;
 use Google\ApiCore\Middleware\RetryMiddleware;
 use Google\ApiCore\Transport\TransportInterface;
+use Google\Auth\Cache\MemoryCacheItemPool;
```
dwsupplee: These three imports aren't used.

src/ApiCore/Middleware/AuthWrapperMiddleware.php
```diff
@@ -0,0 +1,64 @@
...
+{
+    /** @var callable */
+    private $nextHandler;
+
+    /** @var AuthWrapper */
+    private $authWrapper;
+
+    public function __construct(
+        callable $nextHandler,
+        $authWrapper
```
dwsupplee: `AuthWrapper` typehint could be used here

src/ApiCore/Transport/GrpcTransport.php
```diff
@@ -161,7 +156,9 @@ private function getCallOptions(array $options)
             ? $options['transportOptions']['grpcOptions']
             : [];
 
-        $callOptions += ['call_credentials_callback' => $this->authWrapper->getAuthorizationHeaderCallback()];
+        if (isset($options['authWrapper'])) {
+            $callOptions += ['call_credentials_callback' => $options['authWrapper']->getAuthorizationHeaderCallback()];
```
dwsupplee: Could this be an assignment instead of a union?

```php
$callOptions['call_credentials_callback'] = $options['authWrapper']->getAuthorizationHeaderCallback();
```

src/ApiCore/AuthWrapper.php
michaelbausor: I have done a bit of a restructuring of this method and moved the logic to check for `credentialsLoader` into the `GapicClientTrait`, PTAL

src/ApiCore/AuthWrapper.php
michaelbausor: I have converted to `ValidationException`, PTAL.

bshaffer:
Definitely agree AuthWrapper should be moved to auth library as soon as possible, and sorry we haven't done so earlier.

src/ApiCore/AuthWrapper.php
```diff
@@ -31,13 +31,24 @@
...
+use Psr\Cache\CacheItemPoolInterface;
 
 /**
  * The AuthWrapper object provides a wrapper around a FetchAuthTokenInterface.
  */
 class AuthWrapper
 {
+    use ValidationTrait;
+
     private $fetchAuthTokenInterface;
```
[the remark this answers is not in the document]
michaelbausor: I'm not sure what you mean here - we accept a `FetchAuthTokenInterface` in the constructor

src/ApiCore/AuthWrapper.php
```diff
@@ -79,6 +138,41 @@ public function getAuthorizationHeaderCallback()
...
+    {
+        try {
+            return HttpHandlerFactory::build();
+        } catch (Exception $ex) {
+            throw new ValidationException("Failed to build HttpHandler", $ex->getCode(), $ex);
+        }
+    }
+
+    /**
+     * @param $scopes
```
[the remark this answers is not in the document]
michaelbausor: Done

src/ApiCore/AuthWrapper.php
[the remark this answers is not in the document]
michaelbausor: Hmm ok, I guess I disagree but perhaps I am coming at it from a Java perspective, where this naming is normal (although it is not usual to have `Interface` in the name of the interface). Happy to change it, but then what should we call it, given that we can accept any implementing class?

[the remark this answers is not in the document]
michaelbausor: Re updating the name in the linked code, I agree the type should be `FetchAuthTokenInterface` instead of `CredentialsLoader`. That will be changing in the coming updates to constructor args, so we probably don't need to change it separately.

For the var name, what about `credentialsFetcher`? I wanted to not name it `credentialsLoader` specifically because it doesn't have to be a CredentialsLoader object.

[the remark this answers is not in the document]
dwsupplee: That sounds good! In fact I believe that is what we use in the handwritten code, so it would be great to use the same term.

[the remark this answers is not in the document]
michaelbausor: Updated

michaelbausor:
Opened https://github.com/google/google-auth-library-php/pull/194 to get the ball rolling on what should be included in the auth library. I would still like to move ahead with this PR and https://github.com/googleapis/gax-php/pull/146

PR #173, opened 2018-05-03
Create client options trait

michaelbausor:
This just splits gapicclienttrait partially into a new trait, ClientOptionsTrait. Looking for feedback on whether this sort of change (decomposition into smaller traits) is desirable?


dwsupplee:
In general I am on board with decomposition. In this case, though, it seems as though you would always use `ClientOptionsTrait` and `GapicClientTrait` in tandem, and never exclusively. Are there cases you think where we may end up using them on their own, or is this more to help break things down a bit for the sake of maintainability (if so, I can see the argument for that and would be happy with the change)?

michaelbausor:
Yes, they would always be used together (in fact there is a dependency), and this is more about maintainability.

PR #174, opened 2018-05-04
Option construction fixes

michaelbausor:
Fixes and updates for constructor args

### DIFF
--- a/src/ApiCore/GapicClientTrait.php
+++ b/src/ApiCore/GapicClientTrait.php
@@ -34,7 +34,7 @@ namespace Google\ApiCore;
 
 use Google\ApiCore\LongRunning\OperationsClient;
 use Google\ApiCore\Middleware\AgentHeaderMiddleware;
-use Google\ApiCore\Middleware\AuthWrapperMiddleware;
+use Google\ApiCore\Middleware\CredentialsWrapperMiddleware;
 use Google\ApiCore\Middleware\RetryMiddleware;
 use Google\ApiCore\Transport\GrpcTransport;
 use Google\ApiCore\Transport\RestTransport;
@@ -94,6 +94,45 @@ trait GapicClientTrait
     }
 
     /**
+     * Get default options. This function should be "overridden" by clients using late static
+     * binding to provide default options to the client.
+     *
+     * @return array
+     */
+    private static function getClientDefaults()
+    {
+        return [];
+    }
+
+    protected function buildClientOptions(array $options)
+    {
+        $defaultOptions = self::getClientDefaults() + [
+                'disableRetries' => false,
+                'credentials' => null,
+                'credentialsConfig' => [],
+                'transport' => null,
+                'transportConfig' => [],
+                'gapicVersion' => null,
+                'libName' => null,
+                'libVersion' => null,
+            ];
+        $defaultOptions['transportConfig'] += [
+            'grpc' => [],
+            'rest' => [],
+        ];
+
+        $options += $defaultOptions;
+
+        // For nested arrays, we want to preserve keys in $defaultOptions that are not specified
+        // in $options
+        $options['credentialsConfig'] += $defaultOptions['credentialsConfig'];
+        $options['transportConfig']['grpc']  += $defaultOptions['transportConfig']['grpc'];
+        $options['transportConfig']['rest']  += $defaultOptions['transportConfig']['rest'];
+
+        return $options;
+    }
+
+    /**
      * Configures the GAPIC client based on an array of options.
      *
      * @param array $options {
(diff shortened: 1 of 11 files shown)


src/ApiCore/LongRunning/Gapic/OperationsGapicClient.php
```diff
@@ -116,26 +112,18 @@ class OperationsGapicClient
...
+            'descriptorsConfigPath' => __DIR__.'/../resources/operations_descriptor_config.php',
+            'credentialsConfig' => [
+            ],
             'transportConfig' => [
-                'rest' => [
-                    'restClientConfigPath' => __DIR__.'/../resources/operations_rest_client_config.php',
-                ]
+                'rest' => ['restClientConfigPath' => __DIR__.'/../resources/operations_rest_client_config.php'],
             ],
+            'versionFile' => __DIR__.'/../../../VERSION',
```
dwsupplee: Hmm, this may only apply for the GAPICs living in google-cloud-php. Would we want to omit this here?

src/ApiCore/LongRunning/Gapic/OperationsGapicClient.php
```diff
@@ -1,6 +1,6 @@
 <?php
 /*
- * Copyright 2017 Google LLC
+ * Copyright 2018 Google LLC
```
dwsupplee: No need to update the copyright on this `OperationsClient`

src/ApiCore/GapicClientTrait.php
```diff
@@ -93,6 +93,45 @@ private static function getGapicVersion(array $options)
         return self::$gapicVersion;
     }
 
+    /**
+     * Get default options. This function should be "overridden" by clients using late static
+     * binding to provide default options to the client.
+     *
+     * @return array
+     */
+    private static function getClientDefaults()
```
dwsupplee: I'm debating whether we want to keep this static or not. If we give up the fact it is static we can define it as abstract and have this be an enforceable contract. On the other hand, the only code using this is generated and will always end up implementing the method, so the contract isn't that useful. WDYT?

michaelbausor: I don't have a strong opinion. In other circumstances I would favour making this not static - but it is used in the client constructor, and calling non-static methods from the constructor is also not ideal.

src/ApiCore/GapicClientTrait.php
```diff
@@ -93,6 +93,45 @@ private static function getGapicVersion(array $options)
...
+    {
+        return [];
+    }
+
+    protected function buildClientOptions(array $options)
```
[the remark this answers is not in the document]
michaelbausor: I hit an issue with this method body also, when 'grpc' key is provided in options but is missing from defaultOptions. I have updated with a different method body that is hopefully easier to reason about, and added tests and some comments, PTAL.

src/ApiCore/GapicClientTrait.php
```diff
@@ -93,6 +93,45 @@ private static function getGapicVersion(array $options)
...
+    {
+        $defaultOptions = self::getClientDefaults() + [
+                'disableRetries' => false,
```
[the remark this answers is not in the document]
michaelbausor: Fixed

src/ApiCore/LongRunning/Gapic/OperationsGapicClient.php
```diff
@@ -116,26 +112,18 @@ class OperationsGapicClient
...
-            'authConfig' => null,
-            'transport' => null,
+            'descriptorsConfigPath' => __DIR__.'/../resources/operations_descriptor_config.php',
+            'credentialsConfig' => [
+            ],
             'transportConfig' => [
-                'rest' => [
-                    'restClientConfigPath' => __DIR__.'/../resources/operations_rest_client_config.php',
-                ]
+                'rest' => ['restClientConfigPath' => __DIR__.'/../resources/operations_rest_client_config.php'],
```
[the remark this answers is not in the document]
michaelbausor: Fixed

src/ApiCore/LongRunning/Gapic/OperationsGapicClient.php
michaelbausor: I think there is also a problem that can occur with the path being incorrect for deeper nested generated code. I have filed https://github.com/googleapis/gapic-generator/issues/1978, will fix manually for now.

src/ApiCore/GapicClientTrait.php
dwsupplee: Okay, let's leave as is for now then :).

src/ApiCore/GapicClientTrait.php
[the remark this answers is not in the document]
dwsupplee: Ah, yeah, we would need to keep:

```
$defaultOptions['transportConfig'] += [
    'grpc' => [],
    'rest' => []
];
```

The new updates LGTM.

src/ApiCore/GapicClientTrait.php
```diff
@@ -181,45 +226,42 @@ protected function setClientOptions(array $options)
...
+            'gapicVersion' => $options['gapicVersion'] ?: self::getGapicVersion($options),
         ]);
 
         self::validateFileExists($options['descriptorsConfigPath']);
         $descriptors = require($options['descriptorsConfigPath']);
         $this->descriptors = $descriptors['interfaces'][$this->serviceName];
 
-        $authConfig = $options['authConfig'] ?: [];
-        $this->authWrapper = $this->createAuthWrapper($options['auth'], $authConfig);
+        $this->authWrapper = $this->createCredentialsWrapper($options['credentials'], $options['credentialsConfig']);
```
fiboknacky: Would $this->authWrapper need an update to $this->credentialsWrapper?

michaelbausor: It got this update in the PR merged today: https://github.com/googleapis/gax-php/pull/175/files#diff-c38cfca583ea39046b0c40fcce3dcbe9R246

PR #175, opened 2018-05-08
Response metadata support

michaelbausor:
Refactor GapicClientTrait to support response metadata:
- Restructure paging to allow conversion to middleware
- Add Middleware to support longrunning and paging
- Route unary calls through the startUnaryCall method, which can be overridden in a partial veneer
- Add middleware to support fixed headers and response metadata

Note that this PR depends on https://github.com/googleapis/gax-php/pull/174


michaelbausor:
@dwsupplee Updated, PTAL

src/ApiCore/GapicClientExtensionPointsTrait.php
```diff
@@ -0,0 +1,101 @@
...
+        // Do nothing - this method exists to allow option modification by partial veneers.
+    }
+
+    /**
+     * Set custom options on the client.
+     * 
+     * @param array $options
+     * @access private
+     */
+    protected function setCustomClientOptions($options)
```
michaelbausor: Given that we have `modifyClientOptions`, do you think this is necessary? Or should any custom option work just be handled in `modifyClientOptions`? I am thinking that the answer is yes - at least for now, and we can add it back later if we need it.

dwsupplee: Just the `modify` method SGTM. :)

src/ApiCore/GapicClientTrait.php
```diff
@@ -400,34 +414,14 @@ function (Call $call, array $options) {
...
-     *     Optional arguments
-     *
-     *     @type array $headers [optional] key-value array containing headers
-     *     @type int $timeoutMillis [optional] the timeout in milliseconds for the call
-     *     @type array $transportOptions [optional] transport-specific call options
-     * }
-     *
-     * @return array
-     */
-    protected function configureCallOptions(array $optionalArgs)
```
michaelbausor: What are your thoughts on removing this? It acts to restrict the options that can be passed in by the user at call time - is that something that we want? It means that any custom options will be stripped and not be processed by a custom call stack.

michaelbausor: Very good point - we need to either pluck them, or keep the filter.

So for an example, an argument like `withMetadata` that would be handled by a piece of custom middleware. This is implied by using the `modifyUnaryCall` method, which doesn't have access to optional args - it means that the custom call stack should always add their custom middleware/callable modifications, which should check for the argument during the call rather than when constructing the middleware.

michaelbausor: FYI I have just filed https://github.com/googleapis/gapic-generator/issues/1983 - we need to handle the case when optional fields clash with our supported options.

michaelbausor: Added a new OptionsFilterMiddleware to replace this, PTAL

src/ApiCore/Middleware/OperationsCallable.php
```diff
@@ -0,0 +1,79 @@
...
+    ) {
+        $this->nextHandler = $nextHandler;
+        $this->client = $client;
+        $this->descriptor = $descriptor;
+    }
+
+    public function __invoke(Call $call, array $options)
+    {
+        $next = $this->nextHandler;
+        $client = $this->client;
```
dwsupplee: There is an issue where with a callable you can't instantiate it from class scope, e.g.:

```php
$this->nextHandler($call, $options); // blows up
```

But for the other class level variables (client, descriptor), we can access those without the need for the use statement using `$this`. I think this is preferable and a bit cleaner.

src/ApiCore/Middleware/OperationsCallable.php
```diff
@@ -0,0 +1,79 @@
...
+        $descriptor = $this->descriptor;
+        return $next(
+            $call,
+            $options
+        )->then(function (Message $response) use ($client, $descriptor) {
+            $options = $descriptor + [
+                    'lastProtoResponse' => $response
```
dwsupplee: Indentation is a bit off here.

src/ApiCore/Middleware/ResponseMetadataCallable.php
```diff
@@ -0,0 +1,72 @@
...
+    public function __invoke(Call $call, array $options)
+    {
+        $metadataReceiver = new Promise();
+        $options['metadataCallback'] = function ($metadata) use ($metadataReceiver) {
+            $metadataReceiver->resolve($metadata);
+        };
+        $next = $this->nextHandler;
+        return $next($call, $options)->then(
+            function (Message $response) use ($metadataReceiver) {
+                if ($metadataReceiver->getState() == PromiseInterface::FULFILLED) {
```
dwsupplee: Little nit: Could we use strict equals?

src/ApiCore/Transport/GrpcTransport.php
```diff
@@ -174,16 +174,20 @@ public function startUnaryCall(Call $call, array $options)
         );
 
         $promise = new Promise(
-            function () use ($call, &$promise) {
-                list($response, $status) = $call->wait();
+            function () use ($unaryCall, $options, &$promise) {
+                list($response, $status) = $unaryCall->wait();
 
                 if ($status->code == Code::OK) {
+                    if (isset($options['metadataCallback'])) {
```
dwsupplee: Is this something we would want any transport implementation to have support for? If so, do you think it would be worth adding something that handles this logic to the contract?

src/ApiCore/Page.php
```diff
@@ -58,26 +59,23 @@ class Page implements IteratorAggregate
...
      * @param PageStreamingDescriptor $pageStreamingDescriptor
+     * @param Message $response
      */
     public function __construct(
         Call $call,
         array $options,
         callable $callable,
-        PageStreamingDescriptor $pageStreamingDescriptor
+        PageStreamingDescriptor $pageStreamingDescriptor,
+        $response
```
dwsupplee: We could use a `Message` type hint here.

src/ApiCore/GapicClientTrait.php
```diff
@@ -391,41 +409,23 @@ protected function startCall(
...
+        };
+        $callable = new CredentialsWrapperMiddleware($callable, $this->credentialsWrapper);
+        $callable = new AgentHeaderMiddleware($callable, $this->agentHeaderDescriptor);
+        $callable = new RetryMiddleware($callable, $callConstructionOptions['retrySettings']);
+        $callable = new OptionsFilterMiddleware($callable, [
             'headers',
             'timeoutMillis',
             'transportOptions',
-        ], $optionalArgs);
+            'metadataCallback',
```
dwsupplee: Should we document this alongside `headers`, `timeoutMillis`, etc. or did you want to keep this as a "hidden" feature for now?

src/ApiCore/Middleware/OperationsCallable.php
dwsupplee: Yes, it definitely works. It was a feature added in PHP 5.4.

http://php.net/manual/en/functions.anonymous.php#example-167

I wonder if your IDE is using rules defined for PHP 5.3 or lower?

src/ApiCore/Middleware/OperationsCallable.php
```diff
@@ -0,0 +1,79 @@
...
+
+use Google\ApiCore\Call;
+use Google\ApiCore\LongRunning\OperationsClient;
+use Google\ApiCore\OperationResponse;
+use Google\Protobuf\Internal\Message;
+
+/**
+ * Callable which wraps the response in an OperationResponse object.
+ */
+class OperationsCallable
```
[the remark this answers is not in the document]
dwsupplee: The distinction wasn't clear to me without the explanation. I think I would prefer calling it `Middleware` until we have something more formal.

src/ApiCore/GapicClientTrait.php
```diff
@@ -531,12 +535,50 @@ protected function getPagedListResponse(
...
+        // Do nothing - this method exists to allow option modification by partial veneers.
+    }
+
+    /**
+     * Modify the unary callable.
+     *
+     * @param callable $callable
+     * @access private
+     */
+    protected function modifyUnaryCallable(&$callable)
```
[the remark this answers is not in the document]
dwsupplee: It just throws me off a bit because we call the same variable `callStack` everywhere else. Your call!

PR #179, opened 2018-05-23
Refactor VERSION file lookup

dwsupplee:
This change set addresses the fact we can't always determine the correct path for the VERSION file in the GAPICs when they are generated. This instead looks for the VERSION file by checking the directory above "src".


bshaffer:
LGTM!

src/ApiCore/GapicClientTrait.php
```diff
@@ -86,18 +86,26 @@ public function close()
...
         return self::$gapicVersion;
     }
 
+    private static function getVersionFileContents()
+    {
+        $clientFile = (new \ReflectionClass(__CLASS__))->getFileName();
+        $versionFile = substr(
+            $clientFile,
+            0,
+            strrpos($clientFile, '/src/') ?: strrpos($clientFile, '\\src\\')
```
jdpedrie: would `DIRECTORY_SEPARATOR` work here as well? i.e `strrpos($clientFile, DIRECTORY_SEPARATOR . 'src' . DIRECTORY_SEPARATOR)`

src/ApiCore/GapicClientTrait.php
```diff
@@ -86,18 +86,26 @@ public function close()
...
+        ) . DIRECTORY_SEPARATOR . 'VERSION';
+
+        return @file_get_contents($versionFile) ?: null;
```
jdpedrie: Is it safe to assume that if the VERSION file doesn't exist that the component is simply unreleased? If so, perhaps `dev` or some other indicator of a pre-release component would be preferable to `null`?

tests/ApiCore/Tests/Unit/GapicClientTraitTest.php
```diff
@@ -168,15 +168,10 @@ public function testStartOperationsCall()
...
-        $tmpFile = sys_get_temp_dir() . '/VERSION';
-        file_put_contents($tmpFile, $version);
-        $client = new GapicClientTraitStub();
-        $client->set('gapicVersion', $version, true);
-        $options = ['versionFile' => $tmpFile];
-        $this->assertEquals($version, $client->call('getGapicVersion', [
-            $options
-        ]));
+        $client = new \GapicClientStub();
+        $this->assertEquals($version, $client->call('getGapicVersion', [[]]));
```
jdpedrie: Could you add a test for a situation where the VERSION file does not exist?

src/ApiCore/GapicClientTrait.php
dwsupplee: I don't believe that will always be safe to assume, as this may be run outside the context of google-cloud-php.

PR #181, opened 2018-05-26
Update VERSION, travis and docs

michaelbausor:
- Adds VERSION file to replace version constant
- Adds Version class to access VERSION at runtime
- Updates .travis to use stages
- Adds deploy stage to build and publish docs
- Fixes misc doc errors


src/ApiCore/Testing/MockStubTrait.php
```diff
@@ -268,7 +270,8 @@ public static function create($responseObject, $status = null, $deserialize = nu
      * Creates a sequence such that the responses are returned in order.
      * @param mixed[] $sequence
      * @param callable $deserialize
-     * @return MockTransport
+     * @param Status $finalStatus
+     * @return MockStubTrait
```
jdpedrie: I'm not a big fan of this. Traits cannot be used as type hints, the type returned will not be `MockStubTrait`, and `$res instanceof MockStubTrait` will return false.

Could we say `@return static An instance of the current class type`, or something simiar?

michaelbausor:
@dwsupplee @jdpedrie Addressed feedback, PTAL and let me know if there are any other changes you think we need. Thanks!

PR #193, opened 2018-06-08
Make CredentialsWrapper protected

michaelbausor:
Make CredentialsWrapper protected in GapicClientTrait. We have received a request to make this available from other teams using the generated clients.


michaelbausor:
Done, PTAL

PR #198, opened 2018-07-13
Temporary workaround for protobuf extension issue

tmatsuo:
https://github.com/google/protobuf/issues/4761

### DIFF
--- a/src/ApiCore/Testing/GeneratedTest.php
+++ b/src/ApiCore/Testing/GeneratedTest.php
@@ -35,10 +35,66 @@ use Google\ApiCore\Serializer;
 use Google\Protobuf\DescriptorPool;
 use Google\Protobuf\Internal\Message;
 use Google\Protobuf\Internal\RepeatedField;
+use PHPUnit\Framework\Exception;
 use PHPUnit\Framework\TestCase;
 
 abstract class GeneratedTest extends TestCase
 {
+    // Workaround for https://github.com/google/protobuf/issues/4761
+    public static function assertSame($expected, $actual, $message = '')
+    {
+        if ($expected instanceof \Google\Protobuf\Internal\Message) {
+            if (!$actual instanceof \Google\Protobuf\Internal\Message) {
+                throw new Exception("not \\Google\\Protobuf\\Internal\\Message object");
+                return;
+            }
+            parent::assertSame(spl_object_hash($expected), spl_object_hash($actual), $message);
+        }
+        parent::assertSame($expected, $actual, $message);
+    }
+
+    // Workaround for https://github.com/google/protobuf/issues/4761
+    public static function assertEquals($expected, $actual, $message = '', $delta = 0, $maxDepth = 10, $canonicalize = false, $ignoreCase = false)
+    {
+        if (is_array($expected)){
+            if (count($expected) === 0) {
+                parent::assertEquals($expected, $actual, $message);
+                return;
+            }
+            if ($expected[0] instanceof \Google\Protobuf\Internal\Message) {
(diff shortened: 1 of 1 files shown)


tmatsuo:
@dwsupplee Can you take a look?

dwsupplee:
Instead of completely overriding the assertion, I think we could write a custom comparator for this. It should help reduce some of this down, as you won't need to worry about the array checks - for example.

WDYT?

```php
<?php

use Google\Protobuf\Internal\Message;
use SebastianBergmann\Comparator\Comparator;

class ProtobufMessageComparator extends Comparator
{
    /**
     * Returns whether the comparator can compare two values.
     *
     * @param mixed $expected The first value to compare
     * @param mixed $actual The second value to compare
     * @return boolean
     */
    public function accepts($expected, $actual)
    {
       return $expected instanceof Message && $actual instanceof Message;
    }

    /**
     * Asserts that two values are equal.
     *
     * @param Message $expected The first value to compare
     * @param Message $actual The second value to compare
     * @param float|int $delta The allowed numerical distance between two values to
     *                      consider them equal
     * @param  bool $canonicalize If set to TRUE, arrays are sorted before
     *                             comparison
     * @param  bool $ignoreCase If set to TRUE, upper- and lowercasing is
     *                           ignored when comparing string values
     * @throws \PHPUnit_Framework_ComparisonFailure Thrown when the comparison
     *                           fails. Contains information about the
     *                           specific errors that lead to the failure.
     */
    public function assertEquals($expected, $actual, $delta = 0, $canonicalize = FALSE, $ignoreCase = FALSE)
    {
        // Perform the comparison logic here
    }
}
```


In the bootstrap file for the tests then we add the following:

```php
\SebastianBergmann\Comparator\Factory::getInstance()->register(new ProtobufMessageComparator());
```

tmatsuo:
@dwsupplee I agree that is a better approach. I'll add a Comparator.

dwsupplee:
> If there's no easy way to override the exporter, I would say the original PR is a reasonable option for a temporary fix although it's ugly.

It doesn't look like there is unfortunately :(. Agreed with this being okay to move forward with then.

src/ApiCore/Testing/GeneratedTest.php
```diff
@@ -35,10 +35,66 @@
...
 use Google\Protobuf\Internal\RepeatedField;
+use PHPUnit\Framework\Exception;
 use PHPUnit\Framework\TestCase;
 
 abstract class GeneratedTest extends TestCase
 {
+    // Workaround for https://github.com/google/protobuf/issues/4761
+    public static function assertSame($expected, $actual, $message = '')
+    {
+        if ($expected instanceof \Google\Protobuf\Internal\Message) {
```
dwsupplee: This is already imported, could we use the relative name?

src/ApiCore/Testing/GeneratedTest.php
```diff
@@ -35,10 +35,66 @@
...
+            if (!$actual instanceof \Google\Protobuf\Internal\Message) {
+                throw new Exception("not \\Google\\Protobuf\\Internal\\Message object");
+                return;
```
dwsupplee: We shouldn't need to return here when an exception is thrown (applies to a few other instances below as well)

tmatsuo:
@dwsupplee 
On the second thought, maybe I can have a custom ArrayComparator which doesn't try to export the Message objects. Let me try that path.

tmatsuo:
@dwsupplee The custom ArrayComparator works for me. I'd close this and open a new PR for that

PR #201, opened 2018-08-08
Preparatory refactoring to add grpc-fallback support

michaelbausor:
This PR is intended to prepare the ground for adding support for the grpc-fallback transport (serial protobuf over HTTP 1.1). It:
- Adds ApiException::createFromRpcStatus
- Refactors functionality from RestTransport into GuzzleTransportTrait
- No behaviour changes

### DIFF
--- a/src/ApiCore/ApiException.php
+++ b/src/ApiCore/ApiException.php
@@ -96,8 +96,12 @@ class ApiException extends Exception
      * @param \Exception $previous
      * @return ApiException
      */
-    public static function createFromApiResponse($basicMessage, $rpcCode, $metadata = null, $previous = null)
-    {
+    public static function createFromApiResponse(
+        $basicMessage,
+        $rpcCode,
+        array $metadata = null,
+        \Exception $previous = null
+    ) {
         $rpcStatus = ApiStatus::statusFromRpcCode($rpcCode);
 
         $messageData = [
--- a/src/ApiCore/ApiException.php
+++ b/src/ApiCore/ApiException.php
@@ -32,6 +32,7 @@
 namespace Google\ApiCore;
 
 use Exception;
+use Google\Rpc\Status;
 
 /**
  * Represents an exception thrown during an RPC.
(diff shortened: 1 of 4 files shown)


src/ApiCore/Transport/GuzzleTransportTrait.php
```diff
@@ -0,0 +1,119 @@
...
+
+use Exception;
+use Google\ApiCore\Call;
+use Google\ApiCore\ValidationException;
+use Google\Auth\HttpHandler\HttpHandlerFactory;
+
+/**
+ * An trait for shared functionality between Guzzle HTTP-based transports.
+ */
+trait GuzzleTransportTrait
```
dwsupplee: Since guzzle is the default transport and users could supply another - WDYT about naming this something like `HTTP11TransportTrait`?

michaelbausor: Hmm, yep you're right. I'm not that keen on have `11` in the name (purely for aesthetic reasons). Do you think `HttpTransportTrait` is misleading?

dwsupplee: Agreed there - I have the same reservations. However, `HttpTransportTrait` does seem misleading as it is too broad. I've done a scan of the HTTP/1.1 protocol memo and didn't have any luck finding any other suitable names we could swap in. Pondering alternatives..

michaelbausor: What about 'HttpUnaryTransportTrait'? That is trying to indicate that this is suitable for transports that support only simple unary rpc methods, not streaming.

dwsupplee: I like it!

michaelbausor: Updated, PTAL

src/ApiCore/ApiException.php
```diff
@@ -91,10 +92,11 @@ public static function createFromStdClass($status)
     /**
      * @param string $basicMessage
      * @param int $rpcCode
-     * @param string|null $metadata
+     * @param array|null $metadata
+     * @param \Exception $previous
      * @return ApiException
      */
-    public static function createFromApiResponse($basicMessage, $rpcCode, $metadata = null)
+    public static function createFromApiResponse($basicMessage, $rpcCode, $metadata = null, $previous = null)
```
dwsupplee: We could typehint `metadata` and `previous`

src/ApiCore/Transport/HttpUnaryTransportTrait.php
```diff
@@ -0,0 +1,120 @@
...
+ */
+namespace Google\ApiCore\Transport;
+
+use Exception;
+use Google\ApiCore\Call;
+use Google\ApiCore\ValidationException;
+use Google\Auth\HttpHandler\HttpHandlerFactory;
+
+/**
+ * An trait for shared functionality between transports that support only unary RPCs using simple
```
dwsupplee: `A trait` and instead of a slash could we end with a period?

Issue #197, opened 2018-06-25
Unexpected error in GAX JISON parser for Firestore document path.


jdpedrie:
See googlecloudplatform/google-cloud-php#1119 for context.

The parse methods in GAX do not accept various symbols which are valid values in Firestore document names.

`*`
`=`
`}`
`{`

The presence of any of these symbols in arguments given to `FirestoreGapicClient::documentPathName()` raises the following error:

```
Fatal error: Uncaught Exception: Parse error on line 1:
Kind/F{[<,>.?509880146
------^
Expecting FORWARD_SLASH, COLON, RIGHT_BRACE, got 'LEFT_BRACE'
```

Is there any possibility that the parser may be able to support these symbols?

michaelbausor:
Closing this since support was released in gax 0.37. If there are still issues once google-cloud-php moves to 0.37, please feel free to reopen.

PR #203, opened 2018-08-09
Replace Jison implementation for PathTemplates

michaelbausor:
Important changes in this PR:
- Removes Jison (YAY!)
- Adds new implementation for resource templates under the ResourceTemplate directory
  - The new implementation makes a hard distinction between absolute resource templates (must have leading slash, may have verb) and relative resource templates (never leading slash, never verb)
- PathTemplate is refactored to use this new implementation, but retains the ability to contain either an absolute or relative template
- Makes the following **breaking** changes in PathTemplate:
  - Removes `count` method and no longer implements `countable` interface
  - Retains leading '/' if it was present in the original `$path` parsed (previously this was dropped)
  - No longer support verbs that contain '/'
- Minor changes to PathTemplate
  - Exception messages
  - String representation of `*` no longer changed to `{$0=*}`
- Makes ValidationException a RuntimeException
- Adds lots of test cases

When reviewing, please pay close attention to `PathTemplateTest`, to make sure that there are no regressions in `PathTemplate` behaviour, other than the expected ones.


jdpedrie:
Cloud Firestore supports special characters in paths which are not allowed by the jison parser or by this new parser. I added a workaround (https://github.com/GoogleCloudPlatform/google-cloud-php/pull/1122) for this a couple of months ago.

When I made that change, we had been hoping it would be a temporary workaround pending this work we knew you had on your agenda. Would it be possible to add support for optional special characters in paths?

See also #197.

src/ApiCore/ResourceTemplate/Parser.php
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    public static function isValidLiteral($literal)
+    {
+        return preg_match("/^[0-9a-zA-Z\\.\\-~_]+$/", $literal) === 1;
+    }
+
+    public static function isValidBinding($literal)
+    {
+        return preg_match("-^[^/=\\{\\}\\\\]+$-", $literal) === 1;
```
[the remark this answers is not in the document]
michaelbausor: Looking at https://github.com/googleapis/gax-php/issues/197 it seems like we might want to expand the set of permitted characters for bindings. @jdpedrie is that correct?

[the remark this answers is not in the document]
michaelbausor: @jdpedrie just saw your comment, will reply in main thread

michaelbausor:
@jdpedrie Yes, I think we can add such support. Are there any more detailed requirements? E.g., are there restrictions on the first character of the file name? If so, we would like to enforce them by not allowing the special characters.

michaelbausor:
Updated, PTAL

src/ApiCore/ValidationException.php
```diff
@@ -32,10 +32,11 @@
 namespace Google\ApiCore;
 
 use Exception;
+use RuntimeException;
 
 /**
  * ValidationException represents a local error (i.e. not during an RPC call).
  */
-class ValidationException extends Exception
+class ValidationException extends RuntimeException
```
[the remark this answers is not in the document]
jdpedrie: I agree with your reasoning that `RuntimeException` is preferable to `ApiException`.

src/ApiCore/ValidationException.php
```diff
@@ -32,10 +32,11 @@
 namespace Google\ApiCore;
 
 use Exception;
```
jdpedrie: this import is not needed.

dwsupplee:
Running through a review today :). Really happy to see this!

src/ApiCore/ValidationException.php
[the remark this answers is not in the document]
dwsupplee: There are [some cases](https://github.com/googleapis/gax-php/blob/0de75cc58255a13e00ea87e85d2c0e6ea2b234e2/src/ApiCore/RetrySettings.php#L288) we use the `ValidationException` as a substitute for an `InvalidArgumentException`.  In the [SPL exception hierarchy](http://php.net/manual/en/spl.exceptions.php#spl.exceptions.tree) `InvalidArgumentException` and `RuntimeException` belong to different paths (`LogicException` vs `RuntimeException`). Does it seem like it might be worthwhile to keep that separation in our library as well? Perhaps it could mean adding another exception which we use for invalid type cases?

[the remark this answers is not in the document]
michaelbausor: This is an interesting point. Is it the case that we are substituting for an `InvalidArgumentException` here? Or are we substituting for an `UnexpectedValueException`? Even so, are there other places in our code where we are using `ValidationException` in a way that we should really be using a `LogicException`?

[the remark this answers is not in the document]
dwsupplee: Good point, that example does seem to align more with an `UnexpectedValueException`. [This](https://github.com/googleapis/gax-php/blob/master/src/ApiCore/GapicClientTrait.php#L326-L329) seems to be a better example of what could be a `LogicException` or `InvalidArgumentException`.

Also, I could see an argument for something like [this](https://github.com/googleapis/gax-php/blob/master/src/ApiCore/OperationResponse.php#L207) to be a `BadMethodCallException` (which would fall under `LogicException`). WDYT?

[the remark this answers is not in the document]
michaelbausor: I'm not sure about that first example either. I feel like it falls under the category of "you have provided a bad `$options` argument", which is sort of what `ValidationException` is meant to cover - bad inputs from the user. I do think there is a strong case that the exception in your second example should be something else, but I don't think `BadMethodCallException` - from the decription: "Exception thrown if a callback refers to an undefined method or if some arguments are missing.", but this is more like you have called this method at the wrong time.

Any though, my main feeling is that reasonable people can disagree on this issue :-)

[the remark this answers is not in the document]
michaelbausor: I have reverted the change - I would rather err on the side on not churning. Maybe we can discuss some time next week to see if there are any really strong reasons to make this change (my reasons for making it are fairly weak ones) but I don't want to hold up this PR until then.

src/ApiCore/ResourceTemplate/Parser.php
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    /**
+     * Check if $literal is a valid segment literal. Segment literals may only contain numbers,
+     * letters, and any of the following: .-~_
+     *
+     * @param string $literal
+     * @return bool
+     */
+    public static function isValidLiteral($literal)
```
dwsupplee: From what I can tell, this looks to only be used in this class. Could we mark it private until otherwise needed?

src/ApiCore/ResourceTemplate/Parser.php
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    /**
+     * Check if $binding is a valid segment binding. Segment bindings may contain any characters
+     * except a forward slash ('/'), and may not be empty.
+     *
+     * @param $binding
+     * @return bool
+     */
+    public static function isValidBinding($binding)
```
dwsupplee: Could we group the public methods together? It just makes it nice and easy to scan the class and get an idea of what is exposed.

src/ApiCore/ResourceTemplate/RelativeResourceTemplate.php
```diff
@@ -0,0 +1,296 @@
...
+     * @throws ValidationException
+     */
+    public function __construct($path)
+    {
+        if (empty($path)) {
+            throw new ValidationException("Cannot construct RelativeResourceTemplate from empty string");
+        }
+        $this->segments = Parser::parseSegments($path);
+
+        $doubleWildcardCount = self::countDoubleWildcards($this->segments);
```
dwsupplee: Are there use cases where we would want a list of segments that have more than 1 double wildcard? If not, do you think it would be worthwhile to track the wildcards and throw an exception within `Parser::parseSegments()` as we're building the list?

src/ApiCore/ResourceTemplate/RelativeResourceTemplate.php
```diff
@@ -0,0 +1,296 @@
...
+    private function matchException($path)
+    {
+        return new ValidationException("Could not match path '$path' to template '$this'");
+    }
+
+    /**
+     * @param Segment[] $segments
+     * @return array[] A list of [string, Segment] tuples
+     */
+    private static function buildKeySegmentTuples($segments)
```
dwsupplee: We could use an array typehint here

src/ApiCore/ResourceTemplate/RelativeResourceTemplate.php
```diff
@@ -0,0 +1,296 @@
...
+            }
+        }
+        return $keySegmentTuples;
+    }
+
+    /**
+     * @param array[] A list of [string, Segment] tuples
+     * @return array[] A list of [string, Segment] tuples
+     */
+    private static function flattenKeySegmentTuples($keySegmentTuples)
```
dwsupplee: ditto

src/ApiCore/ResourceTemplate/RelativeResourceTemplate.php
```diff
@@ -0,0 +1,296 @@
...
+            }
+        }
+        return $flattenedKeySegmentTuples;
+    }
+
+    /**
+     * @param Segment[] $segments
+     * @return int
+     */
+    private static function countDoubleWildcards($segments)
```
dwsupplee: ditto

src/ApiCore/ResourceTemplate/RelativeResourceTemplate.php
```diff
@@ -0,0 +1,296 @@
...
+            /** @var Segment $segment */
+            switch ($segment->getSegmentType()) {
+                case Segment::VARIABLE_SEGMENT:
+                    // For segment variables, replace the segment with the segments of its children
+                    $template = $segment->getTemplate();
+                    $nestedKeySegmentTuples = self::buildKeySegmentTuples($template->segments);
+                    foreach ($nestedKeySegmentTuples as list($nestedKey, $nestedSegment)) {
+                        /** @var Segment $nestedSegment */
+                        // Nested variables are not allowed
+                        assert($nestedSegment->getSegmentType() !== Segment::VARIABLE_SEGMENT);
```
dwsupplee: It's possible for users to disable/modify the behavior of `assert` calls. If the intent here is to prevent further execution, we may need to throw an exception. (This applies to other usage of assert as well)

src/ApiCore/ResourceTemplate/RelativeResourceTemplate.php
```diff
@@ -0,0 +1,296 @@
...
+        assert(count($flattenedKeySegmentTuples) > 0);
+
+        // We would like to match pieces of our path 1:1 with the segments of our template. However,
+        // this is confounded by the presence of double wildcards ('**') in the template, which can
+        // match multiple segments in the path.
+        // Because there can only be one '**' present, we can determine how many segments it must
+        // match by examining the difference in count between the template segments and the
+        // path pieces.
+
+        if (count($pathPieces) < count($flattenedKeySegmentTuples)) {
```
dwsupplee: WDYT about computing the counts once and re-using the cached value?

src/ApiCore/ResourceTemplate/Segment.php
```diff
@@ -0,0 +1,159 @@
...
+    /** @var string|null */
+    private $key;
+
+    /** @var string|null */
+    private $value;
+
+    /** @var RelativeResourceTemplate|null */
+    private $template;
+
+    public function __construct($segmentType, $key, $value, RelativeResourceTemplate $template = null)
```
dwsupplee: Since this is internal.. no big deal - but just a note that it may be handy to add a docblock since this is public

tests/ApiCore/Tests/Unit/PathTemplateTest.php
```diff
@@ -156,49 +148,47 @@ public function testMatchEscapedChars()
         );
     }
 
+    /**
+     * @expectedException \Google\ApiCore\ValidationException
+     */
     public function testMatchWildcardWithColonInMiddle()
```
dwsupplee: For tests which are expected to fail could we follow the naming others tests use? e.g. `testFailUnexpectedEof`

src/ApiCore/ResourceTemplate/Parser.php
```diff
@@ -0,0 +1,206 @@
...
+
+/**
+ * Collection of methods for parsing Segments.
+ */
+class Parser
+{
+    /**
+     * Parses a path into an array of segments.
+     *
+     * @param $path
```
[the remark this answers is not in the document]
michaelbausor: Fixed in https://github.com/googleapis/gax-php/pull/209

src/ApiCore/ResourceTemplate/Parser.php
```diff
@@ -0,0 +1,206 @@
...
+            $segmentString = substr($path, $index, $nextSlash - $index);
+            $index = $nextSlash;
+            return self::parse($segmentString);
+        }
+    }
+
+    private static function parse($segmentString)
+    {
+        if ($segmentString === '*') {
+            return new Segment(Segment::WILDCARD_SEGMENT, null, null);
```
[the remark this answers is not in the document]
michaelbausor: Fixed in https://github.com/googleapis/gax-php/pull/209, rearranged args as well

src/ApiCore/ResourceTemplate/Parser.php
```diff
@@ -0,0 +1,206 @@
...
+    /**
+     * @param string $literal
+     * @param string $path
+     * @param int $index
+     * @return string
+     * @throws ValidationException
+     */
+    private static function parseLiteralFromPath($literal, $path, &$index)
+    {
+        if (strlen($path) < ($index + strlen($literal))) {
```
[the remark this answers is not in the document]
michaelbausor: Fixed in https://github.com/googleapis/gax-php/pull/209

src/ApiCore/ResourceTemplate/Parser.php
michaelbausor: Fixed in https://github.com/googleapis/gax-php/pull/209, also moved `isValidBinding` functions

src/ApiCore/ResourceTemplate/Parser.php
michaelbausor: Fixed in https://github.com/googleapis/gax-php/pull/209

src/ApiCore/ResourceTemplate/RelativeResourceTemplate.php
michaelbausor: I think I would rather leave `parseSegments` as just parsing unconnected segments which must be valid on their own, and then apply check that apply to all segments (wildcards, duplicate keys) afterwards. WDYT?

src/ApiCore/ResourceTemplate/RelativeResourceTemplate.php
michaelbausor: Fixed in https://github.com/googleapis/gax-php/pull/209

src/ApiCore/ResourceTemplate/RelativeResourceTemplate.php
michaelbausor: Fixed in https://github.com/googleapis/gax-php/pull/209

src/ApiCore/ResourceTemplate/RelativeResourceTemplate.php
michaelbausor: Fixed in https://github.com/googleapis/gax-php/pull/209

src/ApiCore/ResourceTemplate/RelativeResourceTemplate.php
michaelbausor: This is deliberate. All of the asserts are things that should never happen, regardless of user input - if they happen, it is a bug in the code. So they are there so be checked during testing, but users do not need them to validate their inputs.

src/ApiCore/ResourceTemplate/RelativeResourceTemplate.php
michaelbausor: Fixed in https://github.com/googleapis/gax-php/pull/209

src/ApiCore/ResourceTemplate/Segment.php
michaelbausor: Fixed in https://github.com/googleapis/gax-php/pull/209

tests/ApiCore/Tests/Unit/PathTemplateTest.php
michaelbausor: Fixed in https://github.com/googleapis/gax-php/pull/209

src/ApiCore/ResourceTemplate/RelativeResourceTemplate.php
dwsupplee: Sounds good, I like that

PR #206, opened 2018-08-14
Refactor AgentHeaderDescriptor

michaelbausor:
- Replace AgentHeaderDescriptor with AgentHeader, a class with only static methods
- In GapicClientTrait, cache only the gapic version loaded from file
- Remove AgentHeaderMiddleware and use FixedHeaderMiddleware in its place


src/ApiCore/AgentHeader.php
```diff
@@ -93,30 +92,22 @@ public function __construct($headerInfo)
...
         return [self::AGENT_HEADER_KEY => [implode(" ", $metricsList)]];
     }
 
-    /**
-     * Returns the version string for ApiCore.
-     *
-     * @return string
-     */
-    public static function getApiCoreVersion()
+    public static function readGapicVersionFromFile($callingClass)
```
tmatsuo: Can you add the doc?

src/ApiCore/Version.php
```diff
@@ -42,12 +42,17 @@ public static function getApiCoreVersion()
...
-            self::$version = trim($versionString);
+            self::$version = self::readVersionFile($versionFile);
         }
         return self::$version;
     }
 
+    public static function readVersionFile($file)
+    {
+        $versionString = @file_get_contents($file) ?: "";
+        return trim($versionString);
```
tmatsuo: If for some reason there's no `VERSION` file, it will return an empty string.

Are you sure that an empty string will work without problem?
How about to return "Unknown-Version" or sth?

src/ApiCore/AgentHeader.php
michaelbausor: Added

src/ApiCore/Version.php
michaelbausor: Hmm, I think I prefer empty string rather than some placeholder. We want to keep the possibility of testing for the version being missing. What about returning null? That has disadvantages as well (it is even more likely to not work somewhere).

Another alternative - have the function take a second argument, $defaultValue, that is the value to use if no version file is found. WDYT?

tmatsuo: Alright, an empty string works for me.

tmatsuo:
Looks like there's a cs error, feel free to merge once it's fixed

PR #207, opened 2018-08-15
Add support to enable grpc_call_invoker

WeiranFang:
Per discussion in https://github.com/GoogleCloudPlatform/google-cloud-php/pull/1146, this PR is to move the logic of enabling channel management to gax library. Another part of the changes resides in gapic-generator library: https://github.com/googleapis/gapic-generator/pull/2231


tests/ApiCore/Tests/Unit/GapicClientTraitTest.php
```diff
@@ -386,6 +386,7 @@ public function buildClientOptionsProvider()
             'serviceName' => 'test.interface.v1.api',
             'clientConfig' => __DIR__ . '/testdata/test_service_client_config.json',
             'descriptorsConfigPath' => __DIR__.'/testdata/test_service_descriptor_config.php',
+            'gcpApiConfigPath' => __DIR__.'/../testdata/test_service_grpc_config.json',
```
michaelbausor: Maybe I am misunderstanding the change, but shouldn't we also need to add ['transportConfig']['grpc']['stubOpts'] to this?

src/ApiCore/GapicClientTrait.php
```diff
@@ -162,6 +172,26 @@ private function buildClientOptions(array $options)
             'rest' => [],
         ];
 
+        if (isset($options['gcpApiConfigPath'])
+                && file_exists($options['gcpApiConfigPath'])
+                && isset($options['serviceAddress'])) {
+            $grpcGcpConfig = initGrpcGcpConfig(
```
michaelbausor: `$this->initGrpcGcpConfig`?

WeiranFang: Thx. Seems like this line is not being covered in tests, I'll double check

tests/ApiCore/Tests/Unit/GapicClientTraitTest.php
```diff
@@ -666,6 +667,7 @@ public static function getClientDefaults()
...
```
ZhouyihaiDing: Seems like the path has an extra `/..`?

WeiranFang: Thanks! This is why those lines are not being triggered. Fixed.

tests/ApiCore/Tests/Unit/GapicClientTraitTest.php
WeiranFang: Done.

src/ApiCore/GapicClientTrait.php
WeiranFang: Done.

composer.json
```diff
@@ -8,7 +8,7 @@
     "require": {
         "php": ">=5.5",
         "google/auth": "^1.2.0",
-	"google/gpc-gcp": "^0.1.0",
+	"google/grpc-gcp": "^0.1.0",
```
ZhouyihaiDing: LGTM.
Just a nit, 4 spaces here.

WeiranFang: done. vim issue

src/ApiCore/GapicClientTrait.php
```diff
@@ -162,6 +172,26 @@ private function buildClientOptions(array $options)
             'rest' => [],
         ];
 
+        if (isset($defaultOptions['gcpApiConfigPath'])
```
michaelbausor: Is it intentional that this acts on `$defaultOptions` instead of on `$options`? I would actually expect this logic to live somewhere like line 312. That way, if a user passes in a value of `gcpApiConfigPath`, or if an API updates the value by overriding `modifyClientOptions()`, then that value would be used. With this implementation, only the value defined in `getClientDefaults()` in the generated client will ever be used, with no way to update it.

WeiranFang: I use $defaultOptions because the actual 'gcpApiConfigPath' is initialized in getClientDefaults. See https://github.com/googleapis/gapic-generator/pull/2231 . It should be defined in the client library, we don't actually want the user to pass in this value.

As for the modifyClientOptions(), I think you are right. The client should be able to use this method to override the path value.

Thanks!

WeiranFang: After reading the code, I think a better place to put this block of logic is line 175, right before the `buildClientOptions()` returns. Because to me `setClientOptions()` is more like a read only operations on options array, but we are adding a new key to that array. So I would rather keep this logic inside `buildClientOptions()` but after `modifyClientOptions()`.

michaelbausor: SGTM

WeiranFang: done.

src/ApiCore/GapicClientTrait.php
```diff
@@ -160,8 +170,30 @@ private function buildClientOptions(array $options)
         $options['transportConfig'] += $defaultOptions['transportConfig'];
         $options['transportConfig']['grpc'] += $defaultOptions['transportConfig']['grpc'];
         $options['transportConfig']['rest'] += $defaultOptions['transportConfig']['rest'];
-
+
```
dwsupplee: Extra whitespace introduced here

tests/ApiCore/Tests/Unit/GapicClientTraitTest.php
```diff
@@ -377,17 +379,27 @@ public function testBuildClientOptions($options, $expectedUpdatedOptions)
 
     public function buildClientOptionsProvider()
     {
+        $apiConfig = new ApiConfig();
+        $apiConfig->mergeFromJsonString(
+            file_get_contents(__DIR__.'/testdata/test_service_grpc_config.json'));
```
dwsupplee: Could we please put the closing parenthesis on a newline?

```php
$apiConfig->mergeFromJsonString(
    file_get_contents(__DIR__.'/testdata/test_service_grpc_config.json')
);
```

src/ApiCore/GapicClientTrait.php
```diff
@@ -160,8 +170,30 @@ private function buildClientOptions(array $options)
         $options['transportConfig'] += $defaultOptions['transportConfig'];
         $options['transportConfig']['grpc'] += $defaultOptions['transportConfig']['grpc'];
         $options['transportConfig']['rest'] += $defaultOptions['transportConfig']['rest'];
-
+        
         $this->modifyClientOptions($options);
+
+        if (isset($options['gcpApiConfigPath'])
```
dwsupplee: Actually, one more small note: WDYT about adding a check for semaphores here before enabling the optimizer, like in [the PR on google-cloud-php](https://github.com/GoogleCloudPlatform/google-cloud-php/pull/1146/files#diff-a759887e9614192a96db41b91bc02147R173)?

src/ApiCore/GapicClientTrait.php
WeiranFang: done.

tests/ApiCore/Tests/Unit/GapicClientTraitTest.php
WeiranFang: done. Habits for Python :)

src/ApiCore/GapicClientTrait.php
WeiranFang: done. Nice find!

dwsupplee: Thanks! :)

Do you think we might need to update the test case a bit now too? I have a suspicion if someone running the tests doesn't have semaphores support installed they may see failures.

WeiranFang: Thank you David. Actually I'm not quite familiar with this, can you tell me a bit more on how to test this? Do we want to create a test environment with that extension being disabled? Or just simply mock the flag to be false?

dwsupplee: I would be satisfied for now with marking the test as skipped if the extension isn't available.

```php
if (!extension_loaded('sysvshm')) {
    $this->markTestSkipped('The sysvshm extension must be installed to execute this test.');
}
```

WDYT?

WeiranFang: done.

PR #208, opened 2018-08-15
Add support for grpc-fallback

michaelbausor:
Adds support for a new transport, grpc-fallback, which sends binary protobuf over HTTP 1.1. This new transport is similar to the existing rest transport, but simpler because it does not need to map RPC requests to rest URLs, and thus does not need the rest config file.

This PR only adds support - once this is released and tested, we will switch the default transport when grpc is not available from rest to grpc-fallback.

### DIFF
--- a/src/ApiCore/Transport/GrpcFallbackTransport.php
+++ b/src/ApiCore/Transport/GrpcFallbackTransport.php
@@ -1,6 +1,6 @@
 <?php
 /*
- * Copyright 2018, Google Inc.
+ * Copyright 2018 Google LLC
  * All rights reserved.
  *
  * Redistribution and use in source and binary forms, with or without
--- a/src/ApiCore/Transport/GrpcFallbackTransport.php
+++ b/src/ApiCore/Transport/GrpcFallbackTransport.php
@@ -1,0 +1,200 @@
+<?php
+/*
+ * Copyright 2018, Google Inc.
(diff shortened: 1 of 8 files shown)


src/ApiCore/Transport/GrpcFallbackTransport.php
```diff
@@ -0,0 +1,202 @@
...
+     * }
+     * @return GrpcFallbackTransport
+     * @throws ValidationException
+     */
+    public static function build($serviceAddress, array $config = [])
+    {
+        $config += [
+            'httpHandler'  => null,
+        ];
+        list($baseUri, $port) = self::normalizeServiceAddress($serviceAddress);
```
dwsupplee: I noticed this is the same behavior as in the `RestTransport`. I'm trying to remember why we split the port out here - it seems like it could cause an issue since we aren't passing it down into the constructed transport. We may just want to update line 92 to `return new GrpcFallbackTransport("$baseUri:$port", $httpHandler);`. WDYT?

michaelbausor: Hmm, I'm not sure. Right now we don't support an alternate port for those transports - if we want to pass it into the constructor, then we should also add support by plumbing it into guzzle, right? Maybe this could be flagged as a separate issue to be fixed?

PR #212, opened 2018-08-27
Clean up warnings and docs

michaelbausor:
- Mostly doc updates
- Minor changes to behaviour include exception transformation and type hints


michaelbausor:
This looks like a pretty odd inconsistency between php 5.6 and 7.2 (maybe 7.*?). Filed https://github.com/protocolbuffers/protobuf/issues/5090. I guess I will remove the type hints. Thoughts on if more is required? It seems like all `instanceof Message` checks may be unsafe.

PR #215, opened 2018-09-06
Improve path template error messages

### LINKED ISSUE #214: GAX Parser empty/falsey value throws cryptic exception.
The old JISON parser accepted various values evaluating to `''` for template placeholders. The replacement parser released in v0.37.0 doesn't support empty values.

This strikes me as a good thing, since resource names missing (for instance) a project ID aren't valid, however the error message arising from this case is a bit hard to understand for new users. (see https://github.com/GoogleCloudPlatform/google-cloud-php/issues/1276).

Current behavior:

```php
echo FirestoreGapicClient::databaseRootName(null, '(default)');
// Uncaught Google\ApiCore\ValidationException: Cannot bind segment '{project=*}' to value ''
```

Improved behavior:

```php
echo FirestoreGapicClient::databaseRootName(null, '(default)');
// Uncaught Google\ApiCore\ValidationException: Segment '{project=*}' expected non-empty string, instead got null.
```

michaelbausor:
- Fixes #214
- Improves other messages
- Adds additional test cases

### DIFF
--- a/src/ApiCore/ResourceTemplate/AbsoluteResourceTemplate.php
+++ b/src/ApiCore/ResourceTemplate/AbsoluteResourceTemplate.php
@@ -111,19 +111,22 @@ class AbsoluteResourceTemplate implements ResourceTemplateInterface
      */
     public function match($path)
     {
-        if (empty($path) || $path[0] !== '/') {
-            throw $this->matchException($path);
+        if (empty($path)) {
+            throw $this->matchException($path, "path cannot be empty");
+        }
+        if ($path[0] !== '/') {
+            throw $this->matchException($path, "missing leading '/'");
         }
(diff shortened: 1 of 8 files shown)


michaelbausor:
@dwsupplee @jdpedrie I haven't added test cases that explicitly check the entire exception message. Do you think that is worth doing for some subset of invalid inputs for parsing, rendering and/or matching?

dwsupplee:
I can definitely see some value in validating the exception messages. Were you thinking of just adding the messages to the `invalidPathProvider`?

michaelbausor:
Basically yes. I have added expected messages for some representative tests. PTAL and let me know if you think there are other particular cases that need their own messages, or it these cases cover most of the bases.

PR #220, opened 2018-12-10
Remove common protos

michaelbausor:
Removes common protos and adds dependency on google/common-protos


composer.json
```diff
@@ -21,14 +22,7 @@
     },
     "autoload": {
         "psr-4": {
-            "Google\\Api\\": "src/Api",
             "Google\\ApiCore\\": "src/ApiCore",
```
dwsupplee: If you're open to it, we should be able to put everything in the `ApiCore` directory right into `src` now and replace this line with `"Google\\ApiCore\\": "src"`

michaelbausor: Was just thinking the same thing actually :-)
Done. PTAL at src/Version, which needed a change.

dwsupplee: Great catch 👍

michaelbausor: Should I do the same in tests?

composer.json
```diff
@@ -12,7 +12,8 @@
         "grpc/grpc": "^1.4",
         "google/protobuf": "^3.5.1",
```
dwsupplee: I noticed [here](https://github.com/googleapis/google-cloud-php/pull/1517/files#diff-ad6978f3626f02acfb16dbe103da2356R9) we are requiring a higher version of protobuf, would there be any harm in doing the same here?

composer.json
dwsupplee: Yes, SGTM :)

composer.json
michaelbausor: SGTM, done

PR #221, opened 2018-12-11
Make PagedListResponse iterable

### LINKED ISSUE #153: Make PagedListResponse Iterable
We should make `PagedListResponse` iterable. This would clean up the code:

```php
$devices = $client->getDevices($parent, $registry);

// current code, not intuitive
foreach ($devices->iterateAllElements() as $device) {
    printf('Device ID: %s' . PHP_EOL, $device->getId());
}

// better code
foreach ($devices as $device) {
    printf('Device ID: %s' . PHP_EOL, $device->getId());
}
```

Another issue is the second one does not throw an error, it just fails to print out anything, which is the same behavior as if it had no elements. So it's very confusing!

michaelbausor:
PTAL, and let me know whether we need more or clearer documentation. cc @sirtorry, @tmatsuo 

Closes #153

### DIFF
--- a/src/PagedListResponse.php
+++ b/src/PagedListResponse.php
@@ -53,7 +53,7 @@ use IteratorAggregate;
  * ```
  * $pagedListResponse = $client->getList(...);
  * foreach ($pagedListResponse as $element) {
- *   // doSomethingWith($element);
+ *     // doSomethingWith($element);
  * }
  * ```
  *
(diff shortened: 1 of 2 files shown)


michaelbausor:
@bshaffer added examples in the class comments - WDYT?

src/PagedListResponse.php
```diff
@@ -43,12 +44,42 @@
...
+ *  - As a single iterable used in a foreach loop or via the getIterator method
  *  - As pages of elements, using the getPage and iteratePages methods
  *  - As fixed size collections of elements, using the
  *    getFixedSizeCollection and iterateFixedSizeCollections methods
+ *
+ * Example of using PagedListResponse as an iterator:
+ * ```
+ * $pagedListResponse = $client->getList(...);
+ * foreach ($pagedListResponse as $element) {
+ *   // doSomethingWith($element);
```
dwsupplee: Could we please use 4 spaces?

bshaffer: agreed, these are great but let's use 4 spaces!!

michaelbausor: Done!

bshaffer:
What a guy!

PR #223, opened 2018-12-13
Add internal tags, remove Message from Call

michaelbausor:
WIP PR to add internal tags and remove references in Call to protobuf Message class, PTAL.

### DIFF
--- a/src/Call.php
+++ b/src/Call.php
@@ -36,6 +36,8 @@ use Google\Protobuf\Internal\Message;
 
 /**
  * Contains information necessary to manage a network request.
+ *
+ * @internal
  */
 class Call
 {
--- a/src/Call.php
+++ b/src/Call.php
@@ -32,12 +32,8 @@
 
 namespace Google\ApiCore;
 
-use Google\Protobuf\Internal\Message;
-
 /**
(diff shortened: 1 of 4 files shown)


src/Call.php
```diff
@@ -53,14 +51,14 @@ class Call
     /**
      * @param string $method
      * @param string $decodeType
-     * @param Message $message
+     * @param mixed $message
```
dwsupplee: WDYT  about `Message|mixed`? Using [this](https://github.com/googleapis/gax-php/blob/master/src/ServerStream.php#L88) as precedent.

michaelbausor:
I don't think we can do that with GapicClientTrait. It is used in generated code. If we change it in a breaking way, then anyone who upgrades their gax version but doesn't upgrade their generated code would be broken. But if we want to change it in a non-breaking way, then we are fine to do that even if it isn't marked internal. Does that make sense?

dwsupplee:
Yeah, it does - good call.

src/Call.php
michaelbausor: Agreed, done.

michaelbausor:
Ready for review, PTAL

PR #237, opened 2019-04-22
Fix JSON empty object encoding

bshaffer:
A request with an empty Message object is being JSON encoded as an array `[]` instead of object `{}`. For instance, this request:

```php
$message = new MockRequestBody();
$message->setName('message/foo');
$anotherMessage = new MockRequestBody();
$message->setNestedMessage($anotherMessage);
```

will output this:

```json
{"name":"message\/foo","nestedMessage":[]}
```
when it should output this:

```json
{"name":"message\/foo","nestedMessage":{}}
```

This fix will work when `body = *` in the rest client config, but will not work if the `body` is a property of the message.

**EDIT**: Added fix so it works when body is a property of the message

### DIFF
--- a/src/RequestBuilder.php
+++ b/src/RequestBuilder.php
@@ -103,7 +103,7 @@ class RequestBuilder
                     $config['method'],
                     $uri,
                     ['Content-Type' => 'application/json'] + $headers,
-                    $body ? json_encode($body) : null
+                    $body
                 );
             }
         }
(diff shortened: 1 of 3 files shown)


src/RequestBuilder.php
```diff
@@ -145,22 +145,22 @@ private function getConfigsForUriTemplates($config)
...
-
+        $messageData = json_decode($messageDataJson, true);
         foreach ($messageData as $name => $value) {
             if (array_key_exists($name, $config['placeholders'])) {
                 continue;
             }
 
             if (Serializer::toSnakeCase($name) === $config['body']) {
-                $body = $value;
+                $body = $message->{"get$name"}()->serializeToJsonString();
```
dwsupplee: I did a little looking around and it appears we should always expect a message type for the body, but as a safety net do you think it would be worth it to check to make sure we are working with a `Message` instance instead of, for example, a `RepeatedField`?

bshaffer: This is a good idea. Something like this?:

```php
if (($bodyMessage = $message->{"get$name"}) instanceof Message) {
    $body = $bodyMessage
} else {
    $body = $value
}
```

dwsupplee: Yeah, that looks nice. We will also need to make sure `$value` gets JSON encoded if it isn't an instance of `Message` though.

bshaffer: PTAL

PR #239, opened 2019-05-30
[DO NOT SUBMIT] Update c extension version

michaelbausor:


jdpedrie:
Can I ask why we need to specify a version? Can't we just let pecl give us the latest?

michaelbausor:
@jdpedrie sorry, I should have flagged the purpose of this PR - I was intending to force travis to use the latest version of the extension to try to figure out when https://github.com/protocolbuffers/protobuf/issues/6195 would cause a failure. This isn't intended to be submitted.

PR #240, opened 2019-06-05
Disable service config resolution by default

noahdietz:
Disables resolution of the gRPC ServiceConfig from DNS TXT Records. Context https://github.com/googleapis/gapic-generator/issues/2778.

### DIFF
--- a/src/GapicClientTrait.php
+++ b/src/GapicClientTrait.php
@@ -172,6 +172,7 @@ trait GapicClientTrait
         $options['credentialsConfig'] += $defaultOptions['credentialsConfig'];
         $options['transportConfig'] += $defaultOptions['transportConfig'];
         $options['transportConfig']['grpc'] += $defaultOptions['transportConfig']['grpc'];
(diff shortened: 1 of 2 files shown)


src/GapicClientTrait.php
```diff
@@ -160,7 +160,7 @@ private function buildClientOptions(array $options)
             'libVersion' => null,
         ];
         $defaultOptions['transportConfig'] += [
-            'grpc' => [],
+            'grpc' => ['stubOpts' => ['grpc.service_config_disable_resolution' => 1]],
```
michaelbausor: Can you add the following line on 175:
```php
$options['transportConfig']['grpc']['stubOpts'] += $defaultOptions['transportConfig']['grpc']['stubOpts'];
```

This will make sure to merge together a user-specified `stubOpts` arg with your default value. Otherwise, their `stubOpts` will overwrite yours, even if they haven't specified a `grpc.service_config_disable_resolution` setting. Note that even with that line added, the user's value of `grpc.service_config_disable_resolution`, if specified, will override the default (which is I think what we want?)

noahdietz: Yes this is what we want. Thanks.

PR #255, opened 2019-09-06
fix: Properly format grpc metadata keys for non-ASCII values

jdpedrie:
Fixes https://github.com/googleapis/google-cloud-php/issues/2297.


tests/Tests/Unit/RequestParamsHeaderDescriptorTest.php
```diff
@@ -71,4 +71,17 @@ public function testMultipleValues()
 
         $this->assertSame($expectedHeader, $header);
     }
-}
\ No newline at end of file
+
+    public function testNonAsciiChars()
```
dwsupplee:
```suggestion
    public function testNonAsciiCharsAppendsBinToHeaderKey()
```

PR #258, opened 2019-10-22
PHP 7.4 to the build matrix

alexeyshockov:


jdpedrie:
Before this can be done, tests will need to be upgraded in a manner similar to [`google/apiclient`#1710](https://github.com/googleapis/google-api-php-client/pull/1710). That work is on my agenda.

bshaffer:
We've moved tests to Github Actions and have added PHP 7.4

PR #268, opened 2020-06-12
[gax-php] fix: update protobuf to 3.12.2

miraleung:


.travis.yml
```diff
@@ -18,7 +18,7 @@ jobs:
       env: PROTOBUF_C_EXT=true
     - php: '7.2'
       install:
-      - travis_retry pecl install protobuf-3.6.1
+      - travis_retry pecl install protobuf-3.12.2
```
dwsupplee, line 21: This line can be left as is, the intent is to make sure `composer install` fails with an incompatible version of the protobuf extension, otherwise LGTM!

miraleung: Done.

PR #272, opened 2020-08-11
feat: default to self-signed JWTs and use updateMetadata for fetching tokens

bshaffer:
This PR implements changes to allow the GAPIC clients to default to self-signed JWTs by doing the following:

1. Adds `defaultScopes` option for GAPIC clients
1. Uses `updateMetadata` instead of `fetchAuthToken` in `CredentialsWrapper` in order to take advantage of the self-signed JWT logic in `google/auth`
1. When the `SERVICE_ADDRESS` constant is defined, the Gapic clients pass this through the auth library as the `$authUri` when fetching an access token.
1. Calls `getAuthorizationHeaderCallback` in `RestTransport` for consistency
1. Deprecates `getBearerString` in favor of `getAuthorizationHeaderCallback`

No behavior should be impacted until we roll out `defaultScope` config values in Gapic clients.

**Important**
Depends on https://github.com/googleapis/google-auth-library-php/pull/306
Will require the updating minimum version of `google/auth`

### DIFF
--- a/src/CredentialsWrapper.php
+++ b/src/CredentialsWrapper.php
@@ -190,11 +190,11 @@ class CredentialsWrapper
         // be passed into the gRPC c extension, and changes have the potential to trigger very
         // difficult-to-diagnose segmentation faults.
 
-        return function () use ($credentialsFetcher, $authHttpHandler, $audience) {
+        return function (array $headers = []) use ($credentialsFetcher, $authHttpHandler, $audience) {
             $token = $credentialsFetcher->getLastReceivedToken();
             if (self::isExpired($token)) {
                 if ($credentialsFetcher instanceof UpdateMetadataInterface) {
-                    return $credentialsFetcher->updateMetadata([], $audience);
+                    return $credentialsFetcher->updateMetadata($headers, $audience);
                 }
                 $token = $credentialsFetcher->fetchAuthToken($authHttpHandler);
                 if (!self::isValid($token)) {
@@ -202,7 +202,10 @@ class CredentialsWrapper
                 }
             }
             $tokenString = $token['access_token'];
-            return empty($tokenString) ? [] : ['authorization' => ["Bearer $tokenString"]];
+            if (!empty($tokenString)) {
+                $headers['authorization'] = ["Bearer $tokenString"];
+            }
+            return $headers;
         };
     }
 
--- a/src/CredentialsWrapper.php
+++ b/src/CredentialsWrapper.php
@@ -39,6 +39,7 @@ use Google\Auth\CredentialsLoader;
 use Google\Auth\FetchAuthTokenCache;
 use Google\Auth\FetchAuthTokenInterface;
 use Google\Auth\GetQuotaProjectInterface;
+use Google\Auth\UpdateMetadataInterface;
 use Google\Auth\HttpHandler\Guzzle5HttpHandler;
 use Google\Auth\HttpHandler\Guzzle6HttpHandler;
 use Google\Auth\HttpHandler\HttpHandlerFactory;
@@ -94,6 +95,9 @@ class CredentialsWrapper
      *           Cache configuration options.
      *     @type string $quotaProject
      *           Specifies a user project to bill for access charges associated with the request.
+     *     @type string[] $defaultScopes
+     *           A string array of default scopes to use when acquiring
+     *           credentials.
      * }
      * @return CredentialsWrapper
      * @throws ValidationException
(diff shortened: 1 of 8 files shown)


src/GapicClientTrait.php
```diff
@@ -75,6 +75,7 @@ trait GapicClientTrait
         Call::CLIENT_STREAMING_CALL => 'startClientStreamingCall',
         Call::SERVER_STREAMING_CALL => 'startServerStreamingCall',
     ];
+    private static $defaultAudience = 'https://' . self::SERVICE_ADDRESS . '/';
```
bshaffer, line 78: This throws exceptions in testing so a `getDefaultAudience` method may be better.

src/GapicClientTrait.php
```diff
@@ -183,6 +184,14 @@ private function buildClientOptions(array $options)
             $options['apiEndpoint'] = $this->pluck('serviceAddress', $options, false);
         }
 
+        // Use default scopes if an API endpoint is set to ensure the "audience"
+        // does not conflict with the custom endpoint.
+        if (empty($options['credentialsConfig']['scopes'])
+                && $options['apiEndpoint'] != $defaultOptions['apiEndpoint']
+                && isset($options['defaultScopes'])) {
+            $options['credentialsConfig']['scopes'] = $options['defaultScopes'];
```
bshaffer, line 192: `defaultScopes` should be passed for GCE credentials and UserRefresh credentials, so we most likely need to move this logic into `google/auth`

src/Transport/HttpUnaryTransportTrait.php
```diff
@@ -92,10 +92,13 @@ private static function buildCommonHeaders(array $options)
 
         // If not already set, add an auth header to the request
         if (!isset($headers['Authorization']) && isset($options['credentialsWrapper'])) {
-            $bearerString = $options['credentialsWrapper']->getBearerString();
-            if (!empty($bearerString)) {
-                $headers['Authorization'] = $bearerString;
```
bshaffer, line 97: The `authorization` header is lowercase in `updateMetadata` and uppercase here, so we should ensure this change is OK

src/GapicClientTrait.php
```diff
@@ -183,6 +183,16 @@ private function buildClientOptions(array $options)
             $options['apiEndpoint'] = $this->pluck('serviceAddress', $options, false);
         }
 
+        // Set scopes if an API endpoint is set to ensure the "audience"
+        // does not conflict with the custom endpoint.
+        if (empty($options['credentialsConfig']['scopes'])
+                && $options['apiEndpoint'] != $defaultOptions['apiEndpoint']
+                && isset($options['defaultScopes'])) {
+            $options['credentialsConfig']['scopes'] = $options['defaultScopes'];
```
bshaffer, line 191: This ensures when `apiEndpoint` option has been changed, self-signed JWTs are not used. We will be able to remove this once mTLS accepts standard audience strings.

bshaffer:
@dwsupplee @jdpedrie please take a look, this is ready for review / merge!

src/CredentialsWrapper.php
```diff
@@ -167,19 +178,38 @@ public function getBearerString()
...
-            $token = self::getToken($credentialsFetcher, $authHttpHandler);
-            return empty($token) ? [] : ['authorization' => ["Bearer $token"]];
+
+        return function (array $headers = []) use ($credentialsFetcher, $authHttpHandler, $audience) {
+            $token = $credentialsFetcher->getLastReceivedToken();
+            if (self::isExpired($token)) {
+                // Call updateMetadata to take advantage of self-signed JWTs
+                if ($credentialsFetcher instanceof UpdateMetadataInterface) {
+                    return $credentialsFetcher->updateMetadata($headers, $audience);
+                }
```
bshaffer, line 199: Although this looks like a lot, I did try to heed the comment above regarding segfaults. This code only changese the previous behavior to use `updatMetadata` instead of `fetchAuthToken` if the fetcher allows it. And this only has the effect of using self-signed JWTs when available.

This SHOULD be tested with protobuf/grpc extensions enabled! Here is the code I've written to do this, although I do not see a good way in the existng test suites to verify this behavior:

```php
require __DIR__ . '/vendor/autoload.php';

use Google\ApiCore\Call;
use Google\ApiCore\CredentialsWrapper;
use Google\ApiCore\Transport\GrpcTransport;
use Google\ApiCore\Testing\MockRequest;
use Grpc\ChannelCredentials;

$opts = ['credentials' => ChannelCredentials::createInsecure()];
$wrapper = CredentialsWrapper::build();
$transport = new GrpcTransport(
    $hostname = 'storage.googleapis.com',
    $opts
);

$call = new Call(
    'method',
    null,
    new MockRequest()
);

$transport->startUnaryCall($call, [
    'audience' => 'foo.com',
    'credentialsWrapper' => $wrapper
])->wait();
```

dwsupplee (approved):

PR #273, opened 2020-08-19
feat: add quota_project_id support

dwsupplee:
Depends upon: https://github.com/googleapis/google-auth-library-php/pull/299

### DIFF
--- a/src/CredentialsWrapper.php
+++ b/src/CredentialsWrapper.php
@@ -38,6 +38,7 @@ use Google\Auth\Cache\MemoryCacheItemPool;
 use Google\Auth\CredentialsLoader;
 use Google\Auth\FetchAuthTokenCache;
 use Google\Auth\FetchAuthTokenInterface;
+use Google\Auth\GetQuotaProjectInterface;
 use Google\Auth\HttpHandler\Guzzle5HttpHandler;
 use Google\Auth\HttpHandler\Guzzle6HttpHandler;
 use Google\Auth\HttpHandler\HttpHandlerFactory;
@@ -147,11 +148,13 @@ class CredentialsWrapper
     }
 
     /**
-     * @return string The quota project associated with the credentials.
+     * @return string|null The quota project associated with the credentials.
      */
     public function getQuotaProject()
     {
-        return $this->credentialsFetcher->getQuotaProject();
+        if ($this->credentialsFetcher instanceof GetQuotaProjectInterface) {
+            return $this->credentialsFetcher->getQuotaProject();
+        }
     }
 
     /**
--- a/src/CredentialsWrapper.php
+++ b/src/CredentialsWrapper.php
@@ -91,6 +91,8 @@ class CredentialsWrapper
      *           A cache for storing access tokens. Defaults to a simple in memory implementation.
      *     @type array $authCacheOptions
      *           Cache configuration options.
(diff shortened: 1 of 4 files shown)


src/CredentialsWrapper.php
```diff
@@ -136,6 +146,14 @@ public static function build(array $args = [])
         return new CredentialsWrapper($loader, $authHttpHandler);
     }
 
+    /**
+     * @return string The quota project associated with the credentials.
+     */
+    public function getQuotaProject()
+    {
+        return $this->credentialsFetcher->getQuotaProject();
```
bshaffer, line 154: We should upgrade the minimal `google/auth` version to [^1.9](https://github.com/googleapis/google-auth-library-php/releases/tag/v1.9.0) so that this cannot throw a fatal error

bshaffer: Or even better, add an interface check:

```php
public function getQuotaProject()
{
    if ($this->credentialsFetcher instanceof \Google\Auth\GetQuotaProjectInterface) {
        return $this->credentialsFetcher->getQuotaProject();
    }
}
```

bshaffer: I can approve once this is added!

dwsupplee: Done!

bshaffer:
The same error is happening 4 times:

```
Failed asserting that two objects are equal.
--- Expected
+++ Actual
@@ @@
 Google\ApiCore\CredentialsWrapper Object (
     'credentialsFetcher' => Google\Auth\FetchAuthTokenCache Object (
         'fetcher' => Google\Auth\Credentials\GCECredentials Object (...)
         'cacheConfig' => Array (...)
         'cache' => Google\Auth\Cache\MemoryCacheItemPool Object (
-            'items' => Array (...)
+            'items' => null
             'deferredItems' => null
         )
         'maxKeyLength' => 64
     )
     'authHttpHandler' => Google\Auth\HttpHandler\Guzzle7HttpHandler Object (...)
 )
```

I have been looking at the changes and I can't figure out what would be causing this. I ran the tests on `master` just to make sure it wasn't an upstream change that is causing this, and the tests passed.

bshaffer:
I ran the tests on this branch and they passed as well, so it's a true mystery. I'm now restarting the builds to see if it's intermittent...

bshaffer (approved):

tests/Tests/Unit/CredentialsWrapperTest.php
```diff
@@ -47,20 +47,26 @@
...
-     * @dataProvider buildData
+     * @dataProvider buildDataWithoutExplicitKeyFile
      */
-    public function testBuild($args, $expectedCredentialsWrapper)
+    public function testBuildWithoutExplicitKeyFile($args, $expectedCredentialsWrapper)
     {
+        putenv('GOOGLE_APPLICATION_CREDENTIALS=' . __DIR__ . '/testdata/json-key-file.json');
         $actualCredentialsWrapper = CredentialsWrapper::build($args);
         $this->assertEquals($expectedCredentialsWrapper, $actualCredentialsWrapper);
+        putenv('GOOGLE_APPLICATION_CREDENTIALS');
```
bshaffer, line 57: Would it be better to run this in a separate process to make sure the env modifications don't effect other tests?

PR #275, opened 2020-09-03
fix: update gax-php license headers

miraleung:


chrisdunelm:
Are these dates meant to be updated? In other repos, I was under the impression that they're meant to permanently be the year in which the file was first created?

dwsupplee:
I agree with @chrisdunelm's assessment, not sure we want to do this.

miraleung:
SG, reverting.

Issue #264, opened 2020-06-01
Composer V2.0 Deprecation Warning for PSR-4 Namespace


DynamoGeek:
Composer reports that a file in the Testing directory have namespaces that aren't PSR-4 compliant. This can be solved from the user side in the short term in various ways, but the long term fix will need to be making the namespaces PSR-4 compliant.
```
Class GPBMetadata\Google\ApiCore\Tests\Unit\Example ./vendor/google/gax/metadata/Google/ApiCore/Tests/Unit/Example.php
```

ericsala:
Any solution on this issue? Thank you!

yuuan:
in dev:

```json
    "autoload-dev": {
        "psr-4": {
            "Google\\ApiCore\\Dev\\": "dev/src",
            "Google\\ApiCore\\": "tests",
            "GPBMetadata\\": "metadata"
        }
    },
```
`vendor/google/gax/metadata/Google/ApiCore/Tests/Unit/Example.php` is `GPBMetadata\Google\ApiCore\Tests\Unit\Example`.

in no-dev:

```json
    "autoload": {
        "psr-4": {
            "Google\\ApiCore\\": "src",
            "GPBMetadata\\Google\\": "metadata"
        }
    },
```
`vendor/google/gax/metadata/Google/ApiCore/Tests/Unit/Example.php` is `GPBMetadata\Google\Google\ApiCore\Tests\Unit\Example`.

bshaffer:
fixed in #278

PR #276, opened 2020-09-15
Fix autoload rule

yuuan:
Fixes: https://github.com/googleapis/gax-php/issues/264

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,14 +26,13 @@
     "autoload": {
         "psr-4": {
             "Google\\ApiCore\\": "src",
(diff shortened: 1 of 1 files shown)


yuuan:
@googlebot I signed it!

bshaffer:
Tests are failing due to an upstream change that we have a fix for in #273

dwsupplee:
Thanks for taking the time to open this @yuuan! It looks like https://github.com/googleapis/gax-php/pull/278 is the appropriate approach here, so I'm going to close this in favor of it.

PR #278, opened 2020-09-16
fix: autoloading of metadata classes

bshaffer:
addresses #264
Alternative to #276 

I believe this is the proper proper way to fix this, as it ensures the `Mocks` proto get loaded with the "production" code in src, and the "Example" gets loaded only with the tests.

This may actually not satisfy the Composer 2.0 warning, however. I've tested with Composer v2 and the latest version of Composer v1 (v1.10) and cannot see the warning mentioned in the bug.


dwsupplee:
This looks correct to me, the way to check if it resolves the error is to add the `--optimize-autoloader` flag when running `composer install`.

dwsupplee (approved):

PR #291, opened 2020-10-15
feat: adds support for GAPICs to validate REST-only transports

bshaffer:
- Adds `supportedTransports` private static function to `GapicClientTrait` and uses this to validate the transport option in `createTransport`


tests/Tests/Unit/GapicClientTraitTest.php
```diff
@@ -952,3 +973,45 @@ public function set($name, $val, $static = false)
...
+    private static function supportedTransports()
+    {
+        return ['rest', 'fake-transport'];
+    }
+
+    private static function defaultTransport()
+    {
+        return 'rest';
+    }
+}
```
dwsupplee, line 1017: Nit: Need a new line

dwsupplee (approved):

PR #294, opened 2020-10-26
build: Add .circleci/config.yml

miraleung:


dwsupplee:
@miraleung, why circle instead of kokoro?

bshaffer:
We will move to Github Actions instead

see https://github.com/googleapis/gax-php/issues/298

PR #300, opened 2020-11-04
fix: throws exceptions for unexpected header values

bshaffer:
another way to solve #299


dwsupplee (approved):

dwsupplee (approved):
LGTM, could we add just a few simple tests?

PR #301, opened 2020-11-05
chore: move to github actions

bshaffer:
Moves all travis tests into github actions!

Docs generation has been tested successfully on [my fork](https://github.com/bshaffer/gax-php/tree/gh-pages)


.github/workflows/tests.yml
```diff
@@ -0,0 +1,62 @@
...
+                proto_c_ext: [ true, false ]
+
+        name: PHP ${{ matrix.php }} Unit Test
+        steps:
+            - uses: actions/checkout@v2
+            - uses: codecov/codecov-action@v1
+            - name: Setup PHP
+              uses: shivammathur/setup-php@v2
+              with:
+                php-version: ${{ matrix.php }}
```
jdpedrie, line 23: Rather than a separate step for installing grpc, add here:

```
tools: pecl
extensions: grpc
```

See here for ref: https://github.com/googleapis/google-cloud-php/blob/master/.github/workflows/pubsub-emulator-system-tests.yaml#L24

jdpedrie (changes requested):

.github/workflows/tests.yml
bshaffer: working on it!!

dev/sh/build-docs.sh
```diff
@@ -3,7 +3,7 @@
 # Script to build doc site.
 # This script expects to be invoked from the gax-php root.
 #
-# This script will look for the TRAVIS_TAG environment variable, and
+# This script will look for the GITHUB_REF environment variable, and
```
jdpedrie, line 6: nit: may be better to move this to an argument and let the caller specify the ref.

i.e. 
```
<html><head><script>window.location.replace('/gax-php/$1/' + location.hash.substring(1))</script></head><body></body></html>
```

not sure if it needs to be `${1}` in the string...

And in the [action](https://github.com/googleapis/gax-php/pull/301/files#diff-9cf2000c53760d837a449f874e53f792819108d3a4bf346336d0f7d082deae2cR26), `            entrypoint: ./dev/sh/build-docs.sh $GITHUB_REF`

jdpedrie (approved):

dwsupplee (approved):

dev/sh/build-docs.sh
bshaffer: done!

PR #307, opened 2020-12-16
build: repo sync for github actions

bshaffer:


JustinBeckwith (approved):
LGTM.  This would take effect starting tonight when the cron runs, or you can manually trigger the job in Cloud Scheduler.  You may have to manually adjust these if timing is an issue.

JustinBeckwith:
Oh one other thing - I wouldn't classify this as a `fix: `.  Usually we reserve that for changes that fix bugs and have a material impact on the library itself.  release-please would use `fix: ` as a signal to propose a new release.  For this one, I'd use `build: `.

dwsupplee (approved):

PR #313, opened 2021-03-24
chore: remove unnecessary api endpoint check

bshaffer:


jdpedrie (approved):

yash30201:
Hey @bshaffer , is this PR still valid?

yash30201 (approved):
LGTM

bshaffer:
I think the problem here is that we didn't add support to use the Default Host _instead of_ the API Endpoint, which needs to happen before we can remove the endpoint check.

PR #318, opened 2021-04-16
Allow CredentialsWrapper to use the auth cache when fetching ADC.

MrMage:
Motivation: This lets us cache the result of `ApplicationDefaultCredentials::onGce()`, saving us a ping to the metadata server.

Let me know if I'm missing something, or if there was a good reason to provide `null`.


MrMage:
Made obsolete by https://github.com/googleapis/google-cloud-php/pull/3959.

PR #319, opened 2021-04-20
feat: add logical timeout helper to RetrySettings

noahdietz:
This PR adds a helper to the `RetrySettings` for configuring a logical call timeout. This helper, `logicalTimeout`, takes a single integer and applies it to each of the four existing timeout settings and sets the timeout multiplier to 1, returning an associative array for use with `with()` or via the `retrySettings` call option. The result is that each RPC attempt will take no longer than the time remaining in the overall deadline calculated at the start of the first call using the given timeout. The time remaining in this overall deadline is the timeout for each RPC attempt.

This helper is meant to simplify how users can think about the timeout settings, and reduce the need for `DEADLINE_EXCEEDED` to be retryable, which is an anti-pattern, but necessary for the "timeout backoff" strategy available via the existing retry/timeout logic.

### DIFF
--- a/src/RetrySettings.php
+++ b/src/RetrySettings.php
@@ -383,9 +383,7 @@ class RetrySettings
      * Creates an associative array of the {@see Google\ApiCore\RetrySettings} timeout fields configured
      * with the given timeout specified in the $timeout parameter interpreted as a logical timeout.
      *
-     * @param int $timeout {
-     *     The timeout in milliseconds to be used as a logical call timeout.
-     * }
+     * @param int $timeout The timeout in milliseconds to be used as a logical call timeout.
      * @return array
      */
     public static function logicalTimeout($timeout)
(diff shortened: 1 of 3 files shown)


miraleung (approved):
LGTM, please wait for Brent's or David's review.

src/RetrySettings.php
```diff
@@ -349,6 +379,26 @@ public function with(array $settings)
         return new RetrySettings($settings + $existingSettings);
     }
 
+    /**
+     * Creates an associative array of the {@see Google\ApiCore\RetrySettings} timeout fields configured
+     * with the given timeout specified in the $timeout parameter interpreted as a logical timeout.
+     *
+     * @param int $timeout {
```
dwsupplee, line 386: Just a minor nit, the bracket style is used for notating the elements in an array, this can just be `@param int $timeout desc`

dwsupplee (changes requested):

src/RetrySettings.php
noahdietz: Thanks! Done.

dwsupplee (approved):

PR #322, opened 2021-05-03
[gax-php] feat: add REST headers

miraleung:
This is needed primarily for REGAPIC. However, sending the REST header can also benefit gRPC and REST GAPIC libraries (the default) as well, since we don't support gRPC-only libraries.

Using gax-php's version as the default REST version.

### DIFF
--- a/src/AgentHeader.php
+++ b/src/AgentHeader.php
@@ -50,7 +50,7 @@ class AgentHeader
      *     @type string $gapicVersion the code generator version of the GAPIC library.
      *     @type string $apiCoreVersion the ApiCore version.
      *     @type string $grpcVersion the gRPC version.
-     *     @type string $restVErsion the REST transport version (typically same as the
+     *     @type string $restVersion the REST transport version (typically same as the
      *           ApiCore version).
      * }
      * @return array Agent header array
--- a/src/AgentHeader.php
+++ b/src/AgentHeader.php
@@ -50,6 +50,8 @@ class AgentHeader
      *     @type string $gapicVersion the code generator version of the GAPIC library.
      *     @type string $apiCoreVersion the ApiCore version.
      *     @type string $grpcVersion the gRPC version.
+     *     @type string $restVErsion the REST transport version (typically same as the
+     *           ApiCore version).
      * }
      * @return array Agent header array
      */
--- a/src/AgentHeader.php
+++ b/src/AgentHeader.php
@@ -64,6 +64,7 @@ class AgentHeader
         //      - gapicVersion (gapic/)
         //      - apiCoreVersion (gax/)
         //      - grpcVersion (grpc/)
+        //      - restVersion (rest/)
 
         $phpVersion = isset($headerInfo['phpVersion'])
             ? $headerInfo['phpVersion']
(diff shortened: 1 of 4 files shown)


src/AgentHeader.php
```diff
@@ -64,6 +64,7 @@ public static function buildAgentHeader($headerInfo)
         //      - gapicVersion (gapic/)
         //      - apiCoreVersion (gax/)
         //      - grpcVersion (grpc/)
+        //      - restVersion (rest/)
```
bshaffer, line 67: Don't forget to add this to the `$headerInfo` phpdoc for the method!

src/GapicClientTrait.php
```diff
@@ -318,6 +318,20 @@ private function setClientOptions(array $options)
             $clientConfig,
             $options['disableRetries']
         );
+
+        // Edge case: If the client has the gRPC extension installed, but is
+        // a REST-only library, then the grpcVersion header should not be set.
+        if ($this->transport instanceof GrpcFallbackTransport) {
+            $options['grpcVersion'] = phpversion('grpc');
+            $options['restVersion'] = Version::getApiCoreVersion();
```
bshaffer, line 326: As `GrpcFallbackTransport` doesn't actually use the `grpc` extension (or grpc), I'm not sure we want to set the `grpc/` header, as it's most likely going to be empty.

src/AgentHeader.php
miraleung: Done.

src/GapicClientTrait.php
```diff
@@ -318,6 +318,20 @@ private function setClientOptions(array $options)
...
+        } else if ($this->transport instanceof GrpcTransport) {
+            $options['grpcVersion'] = phpversion('grpc');
+            unset($options['restVersion']);
+        } else if ($this->transport instanceof RestTransport) {
+            unset($options['grpcVersion']);
+            $options['restVersion'] = Version::getApiCoreVersion();
```
dwsupplee, line 332: It seems a little odd to send the same value as we do for the gax header, do other languages send the same value twice too?

miraleung: The doc says this should be the "the language-specific REST implementation." If there's a better version we should use that's not gax-php, please let me know!

dwsupplee: This seems to make the most sense then :).

src/AgentHeader.php
```diff
@@ -50,6 +50,8 @@ class AgentHeader
      *     @type string $gapicVersion the code generator version of the GAPIC library.
      *     @type string $apiCoreVersion the ApiCore version.
      *     @type string $grpcVersion the gRPC version.
+     *     @type string $restVErsion the REST transport version (typically same as the
```
dwsupplee, line 53:
```suggestion
     *     @type string $restVersion the REST transport version (typically same as the
```

src/AgentHeader.php
```diff
@@ -87,11 +90,23 @@ public static function buildAgentHeader($headerInfo)
             : Version::getApiCoreVersion();
         $metricsHeaders['gax'] = $apiCoreVersion;
 
+        // Context on library type identification (between gRPC+REST and REST-only):
+        // This uses the gRPC extension's version if 'grpcVersion' is not set, so we
+        // cannot use the presence of 'grpcVersion' to determine whether or not a library
+        // is gRPC+REST or REST-only. However, we cannot use the extension's presence
+        // either, since some clients may have the extension installed but opt to use a
+        // REST-only library (e.g. GCE).
+        // TODO: Should we stop sending empty gRPC headers?
```
dwsupplee, line 99: Can we file an associated issue linking to this PR to help us follow up?

dwsupplee (changes requested):

dwsupplee (approved):

src/AgentHeader.php
miraleung: Done.

src/GapicClientTrait.php
miraleung: Done.

PR #325, opened 2021-06-01
feat: add mTLS support via clientCertSource client option

bshaffer:
- Adds client option `clientCertSource` 
 - Adds support for client certs to all three transports
 - Verified calls to mTLS endpoint works locally

**Usage**

```php
use Google\Cloud\PubSub\V1\PublisherClient;

$projectId = 'YOUR_PROJECT_ID';
$pubsub = new PublisherClient([
    'apiEndpoint' => 'pubsub.mtls.googleapis.com',
    'clientCertSource' => function () { 
         return file_get_contents(__DIR__ . '/cert.pem');
    },
]);

$topics = $pubsub->listTopics('projects/' . $projectId);
```

**TODO**
 - [x] Add support for checking the existence and loading the default client cert (from `~/.secureConnect/context_aware_metadata.json`) (done in googleapis/google-auth-library-php#353)
 - [x] Add support for `GOOGLE_API_USE_MTLS_ENDPOINT` and `GOOGLE_API_USE_CLIENT_CERTIFICATE`
 - [x] Add support to automatically use mTLS endpoint for `apiEndpoint`
 - [x] Tests

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -7,7 +7,7 @@
     "license": "BSD-3-Clause",
     "require": {
         "php": ">=5.5",
-        "google/auth": "^1.2.0",
+        "google/auth": "dev-mtls as 1.13.0",
         "google/grpc-gcp": "^0.1.0",
         "grpc/grpc": "^1.13",
         "google/protobuf": "^3.12.2",
--- a/composer.json
+++ b/composer.json
@@ -7,7 +7,7 @@
     "license": "BSD-3-Clause",
     "require": {
         "php": ">=5.5",
-        "google/auth": "dev-mtls as 1.17.0",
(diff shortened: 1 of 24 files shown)


composer.json
```diff
@@ -7,7 +7,7 @@
     "license": "BSD-3-Clause",
     "require": {
         "php": ">=5.5",
-        "google/auth": "^1.2.0",
+        "google/auth": "dev-mtls as 1.13.0",
```
bshaffer, line 10: Before we merge, this needs to be updated to require the latest tagged version of `google/auth`

dwsupplee (approved):

PR #333, opened 2021-08-18
fix: ensure required query parameters with default values get sent over the wire

dwsupplee:
Follows https://github.com/googleapis/gapic-generator-php/commit/2d5ef5853544d58de2b6fdd5cb8335c38174af23

### DIFF
--- a/tests/Tests/Unit/RequestBuilderTest.php
+++ b/tests/Tests/Unit/RequestBuilderTest.php
@@ -337,6 +337,19 @@ class RequestBuilderTest extends TestCase
         $this->assertEquals('some-value', $query['stringValue']);
     }
 
+    public function testMethodWithRequiredQueryParametersAndDefaultValues()
+    {
+        $message = (new MockRequestBody())
+            ->setName('')
+            ->setNumber(0);
+
+        $request = $this->builder->build(self::SERVICE_NAME . '/MethodWithRequiredQueryParameters', $message);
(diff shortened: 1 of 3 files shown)


tests/Tests/Unit/RequestBuilderTest.php
```diff
@@ -337,6 +337,19 @@ public function testMethodWithoutPlaceholders()
...
+    {
+        $message = (new MockRequestBody())
+            ->setName('')
+            ->setNumber(0);
+
+        $request = $this->builder->build(self::SERVICE_NAME . '/MethodWithRequiredQueryParameters', $message);
+        $query = Psr7\parse_query($request->getUri()->getQuery());
+
+        $this->assertEquals('', $query['name']);
+        $this->assertEquals(0, $query['number']);
```
bshaffer, line 350: Just a question - what happens if no calls are made to `setName` and `setNumber`? Will the query param be empty or will the query parameter be omitted? And if that is the case, how will the API function?

bshaffer (approved):

tests/Tests/Unit/RequestBuilderTest.php
dwsupplee: > what happens if no calls are made to setName and setNumber? Will the query param be empty or will the query parameter be omitted?

If no calls are made to the message class, the test results would be the same. If a query parameter is determined to be required (by the generator implementation) it will always be added to the query string as long as an implicit value or default value are set.

PR #337, opened 2021-08-26
feat: add Compute LRO support

bshaffer:
Opens up `OperationResponse` to accept generic "operations clients" through descriptors and retains Backwards Compatibility, while handling operations from both OnePlatform APIs and Diregapic (e.g. Compute) APIs. 

See https://github.com/googleapis/google-cloud-php/pull/4438 for a Proof of Concept of the changes needed in the Compute GAPIC client

#### OperationsResponse

Adds support in `OperationResponse` for the following options (which are provided via descriptors):

 - `additionalArgs` (optional) - for additional arguments to pass to the Operations Client methods on each method call.
 - `getOperationMethod` - The Operations Client method used to fetch the Operation object.
 - `cancelOperationMethod` - The Operations Client method used to cancel the operation.
 - `deleteOperationMethod` - The Operations Client method used to delete the operation.
 - `operationStatusMethod` - The Operation object method used to get the status of the operation.
 - `operationStatusDoneValue` - The value used to determine if the operation is complete.
 - `operationErrorCodeMethod` - The method used to determine if the operation has errors.

#### OperationsMiddleware

 -  The `$operationsClient` is no longer typehinted to `Google\ApiCore\LongRunning\OperationsClient`, but instead can be any object type.
 - The object returned from the middleware call is no longer typehinted to `Google\LongRunning\Operation`, but instead is typehinted to the more generic `Google\Protobuf\Internal\Message`.

#### GapicClientTrait

 - A client option `operationsClientClass` has been added, which is used in `GapicClientTrait::createOperationsClient` to create an Operations Client of a class different than the default `Google\ApiCore\LongRunning\OperationsClient`.
 - `GapicClientTrait::startOperationsCall` no longer requires the `$client` argument to be of type `Google\ApiCore\LongRunning\OperationsClient`, but instead accepts any type. It also accepts a new `$operationClass` argument for specifying the return type of the operation call.
 - `GapicClientTrait::startOperationsCall` respects the new `additionalArgumentMethods` LRO descriptor, and uses it to call methods on the Request object to populate the `additionalOperationArguments` option passed to OperationResponse.

**NOTE**: The value `-1` for unknown codes has been updated for certain HTTP code values, as more known codes have been added. This changes behavior by giving the user more information, and could potentially break anyone relying on the specific value of "Unknown code". It's a gray area, but we made the call that overall, fewer unknown codes will improve the user experience.

### DIFF
--- a/src/OperationResponse.php
+++ b/src/OperationResponse.php
@@ -319,7 +319,7 @@ class OperationResponse
                 ? call_user_func([$this->lastProtoResponse, $this->operationErrorMessageMethod])
                 : null;
             return (new Status())
-                ->setCode($errorCode)
+                ->setCode(ApiStatus::rpcCodeFromHttpStatusCode($errorCode))
                 ->setMessage($errorMessage);
         }
 
--- a/src/ApiStatus.php
+++ b/src/ApiStatus.php
@@ -95,9 +95,12 @@ class ApiStatus
         Code::UNAUTHENTICATED => ApiStatus::UNAUTHENTICATED,
     ];
     private static $httpStatusCodeToRpcCodeMap = [
+        400 => Code::INVALID_ARGUMENT,
         401 => Code::UNAUTHENTICATED,
         403 => Code::PERMISSION_DENIED,
         404 => Code::NOT_FOUND,
+        409 => Code::ABORTED,
+        416 => Code::OUT_OF_RANGE,
         429 => Code::RESOURCE_EXHAUSTED,
         499 => Code::CANCELLED,
         501 => Code::UNIMPLEMENTED,
@@ -140,8 +143,7 @@ class ApiStatus
 
     /**
      * Maps HTTP status codes to Google\Rpc\Code codes.
-     * Only map codes which do not map to multiple gRPC codes (e.g. excludes
-     * 400, 409, and 500).
+     * Some codes are left out because they map to multiple gRPC codes (e.g. 500).
      *
      * @param int $httpStatusCode
      * @return int
@@ -151,6 +153,19 @@ class ApiStatus
         if (array_key_exists($httpStatusCode, self::$httpStatusCodeToRpcCodeMap)) {
             return self::$httpStatusCodeToRpcCodeMap[$httpStatusCode];
         }
+        // All 2xx
+        if ($httpStatusCode >= 200 && $httpStatusCode < 300) {
+            return Code::OK;
+        }
+        // All 4xx
+        if ($httpStatusCode >= 400 && $httpStatusCode < 500) {
+            return Code::FAILED_PRECONDITION;
+        }
+        // All 5xx
+        if ($httpStatusCode >= 500 && $httpStatusCode < 600) {
+            return Code::INTERNAL;
+        }
+        // Everything else (We cannot change this to Code::UNKNOWN because it would break BC)
         return ApiStatus::UNRECOGNIZED_CODE;
     }
 }
--- a/tests/Tests/Unit/ApiStatusTest.php
+++ b/tests/Tests/Unit/ApiStatusTest.php
@@ -116,4 +116,39 @@ class ApiStatusTest extends TestCase
             [-1, ApiStatus::UNRECOGNIZED_STATUS]
         ];
     }
+
+    /**
+     * @dataProvider getHttpCodeAndStatus
+     */
+    public function testRpcCodeFromHttpStatus($httpCode, $rpcCode)
+    {
+        $this->assertSame($rpcCode, ApiStatus::rpcCodeFromHttpStatusCode($httpCode));
+    }
+
+    public function getHttpCodeAndStatus()
+    {
+        return [
+            [400, Code::INVALID_ARGUMENT],
+            [401, Code::UNAUTHENTICATED],
+            [403, Code::PERMISSION_DENIED],
+            [404, Code::NOT_FOUND],
+            [409, Code::ABORTED],
+            [416, Code::OUT_OF_RANGE],
+            [429, Code::RESOURCE_EXHAUSTED],
+            [499, Code::CANCELLED],
+            [501, Code::UNIMPLEMENTED],
+            [503, Code::UNAVAILABLE],
+            [504, Code::DEADLINE_EXCEEDED],
+            // Unmapped 2xx returns Status::OK
+            [201, Code::OK],
+            // Unmapped 4xx returns Status::FAILED_PRECONDITION
+            [405, Code::FAILED_PRECONDITION],
+            // Unmapped 5xx returns Status::INTERNAL
+            [505, Code::INTERNAL],
+            // Anything else returns Status::UNRECOGNIZED_CODE
+            [-1, ApiStatus::UNRECOGNIZED_CODE],
+            [100, ApiStatus::UNRECOGNIZED_CODE],
+            [300, ApiStatus::UNRECOGNIZED_CODE],
+        ];
+    }
 }
--- a/tests/Tests/Unit/OperationResponseTest.php
+++ b/tests/Tests/Unit/OperationResponseTest.php
@@ -37,6 +37,7 @@ use Google\ApiCore\OperationResponse;
 use Google\LongRunning\Operation;
 use Google\Protobuf\Any;
 use PHPUnit\Framework\TestCase;
+use Google\Rpc\Code;
 
 class OperationResponseTest extends TestCase
 {
@@ -263,7 +264,7 @@ class OperationResponseTest extends TestCase
         $error = $operationResponse->getError();
 
         $this->assertNotNull($error);
-        $this->assertEquals(500, $error->getCode());
+        $this->assertEquals(Code::INTERNAL, $error->getCode());
         $this->assertEquals('It failed, sorry :(', $error->getMessage());
     }
 


src/OperationResponse.php
```diff
@@ -340,4 +377,13 @@ public function getMetadata()
         $metadata->mergeFromString($any->getValue());
         return $metadata;
     }
+
+    private function operationsCall($method, $name, array $additionalArgs)
+    {
+        if (is_null($method)) {
+            throw new LogicException('cancelling operations is not supported by this API');
```
bshaffer, line 384: This needs to throw an RPC error code for not implemented, to be consistent with OnePlatform

bshaffer: On second thought, do we really want to throw an `ApiException` when no API call was made? This is misleading. Rather, what we have here seems appropriate.

src/Middleware/OperationsMiddleware.php
```diff
@@ -66,11 +65,14 @@ public function __invoke(Call $call, array $options)
...
             $options
-        )->then(function (Operation $response) {
+        )->then(function (Message $response) {
             $options = $this->descriptor + [
                 'lastProtoResponse' => $response
             ];
-            return new OperationResponse($response->getName(), $this->client, $options);
+            $operationNameMethod = isset($options['operationNameMethod'])
+                ? $options['operationNameMethod'] : 'getName';
+            $operationName = call_user_func([$response, $operationNameMethod]);
```
noahdietz, line 74: Why use `call_user_func` here and a "dynamic" variable function earlier (`$request->$additionalArgsMethodName()`)? Is there a different between the two? My PHP n00b eyes like the variable function style a little better, but not sure if it makes a difference.

src/OperationResponse.php
```diff
@@ -127,9 +154,16 @@ public function __construct($operationName, $operationsClient, $options = [])
...
     public function isDone()
     {
-        return (is_null($this->lastProtoResponse) || is_null($this->lastProtoResponse->getDone()))
-            ? false
-            : $this->lastProtoResponse->getDone();
+        if (!$this->hasProtoResponse()) {
+            return false;
+        }
+
+        $status = call_user_func([$this->lastProtoResponse, $this->operationStatusMethod]);
```
noahdietz, line 161: Same thing here regarding `call_user_func` vs. variable function calls.

src/OperationResponse.php
```diff
@@ -241,9 +299,9 @@ public function getResult()
      * @return Status|null The status of the operation in case of failure, or null if
      *                                 operationFailed() is false.
      */
-    public function getError()
+    public function getError(Message $lastProtoResponse = null)
```
noahdietz, line 302: Where is this parameter used in this function? Doesn't `$this->lastProtoResponse` refer to the member field of the object, not the parameter?

src/OperationResponse.php
```diff
@@ -241,9 +299,9 @@ public function getResult()
...
     {
-        if (!$this->isDone() || is_null($this->lastProtoResponse->getError())) {
+        if (!$this->isDone()) {
             return null;
         }
         return $this->lastProtoResponse->getError();
```
noahdietz, line 307: Just want to double check: if `$lastProtoResponse` is in fact `null`, does this throw an exception? I guess this case isn't possible because `$this->isDone()` would return `false` if `$this->lastProtoResponse` was `null`.

src/OperationResponse.php
noahdietz, line 302: Nit: I don't see use of the `operation_field = ERROR_MESSAGE` or `operation_field = ERROR_CODE` configurations. Do we need to use these? FWIW I think the `Operation` in `compute` has an `Error error` field so this might not be necessary.

noahdietz (commented):
Thought I'd take a look so I could better understand how it all works :)

src/GapicClientTrait.php
```diff
@@ -662,18 +669,32 @@ private function startOperationsCall(
...
+
+        // Call the methods supplied in "additionalArgumentMethods" on the request Message object
+        // to build the "additionalOperationArguments" option for the operation response.
+        if (isset($descriptor['additionalArgumentMethods'])) {
+            $additionalArgs = [];
+            foreach ($descriptor['additionalArgumentMethods'] as $additionalArgsMethodName) {
+                $additionalArgs[] = $request->$additionalArgsMethodName();
+            }
+            $descriptor['additionalOperationArguments'] = $additionalArgs;
+            unset($descriptor['additionalArgumentMethods']);
```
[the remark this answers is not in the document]
bshaffer: It's not strictly necessary, but why pass options that aren't supported down to the `OperationResponse` object?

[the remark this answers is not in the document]
noahdietz: Fair point! Thanks!

src/Middleware/OperationsMiddleware.php
bshaffer: They are identical!

The only reason I use `call_user_func` and `call_user_func_array` in `OperationResponse` is because they're using `$this->variableName`, which can't be called with the syntax mentioned here. I wanted to be consistent so I used it again here, but forgot about `$request->$additionalArgsMethodName()`. 

I'm ambivalent, I do not think there's any difference other than readability.

```suggestion
            $operationName = $response->$operationNameMethod();
```

src/OperationResponse.php
bshaffer: As mentioned above, here it does matter, because you can't call a property as a method without saving it as a local variable.

src/OperationResponse.php
bshaffer: Good catch! This was refactored and so the parameter here should be removed.

src/Middleware/OperationsMiddleware.php
noahdietz: Oh ok the `$this_>variableName` syntax being a "limiting factor" is new to me. TIL. Thanks, feel free to leave it all the way it is, just practicing my PHP code review sk1lLz 😋

src/OperationResponse.php
bshaffer: That's correct. But for extra readability and safety I'll add a call to `$this->hasProtoResponse()` here also.

noahdietz: SGTM. Thanks.

src/OperationResponse.php
bshaffer: Currently ee just return the raw error object to our customers to inspect, and don't currently provide any convenience functions around it.

We could do something like add a `getErrorMessage` and `getErrorCode` method on the `OperationResponse` object. That way users could handle operations from Compute and OnePlatform the same. But this could be added at any time.

@dwsupplee what do you think?

noahdietz: > Currently we just return the raw error object to our customers to inspect, and don't currently provide any convenience functions around it.

Right. But does it make sense for the `getError` method to construct an error using the `Operation` fields `error_code` and `error_message` if the `Operation` proto doesn't have an `error` field? Feel like these annotations must have _some_ use since they are there. As you said, we could add something later for that.

bshaffer: The other operation responses returns a `Google\Rpc\Status` object (see [here](https://github.com/googleapis/common-protos-php/blob/master/src/Rpc/Status.php)) on calls to `$lastProtoResponse->getError()`. We could construct one of these, which feels _slightly_ hacky to me, since that's not actually what the API returns. But it could be done.

Is that what the other languages are doing?

noahdietz: > which feels _slightly_ hacky to me, since that's not actually what the API returns

I agree, that's kind of hacky...IDK what other languages are doing, but maybe we should consult with Vadym about the purpose of the annotation.

bshaffer: @vam-google

src/OperationResponse.php
```diff
@@ -408,9 +414,6 @@ public function getMetadata()
 
     private function operationsCall($method, $name, array $additionalArgs)
     {
-        if (is_null($method)) {
```
noahdietz: Why remove this? There is a "new" seg fault in the unit tests, could this be why?

noahdietz (approved):
LGTM aside from one question. @dwsupplee should still review.

src/OperationResponse.php
vam-google: **tl;dr;** I guess there are two parts here:
1) Do we need those annotations to do polling right?
2) Do we need those annotations to expose error details to the users in a well-defined form (i.e. do some wrapping around raw objects to  make standard and non-standard error details look the same on the surface).

The first question is critical because it is about correctness (i.e. required). The second one is less important, because it is about usability (optional, and you may choose to not do it, if you think is better not do it).

More details below:

Whatever PHP does here the following functionality should be present at minimum in the generated gapic library:
1) Users should be able to get error details in case if error occurs during polling.

2) If polling is done automatically, the polling logic should be able to take reasonable actions in case of an error (usually it just means stop polling and report an error in a langauge-idiomatic way). To take the reasonable action, the logic needs to know where the error information is supposed to come from - this is why the annotations exist in the first place.

3) If polling is not done automatically, then please clarify what exactly LRO support does and what it doesn't do for the user in PHP. To the best of my understanding polling is started by user explicitly in PHP, but then it is done automatically, and to properly stop in case of an error, we need that info from the operation. Note, the polling request may potentially return 200 OK as HTTP error code, but the returned object itself may contain error details if polling didn't go well (i.e. http error codes and LRO polling are on different layers - success on rpc layer does not mean success on LRO logic layer, which is built on top of that rpc). 

4) If in PHP the standard Operation object is not wrapped in some return object-independent abstraction, and the same approach was chosen for DIREGAPIC lro (i.e. the custom Operation object is still exposed raw to the user), then, I guess there is no other option for the error part of the object - it must be exposed raw too. I.e. if the whole object exposed as is the error details fields of it have to be exposed as is as well. So in that case, yes, you do not need extra annotations to reason about error information in the object, because the object as a whole is not processed/wrapped in a special way. I would guess that the error details are still needed to properly stop pppolling in case of an error. If that is not the case, please clarify why it is not true for PHP.

As for constructing fake `Google\Rpc\Status` object I would advice against it. Basically the idea is to minimize numbers of "fakes"  and not try to make API look different from what it actually is, if possible. Sometimes it may be unavoidable, if "fake" intermediate objects are needed to make the rest of LRO machinery work, but for cases when it can be avoided, please do not construct unnecessary facade objects.


On the other hand, form Brent's response:

> We could do something like add a getErrorMessage and getErrorCode method on the OperationResponse object. That way users could handle operations from Compute and OnePlatform the same. But this could be added at any time.

I think it has value in it, but as long as the unified interface is generic and does not make one API pretend that it is the other one (i.e. make compute LRO pretend like they return Rpc/Status  while it is not). If both standard and non/standard error responses are wrapped in a generic abstraction (and neither one tries to pretend that it is the other one), this is what is exposed to the user, and that can be done in a backward compatible way - i believe that would be the best option.

src/OperationResponse.php
bshaffer: I didn't remove it, I just moved it into the two methods where we think it could happen.

src/OperationResponse.php
```diff
@@ -274,7 +329,7 @@ public function getLastProtoResponse()
     }
 
     /**
-     * @return OperationsClient The OperationsClient object used to make
+     * @return The OperationsClient object used to make
```
dwsupplee, line 332:
```suggestion
     * @return object The OperationsClient object used to make
```

src/OperationResponse.php
```diff
@@ -297,7 +352,10 @@ public function getOperationsClient()
      */
     public function cancel()
     {
-        $this->operationsClient->cancelOperation($this->getName());
```
dwsupplee, line 300: The docblock here mentions a status of `\Google\Rpc\Code::CANCELLED` being set to the response of `getError` - will that hold true with DIREGAPIC? This question applies generally to all the RPCs and error codes.

src/OperationResponse.php
```diff
@@ -243,7 +298,7 @@ public function getResult()
      */
     public function getError()
```
dwsupplee, line 299: Was there any further discussion on standardizing the error response? After some thought, I lean towards it being the right thing to do - assuming it isn't a big hurdle to jump.

src/OperationResponse.php
```diff
@@ -243,7 +298,7 @@ public function getResult()
...
     {
-        if (!$this->isDone() || is_null($this->lastProtoResponse->getError())) {
+        if (!$this->hasProtoResponse() || !$this->isDone()) {
             return null;
         }
         return $this->lastProtoResponse->getError();
```
dwsupplee, line 304: Should this be using the configurable `operationErrorCodeMethod `?

tests/Tests/Unit/Middleware/OperationsMiddlewareTest.php
```diff
@@ -0,0 +1,69 @@
+<?php
+/*
+ * Copyright 2018 Google LLC
```
dwsupplee, line 3:
```suggestion
 * Copyright 2021 Google LLC
```

dwsupplee (changes requested):
Since there isn't a diff on it, I can't comment directly on this line - but `getLastProtoResponse` in `OperationResponse` will need the return type updated to something like `Operation|mixed`.

src/OperationResponse.php
bshaffer: The Compute team said that checking the existing of `error` is sufficient, which is what we are doing for OnePlatform. We currently do not have a use case for using the Error Code and Error Message annotations, so I am leaning towards removing them.

> I lean towards it being the right thing to do

Can you explain what you mean by this?

src/OperationResponse.php
bshaffer: Again, I think it would be best to remove that field, but if we keep it, how do you see it being used here?

src/OperationResponse.php
dwsupplee: We discussed offline but for anyone else reading, compute returns a different message type for an error response. The discussion is around whether we should standardize compute and other possible APIs error responses into the existing status message when feasible.

src/OperationResponse.php
bshaffer: This is updated... Now it returns a `Google\Rpc\Status` object populated with the result of `operationErrorCodeMethod` and `operationErrorMessageMethod`. 

Additionally, the `hasErrors` private method (called in `operationSuccessful` and `operationFailed`) has the following behavior: 
1. checks for the existence of `getError`, if it exists, returns false if it's empty, true otherwise.
2. checks for the existence of `operationErrorCodeMethod`, if it exists, returns false if that returns empty, true otherwise
3. throws an exception saying it cannot determine the error code status (e.g. it's misconfigured)

This means the error annotations are mostly only used to return a consistent object from `OperationResponse::getError`, but _can_ be used to determine the success of the operation _if_ the `error` field doesn't exist.

src/OperationResponse.php
bshaffer: See the comment below - this now returns `Google\Rpc\Status` based on the descriptors for error code and message.

src/OperationResponse.php
bshaffer: I updated the doc block to be more clear! PTAL!

dwsupplee: The docblock LGTM, thanks! 👍 I just wanted to double check, do the error codes of `\Google\Rpc\Code` map properly to what we get back from a DIREGAPIC? An example would be, if someone has written a snippet to check for a specific error code with a oneplatform API, will that work correctly with DIREGAPIC?

bshaffer: No, they do not map to the same codes.

src/OperationResponse.php
bshaffer: The users would still be able to access it by calling `$operationResponse->getLastProtoResponse()->getError()`

PR #342, opened 2021-10-06
feat: add JsonStreamDecoder

noahdietz:
Our REST server stream API implementations conform to the following protocol:
- Use `Content-Type: application/json`
- The response body takes the form of a JSON Array, starting with `[` as the first body byte.
- Each message sent by the server is a protobuf-JSON encoded object 
- Each message sent by the server is separated by a comma (valid JSON Array syntax)
- The end of the stream is denoted by a closing `]` for the root JSON Array

This `JsonStreamDecoder` could technically support any type of stream that is valid JSON, but it is not a strict requirement. Follow up work can expand support here.

Note: This is marked as `@experimental` to start.


src/Transport/Rest/JsonStreamDecoder.php
```diff
@@ -0,0 +1,106 @@
...
+        $this->buffer = new BufferStream($bufSize);
+    }
+
+    public function decode()
+    {
+        $message = $this->decodeType;
+        $open = 0;
+        $str = false;
+        while (!$this->stream->eof()) {
+            $b = $this->stream->read(1);
```
alexander-fenster, line 64: Knowing nothing about PHP and especially nothing about its streaming, I wonder if there is a performance difference between reading 1 byte from the stream versus reading everything that is available at this moment and then iterating over the bytes that were read.  In case if `$this->stream->read()` invocation is mapped to a system call, which is possible, the difference might be substantial.

src/Transport/Rest/JsonStreamDecoder.php
```diff
@@ -0,0 +1,106 @@
...
+            // Open/close double quotes of a key or value.
+            if ($b === '"') {
+                $str = !$str;
+            }
+            // Blank space between messages.
+            if ($b === '' || (ctype_space($b) && !$str)) {
```
alexander-fenster, line 70: Not sure how `$b` can be an empty string here.

alexander-fenster: ...also, from what I understand, it will ignore the blank space regardless of the value of `$open`, effectively making the JSON minified?

noahdietz: `read` will only return what is available, it doesn't block, so if there is nothing on the stream then read returns empty string.

noahdietz: > effectively making the JSON minified

Yep

src/Transport/Rest/JsonStreamDecoder.php
```diff
@@ -0,0 +1,106 @@
...
+                    break;
+                }
+            }
+            $this->buffer->write($b);
+            // A message-closing byte was just buffered. Decode the
+            // message with the decode type, clearing the buffer,
+            // and yield it.
+            if ($open === 1) {
+                $return = new $message();
+                $return->mergeFromJsonString((string)$this->buffer, $this->ignoreUnknown);
```
alexander-fenster, line 98: Is the buffer cleaned in `mergeFromJsonString`?

tests/Tests/Unit/Transport/Rest/JsonStreamDecoderTest.php
```diff
@@ -0,0 +1,58 @@
...
+    public function testJsonStreamDecoder() {
+        $stream = Psr7\Utils::streamFor('[
+            {"name": "foo"},
+            {"name": "bar"}
+        ]');
+        $decoder = new JsonStreamDecoder($stream, Operation::class, ['bufferSizeBytes' => 1024]);
+        foreach($decoder->decode() as $op) {
+            $name = $op->getName();
+            $this->assertNotNull($name);
+            $this->assertNotEmpty($name);
```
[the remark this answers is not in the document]
noahdietz: Done.

src/Transport/Rest/JsonStreamDecoder.php
alexander-fenster: Oh no, I went to read PHP documentation 😨 

Maybe [this](https://www.php.net/manual/en/function.stream-set-blocking.php) is related? Then you could probably do a `select`-style polling.

alexander-fenster (approved):
My approval isn't worth much here, but anyway :)

src/Transport/Rest/JsonStreamDecoder.php
noahdietz: That's a great lead, Alex, thank you. I will need to determine if Guzzle's Stream is blocking or not by default before refactoring this to potentially _make_ it blocking.

src/Transport/Rest/JsonStreamDecoder.php
```diff
@@ -0,0 +1,177 @@
...
+        $level = 0;
+        $str = false;
+        $prev = '';
+        $chunk = '';
+        $cursor = 0;
+        $start = 0;
+        $end = 0;
+        while ($chunk !== '' || !$this->stream->eof()) {
+            // Read up to $readChunkSize bytes from the stream.
+            $chunk .= $this->stream->read($this->readChunkSize);
```
summer-ji-eng, line 98: Is this `readChunkSize` fixed number? 
The data is sent in series of chunks, the length could be different. For [example](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Transfer-Encoding#chunked_encoding),
```
HTTP/1.1 200 OK
Content-Type: text/plain
Transfer-Encoding: chunked

7\r\n
Mozilla\r\n
9\r\n
Developer\r\n
7\r\n
Network\r\n
0\r\n
\r\n
```
We should support random chunk size. But maybe I misreading here, correct me if I am wrong. :)

tests/Tests/Unit/Transport/Rest/JsonStreamDecoderTest.php
```diff
@@ -0,0 +1,94 @@
...
+
+
+class JsonStreamDecoderTest extends TestCase
+{
+    use TestTrait;
+
+    /**
+     * @dataProvider buildResponseStreams
+     */
+    public function testJsonStreamDecoder(array $responses, $decodeType, $stream, $readChunkSizeBytes) {
```
summer-ji-eng, line 51: Bear with me, I am not familiar with PHP.
Where does `testJsonStreamDecoder` invoke?

src/Transport/Rest/JsonStreamDecoder.php
noahdietz: It's a fixed number that is configurable by users and defaults to 1024 bytes. The code here will read messages that are split over the "chunk" boundary i.e. half of a message is in the first 1024 bytes, the next half of the message is in the next 1024 bytes. It could even be read 1 byte at a time. The chunk read size is really just a concern for performance.

tests/Tests/Unit/Transport/Rest/JsonStreamDecoderTest.php
noahdietz: I don't follow the question. This is a unit test and it instantiates a JsonStreamDecoder then calls `decode` which generates response messages as they are decoded.

bshaffer (approved):

PR #349, opened 2021-11-11
chore: Lint the code

lucasmichot:
When a package is supported by a very wide range of PHP versions, it is very easy to introduce some syntax errors.

Those errors should ideally be found before the testing phase, eg. in the linting phase

This PR adds 2 packages to lint the code and update the main testing workflows to introduce enforce this linting.
One linting analysis will be run for each PHP version

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -35,6 +35,8 @@ jobs:
                 timeout_minutes: 10
                 max_attempts: 3
                 command: composer install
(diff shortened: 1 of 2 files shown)


bshaffer:
We already have a linter running as part of the GitHub action. Between that, the tests, and code coverage, I don't think this would add anything.

Thank you for your contribution. If you think I'm wrong about that, please open another PR.

PR #351, opened 2021-11-15
feat: support for API Key client option

bshaffer:
Adds support for API Keys in GAPIC clients via a new `apiKey` client option:

```php
$productSearchClient = new ProductSearchClient([
    'apiKey' => $myApiKey
]);
```

AIP: https://google.aip.dev/auth/4110

- Supports the `apiKey` client option - this will be sent in using the `x-goog-api-key` header for all transports.
- If the `apiKey` and `credentials` or `credentialsConfig.keyFile` option are supplied, an exception is thrown
- If the `apiKey` option AND the `quotaProject` option are set, the `$quotaProject` is still sent in


vishwarajanand:
@bshaffer Is this still in works, else we can close this?
The AIP now doesn't seem to talk about the `apiKey`?

bshaffer:
@vishwarajanand looking at the changelog, guidance for the API Key variable was removed. Let's check with the auth team and see what the status is on supporting API keys

@TimurSadykov what are your thoughts on this one?

TimurSadykov:
I think the latest is that we don't add a generic support in auth libraries
@arithmetic1728 to confirm and reference a doc

arithmetic1728:
we are not going to add a generic support for api key due to multiple push backs

bshaffer:
Closing for now

src/ClientOptionsTrait.php
```diff
@@ -261,6 +261,10 @@ private static function determineMtlsEndpoint(string $apiEndpoint)
     private function createCredentialsWrapper($credentials, array $credentialsConfig, string $universeDomain)
     {
         if (is_null($credentials)) {
+            // If the user has explicitly set the apiKey option, use Api Key credentials
+            if (isset($credentialsConfig['apiKey'])) {
+                return ApiKeyHeaderCredentials::build($credentialsConfig);
+            }
```
clundin25, line 267: I want to call out these two conditions from the API Key Design Doc.

1. If both credentials and api key are explicitly set, return an error to the user.
2. If api key is set, do not attempt the ADC flow.

I think the first condition is missing in this code. 

Is the second condition also missing?

src/ApiKeyHeaderCredentials.php
```diff
@@ -0,0 +1,109 @@
...
+ */
+class ApiKeyHeaderCredentials implements HeaderCredentialsInterface
+{
+    private $apiKey;
+    private $quotaProject;
+
+    /**
+     * ApiKeyHeaderCredentials constructor.
+     * @param string $apiKey The API key to set in the header for the request
+     * @param string|null $quotaProject The quota project associated with the credentials.
```
clundin25, line 45: nit:

```suggestion
     * @param string|null $quotaProject The quota project associated with the API key.
```

src/ApiKeyHeaderCredentials.php
```diff
@@ -0,0 +1,109 @@
+<?php
+/*
+ * Copyright 2022 Google LLC
```
clundin25, line 3: nit:

```suggestion
 * Copyright 2024 Google LLC
```

src/ApiKeyHeaderCredentials.php
```diff
@@ -0,0 +1,109 @@
...
+ */
+namespace Google\ApiCore;
+
+/**
+ * The ApiKeyHeaderCredentials object provides a wrapper around an API key.
+ */
+class ApiKeyHeaderCredentials implements HeaderCredentialsInterface
+{
+    private $apiKey;
+    private $quotaProject;
```
clundin25, line 40: Do you foresee any issues to adding a universe domain?

I don't have any clear guidance now, but I think eventually it may be relevant

bshaffer: We do have guidance on universe domain with API Key - because there is no way to tie the two together, the universe domain is decided based on the _supplied_ client option... so, while there CAN be a universe domain, it is not possible for the API key to validate it.

src/ClientOptionsTrait.php
bshaffer: In the block you've highlighted, the logic goes like this:

 -  _IF_ `$credentials` are null (e.g. credentials have not been explicitly set), _THEN_
     - _IF_ the `apiKey` option has been set, _THEN_ create `ApiKeyHeaderCredentials`
     - _ELSE_` use `applicationDefaultCredentials` (this happens in `CredentialsWrapper::build`)

So no, the second condition is not missing.

I can add validation for the first at the top of this function.

src/HeaderCredentialsInterface.php
```diff
@@ -0,0 +1,48 @@
+<?php
+/*
+ * Copyright 2022 Google LLC
```
bshaffer, line 3:
```suggestion
 * Copyright 2024 Google LLC
```

tests/Tests/Unit/GapicClientTraitTest.php
```diff
@@ -1576,6 +1664,7 @@ class StubGapicClient
         createCredentialsWrapper as public;
         createOperationsClient as public;
         createTransport as public;
+        createCallStack as public;
```
bshaffer, line 1667:
```suggestion
```

bshaffer:
@westarle PTAL!

westarle (approved):
API Key interface looks good to me!

src/ApiKeyHeaderCredentials.php
```diff
@@ -0,0 +1,95 @@
...
+            throw new ValidationException('API key cannot be empty');
+        }
+        $this->apiKey = $apiKey;
+        $this->quotaProject = $quotaProject;
+    }
+
+    /**
+     * @return string|null The quota project associated with the credentials.
+     */
+    public function getQuotaProject(): ?string
```
Hectorhammett, line 62: Nit:
I think that it looks cleaner with a union type than with a `?`. But with the new PHP update, I don't anymore haha.

src/ApiKeyHeaderCredentials.php
```diff
@@ -0,0 +1,95 @@
...
+    {
+        return $this->quotaProject;
+    }
+
+    /**
+     * @param string $unusedAudience audiences are not supported for API keys.
+     *
+     * @return callable Callable function that returns the API key header.
+     */
+    public function getAuthorizationHeaderCallback($unusedAudience = null): ?callable
```
Hectorhammett, line 72: nit:
Type for `$unusedAudience`
```
public function getAuthorizationHeaderCallback(null|string $unusdeAudience = null): null|callable
```

src/ClientOptionsTrait.php
```diff
@@ -261,6 +261,7 @@ private static function determineMtlsEndpoint(string $apiEndpoint)
     private function createCredentialsWrapper($credentials, array $credentialsConfig, string $universeDomain)
     {
         if (is_null($credentials)) {
```
Hectorhammett, line 263: I know this was written already but all the cases return on each `if`. What if we forego the `elseif` for all of them?:
``` php
if (is_null($credentials) {
    return CredentialsWrapper::buidl($credentialsConfig, universeDomain);
}

if (is_string($credentials) || is_array($credentials)) {
    return CredentialsWrapper::build(['keyFile' => $credentials] + $credentialsConfig, $universeDomain);
}

if ($credentials instanceof FetchAuthTokenInterface) {
    authHttpHandler = $credentialsConfig['authHttpHandler'] ?? null;
    return new CredentialsWrapper($credentials, $authHttpHandler, $universeDomain);
}

if ($credentials instanceof CredentialsWrapper) {
    return $credentials;
}

     'Unexpected value in $auth option, got: ' .
     print_r($credentials, true)
);
```

I find it easier to read and more pleasing.

bshaffer: Good catch, I totally agree!

src/ApiKeyHeaderCredentials.php
bshaffer: good idea! Done

src/ApiKeyHeaderCredentials.php
bshaffer: I am definitely undecided on this. I think sometimes the `?` looks better (because it's more concise).

Hectorhammett (approved):

PR #352, opened 2021-11-29
fix: update REST transport error details

alicejli:
Currently, if invoking a PHP client library via REST transport (e.g. `$newClient = new ServiceClient(['transport' => 'rest']);`), the error details displayed are serialized improperly resulting in gibberish (or close to it) error messages. 
For instance,  the error message currently if invoking a service that a cloud project is not enabled for looks like this:

```
Google\ApiCore\ApiException: {
    "message": "Cloud Translation API has not been used in project 372459398382 before or it is disabled. Enable it by visiting https:\/\/console.developers.google.com\/apis\/api\/translate.googleapis.com\/overview?project=372459398382 then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry.",
    "code": 7,
    "status": "PERMISSION_DENIED",
    "details": [
        {
            "@type": 0,
            "data": "type.googleapis.com\/google.rpc.Help"
        },
        {
            "@type": 0,
            "data": [
                {
                    "description": "Google developers console API activation",
                    "url": "https:\/\/console.developers.google.com\/apis\/api\/translate.googleapis.com\/overview?project=372459398382"
                }
            ]
        },
        {
            "@type": 1,
            "data": "type.googleapis.com\/google.rpc.ErrorInfo"
        },
        {
            "@type": 1,
            "data": "SERVICE_DISABLED"
        },
        {
            "@type": 1,
            "data": "googleapis.com"
        },
        {
            "@type": 1,
            "data": {
                "service": "translate.googleapis.com",
                "consumer": "projects\/372459398382"
            }
        }
    ]
}
```

With this fix, the error message will instead look like this:

```
Google\ApiCore\ApiException: {
    "message": "Cloud Translation API has not been used in project 372459398382 before or it is disabled. Enable it by visiting https:\/\/console.developers.google.com\/apis\/api\/translate.googleapis.com\/overview?project=372459398382 then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry.",
    "code": 7,
    "status": "PERMISSION_DENIED",
    "details": [
        {
            "@type": "type.googleapis.com\/google.rpc.Help",
            "links": [
                {
                    "description": "Google developers console API activation",
                    "url": "https:\/\/console.developers.google.com\/apis\/api\/translate.googleapis.com\/overview?project=372459398382"
                }
            ]
        },
        {
            "@type": "type.googleapis.com\/google.rpc.ErrorInfo",
            "reason": "SERVICE_DISABLED",
            "domain": "googleapis.com",
            "metadata": {
                "consumer": "projects\/372459398382",
                "service": "translate.googleapis.com"
            }
        }
    ]
}
```

This technically could be considered a breaking change, so we may need to bundle this change into a new version. However, the current behavior is clearly buggy so not sure if this needs to be considered a breaking change.


src/ApiException.php
```diff
@@ -114,6 +114,39 @@ public static function createFromApiResponse(
...
     }
 
+    /**
+     * @param string $basicMessage
+     * @param int $rpcCode
+     * @param array|null $metadata
+     * @param \Exception $previous
+     * @return ApiException
+     */
+    public static function createFromApiResponseREST(
```
bshaffer, line 124: Nit, but we use Camelcase even for acronyms, and also, it'll be more readable if we put `Rest` before API:
```suggestion
    public static function createFromRestApiResponse(
```

src/ApiException.php
```diff
@@ -114,6 +114,39 @@ public static function createFromApiResponse(
...
+            );
+        } else {
+            return self::create(
+                $basicMessage,
+                $rpcCode,
+                $metadata,
+                $metadata,
+                $previous
+            );
+        }
```
bshaffer, line 147: implicit else can be used here because we return above
```suggestion
        }

        return self::create(
            $basicMessage,
            $rpcCode,
            $metadata,
            $metadata,
            $previous
        );
```

bshaffer (commented):
This looks great! Just two minor tweaks, and also please add some tests for this.

src/ApiException.php
```diff
@@ -115,6 +115,40 @@ public static function createFromApiResponse(
...
+        return self::create(
+            $basicMessage,
+            $rpcCode,
+            $metadata,
+            $metadata,
+            $previous
+        );
```
bshaffer, line 149: It might be better to simplify this to:
```suggestion
        return self::create(
            $basicMessage,
            $rpcCode,
            $metadata,
            is_null($metadata) ? [] : $metadata,
            $previous
        );
```

tests/Tests/Unit/ApiExceptionTest.php
```diff
@@ -203,10 +204,105 @@ public function testCreateFromApiResponse($metadata, $metadataArray) {
...
+                'message' => '',
+            ],
+        ];
+
+        return [
+            [
+                [[]], [[]],
+                [$unknownBinData, $unknownBinData],
+                [$asciiData, $asciiData],
+                [$allKnownTypesData, $allKnownTypesData]
```
bshaffer, line 275: Your test function doesnt accept a second argument so you don't need to supply the same argument twice

tests/Tests/Unit/ApiExceptionTest.php
```diff
@@ -203,10 +204,105 @@ public function testCreateFromApiResponse($metadata, $metadataArray) {
...
+            ],
+            [
+                '@type' => 'google.rpc.localizedmessage-bin',
+                'locale' => '',
+                'message' => '',
+            ],
+        ];
+
+        return [
+            [
```
bshaffer, line 271: Add a null case for metadata, since the function allows for it?
```suggestion
            [
                [null],
```

src/ApiException.php
alicejli: Ah that is much cleaner - will do, thank you!

tests/Tests/Unit/ApiExceptionTest.php
alicejli: Ah sorry I thought that an empty array (the first test) was testing for null, but I see now that `[]` is different from `[null]`. Done.

src/ApiException.php
alicejli: Much cleaner, thank you! Done.

tests/Tests/Unit/ApiExceptionTest.php
alicejli: Oops thank you, done.

bshaffer (approved):

PR #354, opened 2021-12-03
feat: improve errors by hoisting ErrorInfo fields directly in error message

alicejli:
As part of the overall GAPIC Error Details Improvements project, this feature improves the usability of the error information by including three additional fields (if available) directly in the error itself [e.g. ErrorInfo](https://cloud.google.com/apis/design/errors#error_info) that offer more detailed information for troubleshooting. These three fields are `domain`, `reason`, and `metadata`. 

This PR also includes getters for those fields so users may access them directly. 

This PR is dependent on https://github.com/googleapis/common-protos-php/pull/27 being merged which will not happen until post-freeze, sometime in January.

Please ping me for internal docs if desired. 

**Current REST sample error:**

```
PHP Fatal error:  Uncaught Google\ApiCore\ApiException: {
    "message": "Cloud Translation API has not been used in project <id> before or it is disabled. Enable it by visiting https:\/\/console.developers.google.com\/apis\/api\/translate.googleapis.com\/overview?project=<id> then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry.",
    "code": 7,
    "status": "PERMISSION_DENIED",
    "details": [
        {
            "@type": "type.googleapis.com\/google.rpc.Help",
            "links": [
                {
                    "description": "Google developers console API activation",
                    "url": "https:\/\/console.developers.google.com\/apis\/api\/translate.googleapis.com\/overview?project=<id>"
                }
            ]
        },
        {
            "@type": "type.googleapis.com\/google.rpc.ErrorInfo",
            "reason": "SERVICE_DISABLED",
            "domain": "googleapis.com",
            "metadata": {
                "consumer": "projects\/<id>",
                "service": "translate.googleapis.com"
            }
        }
    ]
}
```

**With this change, REST sample error:**

```
Google\ApiCore\ApiException: {
    "message": "Cloud Translation API has not been used in project <id> before or it is disabled. Enable it by visiting https:\/\/console.developers.google.com\/apis\/api\/translate.googleapis.com\/overview?project=<id> then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry.",
    "domain": "googleapis.com",
    "reason": "SERVICE_DISABLED",
    "errorInfoMetadata": {
        "consumer": "projects\/<id>",
        "service": "translate.googleapis.com"
    },
    "code": 7,
    "status": "PERMISSION_DENIED",
    "details": [
        {
            "@type": "type.googleapis.com\/google.rpc.Help",
            "links": [
                {
                    "description": "Google developers console API activation",
                    "url": "https:\/\/console.developers.google.com\/apis\/api\/translate.googleapis.com\/overview?project=<id>"
                }
            ]
        },
        {
            "@type": "type.googleapis.com\/google.rpc.ErrorInfo",
            "reason": "SERVICE_DISABLED",
            "domain": "googleapis.com",
            "metadata": {
                "consumer": "projects\/<id>",
                "service": "translate.googleapis.com"
            }
        }
    ]
}
```
**Current GRPC sample error:**

```
Google\ApiCore\ApiException: {
    "message": "Cloud Translation API has not been used in project <id> before or it is disabled. Enable it by visiting https:\/\/console.developers.google.com\/apis\/api\/translate.googleapis.com\/overview?project=<id> then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry.",
    "code": 7,
    "status": "PERMISSION_DENIED",
    "details": [
        {
            "@type": "google.rpc.errorinfo-bin",
            "data": "<Unknown Binary Data>"
            }
        },
        {
            "@type": "google.rpc.help-bin",
            "links": [
                {
                    "description": "Google developers console API activation",
                    "url": "https:\/\/console.developers.google.com\/apis\/api\/translate.googleapis.com\/overview?project=<id>"
                }
            ]
        },
        {
            "@type": "grpc-status-details-bin",
            "data": "<Unknown Binary Data>"
        },
        {
            "@type": "grpc-server-stats-bin",
            "data": "<Unknown Binary Data>"
        }
    ]
}
```
**With this change, GRPC sample error:**

```
Google\ApiCore\ApiException: {
    "message": "Cloud Translation API has not been used in project <id> before or it is disabled. Enable it by visiting https:\/\/console.developers.google.com\/apis\/api\/translate.googleapis.com\/overview?project=<id> then retry. If you enabled this API recently, wait a few minutes for the action to propagate to our systems and retry.",
    "domain": "googleapis.com",
    "reason": "SERVICE_DISABLED",
    "errorInfoMetadata": {
        "service": "translate.googleapis.com",
        "consumer": "projects\/<id>"
    },
    "code": 7,
    "status": "PERMISSION_DENIED",
    "details": [
        {
            "@type": "google.rpc.errorinfo-bin",
            "reason": "SERVICE_DISABLED",
            "domain": "googleapis.com",
            "metadata": {
                "service": "translate.googleapis.com",
                "consumer": "projects\/<id>"
            }
        },
        {
            "@type": "google.rpc.help-bin",
            "links": [
                {
                    "description": "Google developers console API activation",
                    "url": "https:\/\/console.developers.google.com\/apis\/api\/translate.googleapis.com\/overview?project=<id>"
                }
            ]
        },
        {
            "@type": "grpc-status-details-bin",
            "data": "<Unknown Binary Data>"
        },
        {
            "@type": "grpc-server-stats-bin",
            "data": "<Unknown Binary Data>"
        }
    ]
}
```


alicejli:
Requesting @noahdietz , @bshaffer  as PHP owners, and @summer-ji-eng as Error Improvement project lead for review. This is a non urgent review as we will not be able to release this until post freeze. Thanks!

src/ApiException.php
```diff
@@ -151,13 +256,26 @@ public static function createFromRestApiResponse(
...
     {
+        $containsErrorInfo = self::containsErrorInfo($decodedMetadata);
         $rpcStatus = ApiStatus::statusFromRpcCode($rpcCode);
-        $messageData = [
-            'message' => $basicMessage,
-            'code' => $rpcCode,
-            'status' => $rpcStatus,
-            'details' => $decodedMetadata
-        ];
+        if ($containsErrorInfo['containsErrorInfo'] === true) {
```
noahdietz, line 261: Don't need to compare to `true`. If the key doesn't exist it will return `null` I believe which is falsey.

src/ApiException.php
```diff
@@ -138,9 +216,36 @@ public static function createFromRestApiResponse(
...
+     */
+    public static function containsErrorInfo(array $decodedMetadata)
+    {
+        $errorInfo = array();
+        $errorInfo['containsErrorInfo'] = false;
+        if (empty($decodedMetadata)) {
+            return $errorInfo;
+        }
+        foreach ($decodedMetadata as $key=>$value) {
+            if (array_key_exists('@type', $value) and ($value['@type'] === 'google.rpc.errorinfo-bin' or $value['@type'] === 'type.googleapis.com/google.rpc.ErrorInfo') or (array_key_exists('reason', $value) && array_key_exists('domain', $value) && array_key_exists('metadata', $value))) {
```
noahdietz, line 236: Nit: can we not mix `and`/`or` with `&&`/`||`. Let's use one or the other?

noahdietz, line 236: So this is checking "Any containing an ErrorInfo OR array with ErrorInfo fields" ?

For readability, you might consider breaking these into two variables like so:
```php
$isErrorInfoAny = array_key_exists('@type', $value) &&
    ($value['@type'] === 'google.rpc.errorinfo-bin' ||
     $value['@type'] === 'type.googleapis.com/google.rpc.ErrorInfo');
$isErrorInfoArray = array_key_exists('reason', $value) &&
    array_key_exists('domain', $value) &&
    array_key_exists('metadata', $value);

if ($isErrorInfoAny || $isErrorInfoArray) {
```

src/ApiException.php
```diff
@@ -78,6 +78,84 @@ public function getStatus()
...
+            if ($detailsRpcStatus['containsErrorInfo']) {
+                return $metadataRpcStatus[$detailsRpcStatus['keyOfErrorInfo']]['domain'];
+            }
+        } elseif (is_Array($metadata)) {
+            // For REST-based responses, the $metadata does not need to be decoded.
+            $detailsRest = self::containsErrorInfo($metadata);
+            if ($detailsRest['containsErrorInfo']) {
+                return $metadata[$detailsRest['keyOfErrorInfo']]['domain'];
+            }
+        }
```
noahdietz, line 97: This `if (is_object) elseif (is_Array)` logic as well as the decoding logic is in multiple spots, that means it is a good thing to try to either 1. pull into a function 2. process ahead of time so that these getters just need to access a variable.

Can we try to refactor this code into one of the two approaches?

noahdietz (changes requested):
A few comments to start. Can you also run the php formatting tool? Should be in the readme somewhere.

src/ApiException.php
alicejli: Makes sense - decided to go with the latter as that should be consistent with the rest.

src/ApiException.php
alicejli: Good point. I've refactored into a separate function.

src/ApiException.php
```diff
@@ -151,13 +230,26 @@ public static function createFromRestApiResponse(
...
+                'details' => $decodedMetadata
+            ];
+        } else {
+            $messageData = [
+                'message' => $basicMessage,
+                'code' => $rpcCode,
+                'status' => $rpcStatus,
+                'details' => $decodedMetadata
+            ];
+        }
```
noahdietz, line 252:
```suggestion
        $messageData = [
                'message' => $basicMessage,
                'code' => $rpcCode,
                'status' => $rpcStatus,
                'details' => $decodedMetadata
        ];
        if ($containsErrorInfo['containsErrorInfo']) {
            $messageData = array_merge($messageData, [
                'domain' => $containsErrorInfo['domain'],
                'reason' => $containsErrorInfo['reason'],
                'errorInfoMetadata' => $containsErrorInfo['errorInfoMetadata'],
            ]);
        }
```

src/ApiException.php
```diff
@@ -138,9 +182,44 @@ public static function createFromRestApiResponse(
...
+        $errorInfo['containsErrorInfo'] = false;
+        if (empty($decodedMetadata)) {
+            return $errorInfo;
+        }
+        foreach ($decodedMetadata as $value) {
+            $isErrorInfoAny = array_key_exists('@type', $value) && ($value['@type'] === 'google.rpc.errorinfo-bin'
+                || $value['@type'] === 'type.googleapis.com/google.rpc.ErrorInfo');
+            $isErrorInfoArray = array_key_exists('reason', $value) && array_key_exists('domain', $value)
+                && array_key_exists('metadata', $value);
+            if ($isErrorInfoAny || $isErrorInfoArray) {
```
noahdietz, line 208: I'm a little confused on how this works. I read it as:

      If the error metadata has type ...errorinfo-bin or ...ErrorInfo OR if it has *all* of the keys we like, access all of those keys

By this logic, the type URL here seems unnecessary since an error info message will always have those fields (or most of them, see other comment).

Inversely, when will the `@type` *not` be present or *not* be `errorinfo-bin`/`ErrorInfo`, but still have the fields ?

src/ApiException.php
```diff
@@ -138,9 +182,44 @@ public static function createFromRestApiResponse(
...
+        $errorInfo = [];
+        $errorInfo['containsErrorInfo'] = false;
+        if (empty($decodedMetadata)) {
+            return $errorInfo;
+        }
+        foreach ($decodedMetadata as $value) {
+            $isErrorInfoAny = array_key_exists('@type', $value) && ($value['@type'] === 'google.rpc.errorinfo-bin'
+                || $value['@type'] === 'type.googleapis.com/google.rpc.ErrorInfo');
+            $isErrorInfoArray = array_key_exists('reason', $value) && array_key_exists('domain', $value)
+                && array_key_exists('metadata', $value);
```
noahdietz, line 207: Must an ErrorInfo have all fields populated? What if one field was omitted? I don't think metadata is required to be set, although I think domain and reason are.

src/ApiException.php
```diff
@@ -78,6 +78,50 @@ public function getStatus()
...
+        } elseif (is_Array($metadata)) {
+            // For REST-based responses, the $metadata does not need to be decoded.
+            $detailsRest = self::containsErrorInfo($metadata);
+            if ($detailsRest['containsErrorInfo']) {
+                return $detailsRest;
+            }
+        }
+        $metadataGrpc = Serializer::decodeMetadata($metadata);
+        $detailsGrpc = self::containsErrorInfo($metadataGrpc);
+        return $detailsGrpc['containsErrorInfo'] ? $detailsGrpc : null;
```
noahdietz, line 101:
```suggestion
        $details = [];
        if (is_object($metadata)) {
            $metadataRpcStatus = Serializer::decodeAnyMessages($metadata);
            $details = self::containsErrorInfo($metadataRpcStatus);
        } elseif (is_Array($metadata)) {
            // For REST-based responses, the $metadata does not need to be decoded.
            $details = self::containsErrorInfo($metadata);
        } else {
            $metadataGrpc = Serializer::decodeMetadata($metadata);
            $details = self::containsErrorInfo($metadataGrpc);
        }
       
        return $details['containsErrorInfo'] ? $details : null;
```

noahdietz: Don't feel like you need to accept the suggestion, you can just copy-paste it if you want to tweak/try it

src/ApiException.php
```diff
@@ -78,6 +78,50 @@ public function getStatus()
...
+    }
+    
+    public function getDomain()
+    {
+        $metadata = $this->metadata;
+        return isset(self::decodeMetadataErrorInfo($metadata)['domain'])
+        ? self::decodeMetadataErrorInfo($metadata)['domain'] : null;
```
noahdietz, line 108: Instead of invoking this twice, should the result just be captured in a variable then evaluated in the ternary operator? This comment goes for the other accessors.

src/ApiException.php
alicejli: Done

src/ApiException.php
alicejli: You are totally right - I think I was getting confused between the different types of metadata, but since this function only looks at decoded metadata anyway, then we only need to check for the presence of the keys we want. I also confirmed with Summer that if ErrorInfo exists, then all three keys should be set. I updated the function to just check that `reason` exists for simplicity.

src/ApiException.php
alicejli: Confirmed with Summer that if ErrorInfo exists, all three fields should exist. But have updated function to just check for `reason` for simplicity.

src/ApiException.php
alicejli: Thanks for the suggestion! I took it and it helped me fix up the function such that all the tests pass, so looks like the original way it was working was creating the seg fault error.

src/ApiException.php
alicejli: Makes sense - updated!

src/ApiException.php
```diff
@@ -78,6 +78,49 @@ public function getStatus()
         return $this->status;
     }
 
+    // Returns null if metadata does not contain error info, or return containsErrorInfo() array
+    // if the metadata does contain error info.
+    public static function decodeMetadataErrorInfo($metadata)
+    {
+        $details = [];
+        // ApiExceptions created from RPC status have metadata that is an array of objects.
+        if (is_object(reset($metadata))) {
```
noahdietz, line 87: What are you trying to do by calling `reset`? Seems to me manipulating the array's internal pointer is probably not a great idea ([docs](https://www.php.net/manual/en/function.reset.php)).

noahdietz (changes requested):
Can you add php doc comments to all of the added public APIs?

src/ApiException.php
alicejli: I was just trying to get the first element of the array; so to your point, I don't actually need the internal pointer to be changed. I could update this to `is_object($metadata[array_key_first($metadata)])` (and similar for line 93) - unless there's a better way to check this?

`Can you add php doc comments to all of the added public APIs?` - sorry, what does that mean?

src/ApiException.php
```diff
@@ -138,9 +181,41 @@ public static function createFromRestApiResponse(
...
+     */
+    public static function containsErrorInfo(array $decodedMetadata)
+    {
+        $errorInfo = [];
+        $errorInfo['containsErrorInfo'] = false;
+        if (empty($decodedMetadata)) {
+            return $errorInfo;
+        }
+        foreach ($decodedMetadata as $value) {
+            $isErrorInfoArray = array_key_exists('reason', $value);
```
summer-ji-eng, line 203: Not pretty sure what if other type has `reason` as key. Here is the code I used in Typescript, maybe it helps
```
const decodedDetail = this.decodeProtobufAny(detail);
          details.push(decodedDetail);
          if (detail.type_url === 'type.googleapis.com/google.rpc.ErrorInfo') {
            errorInfo = decodedDetail as unknown as ErrorInfo;
          }
          ```
https://github.com/googleapis/gax-nodejs/blob/2f3ac59564f112d501ceab5f3a319415e6cb80ef/src/googleError.ts#L211-L214

File: tests/Tests/Unit/Transport/RestTransportTest.php [hunk 11]
```diff
@@ -318,7 +319,6 @@ public function testStartServerStreamingCallThrowsRequestException()
 
         $this->getTransport($httpHandler, $apiEndpoint)
             ->startServerStreamingCall($this->call, []);
-
```
summer-ji-eng, line 321: Probably you don't want to add this line.

tests/Tests/Unit/Transport/RestTransportTest.php
```diff
@@ -32,7 +32,6 @@
```
summer-ji-eng, line 32: Curious test on this file. Looks like ApiException has been removed from this file.

summer-ji-eng (commented):
Does this implementation cover for REST mode?

src/ApiException.php
```diff
@@ -78,6 +78,49 @@ public function getStatus()
         return $this->status;
     }
 
+    // Returns null if metadata does not contain error info, or return containsErrorInfo() array
+    // if the metadata does contain error info.
+    public static function decodeMetadataErrorInfo($metadata)
```
bshaffer, line 83: Two things here:
1. Since this is only being called internally and surfaced by other public methods, it should be private
2. As mentioned by Noah, all methods should use PHP doc! For example (although I am not sure if this is accurate):

```suggestion
    /**
     * Returns null if metadata does not contain error info, or return containsErrorInfo() array
     * if the metadata does contain error info.
     *
     * @param array $metadata
     * @return array|null
     *./
    private static function decodeMetadataErrorInfo(array $metadata)
```

src/ApiException.php
bshaffer: So what is the type of first element of the array for gRPC metadata? It isn't an object?

Personally I think in this case, `reset` is better. 

I added an example above wrt PHPDoc

src/ApiException.php
```diff
@@ -138,9 +181,41 @@ public static function createFromRestApiResponse(
...
+     * Checks if decoded metadata includes errorInfo message.
+     * @param array $decodedMetadata
+     * @return array $errorInfo {
+     *     @type boolean $containsErrorInfo
+     *     @type string $reason
+     *     @type string $domain
+     *     @type array $errorInfoMetadata
+     * }
+     */
+    public static function containsErrorInfo(array $decodedMetadata)
```
bshaffer, line 195: I think this also should _probably_ be private. I see no reason to expose this to the end user
```suggestion
    private static function containsErrorInfo(array $decodedMetadata)
```
```suggestion
    public static function containsErrorInfo(array $decodedMetadata)
```

tests/Tests/Unit/Transport/RestTransportTest.php
bshaffer: The import statement isn't being used in the file so it's better to remove it. I imagine this got caught by a style checker

tests/Tests/Unit/Transport/RestTransportTest.php
bshaffer: Same here... looks like most of the changes in this file are from a style checker

src/ApiException.php
alicejli: Makes sense - updated!

src/ApiException.php
alicejli: Good point - I've updated it to check for `reason`, `domain`, and `metadata` since all three keys should exist if ErrorInfo exists.

src/ApiException.php
```diff
@@ -138,9 +201,42 @@ public static function createFromRestApiResponse(
...
+            return $errorInfo;
+        }
+        foreach ($decodedMetadata as $value) {
+            $isErrorInfoArray = isset($value['reason']) && isset($value['domain']) && isset($value['metadata']);
+            if ($isErrorInfoArray) {
+                $errorInfo['containsErrorInfo'] = true;
+                $errorInfo['reason'] = $value['reason'];
+                $errorInfo['domain'] = $value['domain'];
+                $errorInfo['errorInfoMetadata'] = $value['metadata'];
+                return $errorInfo;
```
noahdietz, line 230: Nit: I might just use a `break` here and let it fall through to the return below.

src/ApiException.php
```diff
@@ -151,13 +247,21 @@ public static function createFromRestApiResponse(
...
             'code' => $rpcCode,
             'status' => $rpcStatus,
             'details' => $decodedMetadata
         ];
+        if ($containsErrorInfo['containsErrorInfo']) {
+            $messageData = array_merge([
+                'domain' => $containsErrorInfo['domain'],
+                'reason' => $containsErrorInfo['reason'],
+                'errorInfoMetadata' => $containsErrorInfo['errorInfoMetadata'],
+            ], $messageData);
```
noahdietz, line 263: Nit: Alternatively, you could `unset($containsErrorInfo['containsErrorInfo'])` and then `array_merge($containsErrorInfo, $messageData)`

src/ApiException.php
```diff
@@ -171,7 +275,7 @@ private static function create($basicMessage, $rpcCode, $metadata, array $decode
             'basicMessage' => $basicMessage,
         ]);
     }
-
+
```
noahdietz, line 278: rogue whitespace 🏴‍☠️ 🏴‍☠️ 🏴‍☠️ ! Ah!

src/ApiException.php
```diff
@@ -78,6 +78,69 @@ public function getStatus()
...
+        $details = [];
+        // ApiExceptions created from RPC status have metadata that is an array of objects.
+        if (is_object(reset($metadata))) {
+            $metadataRpcStatus = Serializer::decodeAnyMessages($metadata);
+            $details = self::containsErrorInfo($metadataRpcStatus);
+        } else {
+            // For GRPC-based responses, the $metadata needs to be decoded.
+            $metadataGrpc = Serializer::decodeMetadata($metadata);
+            $details = self::containsErrorInfo($metadataGrpc);
+            // For REST-based responses, the $metadata does not need to be decoded.
```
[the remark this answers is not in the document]
alicejli: Yep you understood correctly; I see that that is confusing. Have updated to the `if` (check if created from RpcStatus), `else if` (check if REST), `else` (assume GRPC) pattern.

src/ApiException.php
```diff
@@ -138,9 +201,42 @@ public static function createFromRestApiResponse(
...
+     *     @type boolean $containsErrorInfo
+     *     @type string $reason
+     *     @type string $domain
+     *     @type array $errorInfoMetadata
+     * }
+     */
+    private static function containsErrorInfo(array $decodedMetadata)
+    {
+        $errorInfo = [];
+        $errorInfo['containsErrorInfo'] = false;
```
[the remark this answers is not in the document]
alicejli: thanks! done

src/ApiException.php
alicejli: that's neat - done!

src/ApiException.php
alicejli: >.< removed

src/ApiException.php
alicejli: makes sense - done

noahdietz (commented):
LGTM! Great stuff Alice. I will let @bshaffer review/Approve.

summer-ji-eng:
Thanks Alice for your work. LGTM.
Not blocker, but please add the integration test for grpc and rest.

summer-ji-eng (approved):

bshaffer (changes requested):
Some suggestions for cleanup, otherwise looks great

src/ApiException.php
```diff
@@ -78,6 +78,64 @@ public function getStatus()
...
+     * if the metadata does contain error info.
+     * @param array $metadata
+     * @return array|null $details {
+     *     @type bool $containsErrorInfo
+     *     @type string|null $reason
+     *     @type string|null $domain
+     *     @type array|null $errorInfoMetadata
+     * }
+     */
+    private static function decodeMetadataErrorInfo($metadata)
```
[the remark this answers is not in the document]
alicejli: makes sense! done

src/ApiException.php
```diff
@@ -151,13 +244,18 @@ public static function createFromRestApiResponse(
...
     {
+        $containsErrorInfo = self::containsErrorInfo($decodedMetadata);
         $rpcStatus = ApiStatus::statusFromRpcCode($rpcCode);
         $messageData = [
             'message' => $basicMessage,
             'code' => $rpcCode,
             'status' => $rpcStatus,
             'details' => $decodedMetadata
         ];
+        if ($containsErrorInfo['containsErrorInfo']) {
```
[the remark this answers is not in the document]
alicejli: Updated based on comments below!

src/ApiException.php
```diff
@@ -138,9 +196,44 @@ public static function createFromRestApiResponse(
...
+     * }
+     */
+    private static function containsErrorInfo(array $decodedMetadata)
+    {
+        $errorInfo = [];
+        $errorInfo = [
+            'containsErrorInfo' => false
+        ];
+        if (empty($decodedMetadata)) {
+            return $errorInfo;
```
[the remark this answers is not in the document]
alicejli: done

src/ApiException.php
```diff
@@ -138,9 +196,44 @@ public static function createFromRestApiResponse(
...
+        }
+        foreach ($decodedMetadata as $value) {
+            $isErrorInfoArray = isset($value['reason']) && isset($value['domain']) && isset($value['metadata']);
+            if ($isErrorInfoArray) {
+                $errorInfo['containsErrorInfo'] = true;
+                $errorInfo['reason'] = $value['reason'];
+                $errorInfo['domain'] = $value['domain'];
+                $errorInfo['errorInfoMetadata'] = $value['metadata'];
+                break;
```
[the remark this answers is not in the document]
alicejli: Thank you! Done

src/ApiException.php
```diff
@@ -78,6 +78,64 @@ public function getStatus()
...
+        return $details['containsErrorInfo'] ? $details : null;
+    }
+
+    /**
+     * Returns the `reason` in ErrorInfo for an exception, or null if there is no ErrorInfo.
+     * @return string|null $reason
+     */
+    public function getReason()
+    {
+        $decodedMetadata = self::decodeMetadataErrorInfo($this->metadata);
```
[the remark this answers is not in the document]
alicejli: makes sense - done

bshaffer (approved):

VERSION
```diff
@@ -1 +1 @@
-1.11.4
+1.11.4
```
noahdietz, line 1: add a newline please

noahdietz (approved):

PR #357, opened 2022-01-14
chore: allow for google/common-protos 2.0

alicejli:
Update to allow for `google/common-protos 2.0`. We are keeping `google/common-protos 1.0` for backwards-compatibility purposes.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,8 +12,8 @@
         "grpc/grpc": "^1.13",
         "google/protobuf": "^3.12.2",
         "guzzlehttp/promises": "^1.3",
-        "guzzlehttp/psr7": "^1.7.0|^2",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -13,7 +13,7 @@
         "google/protobuf": "^3.12.2",
         "guzzlehttp/promises": "^1.3",
         "guzzlehttp/psr7": "^1.7.0|^2",
-        "google/common-protos": "^1.0"
+        "google/common-protos": "^1.0|^2.0"
```
bshaffer, line 16: Nit: The [double pipe](https://getcomposer.org/doc/articles/versions.md#version-range) is recommended
```suggestion
        "google/common-protos": "^1.0||^2.0"
```

bshaffer (approved):

PR #362, opened 2022-01-20
fix: relax ServerStreamingCallWrapper construct param type

noahdietz:
Relaxes the `construct` param for the `ServerStreamingCallWrapper` to be untyped, but document that it expects a `Grpc\ServerStreamingCall` or a `Grpc\Gcp\GCPServerStreamCall`.

Also updated the `VERSION` to `1.11.2` ahead of the release.


src/Transport/Grpc/ServerStreamingCallWrapper.php
```diff
@@ -36,7 +36,6 @@
 use Google\ApiCore\ServerStreamingCallInterface;
 use Google\Rpc\Code;
 use Google\Rpc\Status;
-use Grpc\ServerStreamingCall;
```
bshaffer, line 39: Keep this use, and add 

```php
use Grpc\Gcp\GCPServerStreamCall;
```

noahdietz: Done.

bshaffer (approved):

Issue #370, opened 2022-01-26
Unimplemented REST method returns configuration error since 1.11.0


PierrickVoulet:
#### Environment details

  - OS: Linux 64
  - PHP version: 8.0.1
  - Package name and version: gax-php, 1.11.0-1.11.3

#### Steps to reproduce

1. Checkout the source code of [`google-ads-php`](https://github.com/googleads/google-ads-php).
2. Go through the [getting started instructions](https://github.com/googleads/google-ads-php#getting-started).
3. Change the version of the dependency `gax-php` to `1.10.0` in the [`composer.json`](https://github.com/googleads/google-ads-php/blob/main/composer.json#L7)
4. Add `->withTransport('rest')` at [the line number 56 of the code example `GetCampaign`](https://github.com/googleads/google-ads-php/blob/main/examples/BasicOperations/GetCampaigns.php#L56).
5. Run `composer update`
6. Run the code example [`GetCampaign`](https://github.com/googleads/google-ads-php/blob/main/examples/BasicOperations/GetCampaigns.php).
7. It fails with the following error

```php
PHP Fatal error:  Uncaught BadMethodCallException: Streaming calls are not supported while using the REST transport. in /usr/local/google/home/pierrick/git/google-ads-php/vendor/google/gax/src/Transport/HttpUnaryTransportTrait.php:148
Stack trace:
#0 /usr/local/google/home/pierrick/git/google-ads-php/vendor/google/gax/src/Transport/HttpUnaryTransportTrait.php(64): Google\ApiCore\Transport\RestTransport->throwUnsupportedException()
#1 /usr/local/google/home/pierrick/git/google-ads-php/vendor/google/gax/src/GapicClientTrait.php(608): Google\ApiCore\Transport\RestTransport->startServerStreamingCall()
#2 /usr/local/google/home/pierrick/git/google-ads-php/vendor/google/gax/src/Middleware/CredentialsWrapperMiddleware.php(61): Google\Ads\GoogleAds\V9\Services\Gapic\GoogleAdsServiceGapicClient->Google\ApiCore\{closure}()
#3 /usr/local/google/home/pierrick/git/google-ads-php/vendor/google/gax/src/Middleware/FixedHeaderMiddleware.php(66): Google\ApiCore\Middleware\CredentialsWrapperMiddleware->__invoke()
[...]

PR #371, opened 2022-01-26
fix: maintain forwards compatibility with REST server streaming

bshaffer:
This is just an idea on how to maintain forwards compatibility with the new REST server streaming methods
see #370


src/Transport/RestTransport.php
```diff
@@ -153,6 +155,12 @@ public function startServerStreamingCall(Call $call, array $options)
             throw new \InvalidArgumentException('A message is required for ServerStreaming calls.');
         }
 
+        // Maintain forwards compatibility with older GAPIC clients not configured for REST server streaming
+        // @see https://github.com/googleapis/gax-php/issues/370
+        if (!$this->requestBuilder->pathExists($call->getMethod())) {
+            throw new BadMethodCallException('Streaming calls are not supported while using the REST transport');
```
noahdietz, line 161: Alternatively, we can call the `HttpUnaryTransportTrait:startServerStreamingCall` and let it throw what it was doing before?

```php
use HttpUnaryTransportTrait {
    startServerStreamingCall as protected unsupportedServerStreamingCall;
}

$this->unsupportedServerStreamingCall($call, $options);
```

noahdietz (approved):

src/Transport/RestTransport.php
bshaffer: 🤯

noahdietz: Totally Googled this because I wanted to know how to call a trait's method like a parent class' that was overloaded. Guess this is how one does that (`parent::` doesn't work because a `trait` isn't a class, of course).

bshaffer: done!

bshaffer: I didn't know you could change the visibility... that's pretty nice

Issue #234, opened 2019-04-01
Support for logging when making REST calls


fiboknacky:
In the [Google Ads API](https://github.com/googleads/google-ads-php) client library, some people may select to make a call using REST instead of gRPC.
Or with some reasons, they cannot use the gRPC extension on their system.

In that case, we probably still want to them to be able to use the logging feature so they can troubleshoot the issues and report them to us, when needed.

Knack

PR #379, opened 2022-03-21
feat: add rest interceptor support

alicejli:
Interceptors are a gRPC feature that wraps RPCs in
continuation-passing-style pre and post method custom functions.
These can be used e.g. for logging, local caching, and tweaking
metadata.

This PR adds interceptor-like functionality to the REST transport. Note that when an end user constructs the REST transport, they should either pass in the interceptors they want to use, OR pass through a custom httpHandler which already includes the interceptors they want to use. The REST transport will ignore any passed interceptors if there is httpHandler passed through the transport config. 

The REST transport interceptors differ from GRPC transport interceptors in a few ways:

* They are not continuations. Instead, they mimic `Middleware` [https://docs.guzzlephp.org/en/stable/handlers-and-middleware.html] which is a concept that already exists in Guzzle and should be familiar to the PHP community.

* Existing gRPC interceptors and the new REST interceptors are not
composable or interoperable.

Should enable https://github.com/googleapis/gax-php/issues/234, although will require some work from the Ads team to implement.

### DIFF
--- a/src/Transport/RestTransport.php
+++ b/src/Transport/RestTransport.php
@@ -42,6 +42,8 @@ use Google\ApiCore\ValidationTrait;
 use Google\Protobuf\Internal\Message;
 use GuzzleHttp\Exception\RequestException;
 use Psr\Http\Message\ResponseInterface;
+use GuzzleHttp\HandlerStack;
+use GuzzleHttp\Client;
 
 /**
  * A REST based transport implementation.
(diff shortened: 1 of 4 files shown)


src/Transport/RestTransport.php
```diff
@@ -91,10 +97,21 @@ public static function build($apiEndpoint, $restConfigPath, array $config = [])
...
+        $interceptors = $config['interceptors'];
+        if (!$config['httpHandler'] && !empty($interceptors)) {
+            $stack = HandlerStack::create();
+            foreach ($interceptors as $interceptor) {
+                $stack->push($interceptor);
+            }
+            $client = new Client(['handler' => $stack]);
+            $httpHandler = self::buildHttpHandlerAsync($client);
+        } else {
+            $httpHandler = $config['httpHandler'] ?: self::buildHttpHandlerAsync();
```
noahdietz, line 113:
```suggestion
            $httpHandler = $config['httpHandler'] ? $config['httpHandler'] : self::buildHttpHandlerAsync();
```
I think this is missing the `true` return value, unless there is some magic that actually assumes the true case returns the result of the ternary expression.

PR #384, opened 2022-04-28
feat: support custom key values in agent header

PierrickVoulet:


src/GapicClientTrait.php
```diff
@@ -162,6 +162,7 @@ private function buildClientOptions(array $options)
             'gapicVersion' => self::getGapicVersion($options),
             'libName' => null,
             'libVersion' => null,
+            'customAgentHeaderKeyValues' => [],
```
bshaffer, line 165: I would make this simpler and just say `customUserAgentValues`

src/AgentHeader.php
```diff
@@ -52,6 +52,9 @@ class AgentHeader
      *     @type string $grpcVersion the gRPC version.
      *     @type string $restVersion the REST transport version (typically same as the
      *           ApiCore version).
+     *     @type array $customKeyValues custom key values to add to the agent header info. They are
+     *           added after the default ones. Any key that has the same name as a default one is
+     *           ignored (no override).
```
bshaffer, line 57: I'd also suggest just keeping this simple and making it `customValues`

src/AgentHeader.php
```diff
@@ -107,6 +111,15 @@ public static function buildAgentHeader($headerInfo)
             : $apiCoreVersion;
         $metricsHeaders['rest'] = $restVersion;
 
+        if (isset($headerInfo['customKeyValues'])) {
```
bshaffer, line 114: Since these custom values are fundamentally different from the existing `$headerInfo` values, I would prefer to add a separate argument to `buildAgentHeader`:

```php
public static function buildAgentHeader($headerInfo, array $additionalValues = [])
```

bshaffer (commented):
This all looks good, but I'd l should at least look into the other options first.

1. Would it make sense to support the custom values you're hoping to add instead of allowing custom ones?
2. Would it be better to just override this function in your GAPIC clients and set the UserAgent header you need? As an example:

```php
class MyClient extends MyGapicClient
{
    public function setClientOptions(array $options)
    {
        parent::setClientOptions($options);
        $this->userAgent .= ' my/custom user/agent key/values';
    }
```

bshaffer:
> 1. Our need is to add the protobuf version being used by each request 

As the protobuf version is generally useful, I am definitely in favor of adding it for all GAPIC clients!

PierrickVoulet:
Closing this because #387 should solve my issue.

PR #389, opened 2022-06-12
feat: Ensure support for PHP >= 7.4

jeromegamez:
Hey there 👋,

This PR aims to ensure support for PHP versions from PHP 7.4 up to 8.1 (I tested the changes locally with PHP 8.1).

I do think that it _does_ already work for the most part (otherwise people would already have complained 😅), but when I ran the tests, I was greeted with a plethora of errors, so I thought it could be useful to actually test the library with all the versions.

I took the same approach as it is already implemented in https://github.com/googleapis/google-cloud-php (using [yoast/phpunit-polyfills](https://github.com/Yoast/PHPUnit-Polyfills)) and can confirm that I can run the tests with PHP 8.1 _(again: on my machine™)_ without error/failure/warning/deprecation/risky reports.

### Changes

- Integrated [yoast/phpunit-polyfills](https://github.com/Yoast/PHPUnit-Polyfills)
- Replaced `@expected*` annotations with `->expect*` method calls
- Added null-checks because of `Deprecated: explode(): Passing null to parameter #2 ($string) of type string is deprecated` and similar deprecation notices

I'm curious what it will look like when they run in GitHub Actions 😅, and if you're even willing to consider merging the changes.

The main reason I'm here that while working on [adding Monolog 3.0 support to `google/cloud`](https://github.com/googleapis/google-cloud-php/pull/5334), which requires PHP 8.1, I ran into the deprecation warning 

> Calling static trait method Google\ApiCore\Testing\MockStubTrait::stripStatusFromResponses is deprecated, it should only be called on a class using the trait

which is addressed in the second commit of this PR.

:octocat: 🙏


src/Testing/GeneratedTest.php
```diff
@@ -35,13 +35,16 @@
...
-use PHPUnit\Framework\TestCase;
+use Yoast\PHPUnitPolyfills\TestCases\TestCase;
 
 abstract class GeneratedTest extends TestCase
 {
     public function assertProtobufEquals(&$expected, &$actual)
     {
         if ($expected === $actual) {
+            // This is not neede but reduces the number of risky "This test did not perform any assertions" tests
+            $this->assertSame($expected, $actual);
```
bshaffer, line 46: Good catch! I hate those!

src/ResourceTemplate/Parser.php
```diff
@@ -215,11 +215,15 @@ private static function parseError($path, $index, $reason)
      * Check if $literal is a valid segment literal. Segment literals may only contain numbers,
      * letters, and any of the following: .-~_
      *
-     * @param string $literal
+     * @param string|null $literal
```
bshaffer, line 218: Do we really want to accept `null` here? It seems like anyone passing in `null` here is doing so erroneously. 

We could change the signature to `string $literal` without breaking backwards compatibility: string casting would be taking place in `preg_match` anyway.

src/ResourceTemplate/Segment.php
```diff
@@ -200,6 +204,10 @@ private static function isValidBinding($binding)
      */
     private static function isValidDoubleWildcardBinding($binding)
     {
+        if ($binding === null) {
+            return false;
+        }
```
bshaffer, line 209: in these two cases as well, I am sure why we are accepting `null` as a valid parameter option. It would be better to just typehint `$binding` as a string.

tests/Tests/Unit/BidiStreamTest.php
```diff
@@ -195,6 +194,10 @@ public function testReadFailure()
                 $this->assertSame($response, $responses[$index]);
                 $index++;
             }
+        } catch(ApiException $e) {
+            $this->assertStringContainsString('read failure', $e->getMessage());
+        } catch(Exception $e) {
+            $this->fail('An '.ApiException::class.' should have been thrown');
```
bshaffer, line 200: Instead we should use `expectException` and `expectExceptionMessage` here. Without a catch, the exception gets thrown, so there's no need to add the catch and assertions (if the annotations work, the function call versions will work as well).

tests/Tests/Unit/BidiStreamTest.php
```diff
@@ -290,7 +289,11 @@ public function testWriteFailureWithoutClose()
 
         try {
             $stream->read();
-        } finally {
+        } catch(ApiException $e) {
+            $this->assertStringContainsString('write failure without close', $e->getMessage());
+        } catch (Exception $e) {
+            $this->fail('An '.ApiException::class.' should have been thrown');
+        }finally {
```
bshaffer, line 296: same here. Also, watch the whitespace!

tests/Tests/Unit/ClientStreamTest.php
```diff
@@ -154,6 +151,8 @@ public function testWriteAllFailure()
 
         try {
             $stream->writeAllAndReadResponse($requests);
+        } catch (ApiException $e) {
+            $this->assertStringContainsString('write all failure', $e->getMessage());
```
bshaffer, line 155: Here too... just use `expectException` and `expectException` merthods.

tests/Tests/Unit/ServerStreamTest.php
```diff
@@ -96,6 +94,10 @@ public function testStringsFailure()
                 $this->assertSame($response, $responses[$index]);
                 $index++;
             }
+        } catch(ApiException $e) {
+            $this->assertStringContainsString('strings failure', $e->getMessage());
+        } catch(Exception $e) {
+            $this->fail('An '.ApiException::class.' should have been thrown');
```
bshaffer, line 100: Here too, remove catches in favor of `expectException` methods.

tests/Tests/Unit/ServerStreamTest.php
```diff
@@ -199,6 +198,10 @@ public function testResourcesFailure()
                 $this->assertSame($response, $resources[$index]);
                 $index++;
             }
+        } catch (ApiException $e) {
+            $this->assertStringContainsString('resources failure', $e->getMessage());
+        } catch (Exception $e) {
+            $this->fail('An '.ApiException::class.' should have been thrown');
```
bshaffer, line 204: Here too, remove catches in favor of `expectException` methods.

tests/Tests/Unit/ServerStreamTest.php
```diff
@@ -141,11 +139,16 @@ public function testObjectsFailure()
...
+
         try {
             foreach ($stream->readAll() as $response) {
                 $this->assertEquals($response, $responses[$index]);
                 $index++;
             }
+        } catch (ApiException $e) {
+            $this->assertStringContainsString('objects failure', $e->getMessage());
+        } catch (Exception $e) {
+            $this->fail('An '.ApiException::class.' should have been thrown');
```
bshaffer, line 151: Here too, remove catches in favor of `expectException` methods.

tests/Tests/Unit/TestTrait.php
```diff
@@ -39,6 +39,8 @@
 use Google\Protobuf\Any;
 use Google\Rpc\Code;
 use Google\Rpc\Status;
+use stdClass;
```
bshaffer, line 42: I can't find where this is used, why are you adding it?

bshaffer (changes requested):
Thank you for your work on this! I can see a lot of tedius work went into it. There are, unfortunately, a few issues.

1. You're not including the `Trait` classes needed for the polyfills to work. For instance, if you want the test case to have access to the `assertStringContainsString` polyfill, you need to include the `AssertStringContains` trait (see [the yoast docs](https://github.com/Yoast/PHPUnit-Polyfills#phpunit--750-yoastphpunitpolyfillspolyfillsassertstringcontains) for all the traits you need for which methods).
1. You've extended the `YoastTestCase` base class in many instances where it doesn't _need_ to be extended (it only needs to be extended when the `setUp`/`setUpBeforeClass`/`tearDown`/`tearDownAfterClass` methods are used in the test case)

There is no way you'll be able to get this PR in shape unless you actually run it against an earlier version of PHPUnit. Just require PHPUnit 5 and then run the tests you have here, and you'll start getting the errors you expect to see.

Thanks again for your work on this. I know it's tedius, so we really appreciate it. If you're still interested in making the requested changes, then I'm happy to continue working with you on this PR!

tests/Tests/Unit/BidiStreamTest.php
jeromegamez: Dangit, I was sure I was running into errors when just using the `->expectException*` methods. I'll change them, of course!

tests/Tests/Unit/TestTrait.php
jeromegamez: It was an oversight, I added it to test something and then forgot to remove it. This will be done in my next commit 💪

jeromegamez:
> * You're not including the `Trait` classes needed for the polyfills to work. For instance, if you want the test case to have access to the `assertStringContainsString` polyfill, you need to include the `AssertStringContains` trait (see [the yoast docs](https://github.com/Yoast/PHPUnit-Polyfills#phpunit--750-yoastphpunitpolyfillspolyfillsassertstringcontains) for all the traits you need for which methods).
> 
> * You've extended the `YoastTestCase` base class in many instances where it doesn't _need_ to be extended (it only needs to be extended when the `setUp`/`setUpBeforeClass`/`tearDown`/`tearDownAfterClass` methods are used in the test case)

You're of course right that it's not needed to extend the YoastTestCase, i think I was just lazy in doing so because all traits are already included there 😅. I'll change it!

jeromegamez:
Oh, and thank **you** for the valuable feedback, reading through all of this must have been tedious as well!

Now with the released GH Actions I will be able to verify my changes better - I just spent half an hour teaching my system PHP 5.6, I hope it's reversible 😂

jeromegamez:
Do you see a chance that PHP 5.5 support could be dropped in a minor release?

That would mean that PHP 5.5 users are stuck with the current version, but looking at https://packagist.org/packages/google/gax/php-stats, no PHP 5.5 projects have updated to 1.12 anyway 🤞

bshaffer:
> Do you see a chance that PHP 5.5 support could be dropped in a minor release?

Absolutely... we've had < PHP 7.0 dropped in our auth library for a while now.

jeromegamez:
Alright, I think I got somewhere!

I implemented your change requests (I hope all of them) and tested the whole thing with all PHP versions that I could make available on my local computer (5.6, 7.0, 7.1, 8.0, 8.1). Except the following (known?) problem that I didn't know how to tackle it, everything seems to work.

https://github.com/googleapis/gax-php/blob/745c9ec15282d975cb4f995862a850be78784703/tests/Tests/Unit/RequestBuilderTest.php#L308-L312

Now even without any deprecation warnings like these (straight from past GH action runs)

image

Some warnings are still present (the higher the PHP version, the higher the number of warnings), but it seems to work fine (as always: on my machine™).

I'm looking forward to you approving the workflow runs, I have to say I'm very curious if it really pans out.

jeromegamez:
Oh, I completely forgot to say what I did:

Depending on the grpc version and method signatures, there currently are `TestInterceptor` and `TestInterceptorDeprecated` classes. Before this PR, these classes were all declared in the `GrpcTransportTest`: lower PHP versions warned about them, and higher versions caused this to fail.

So I extended the tests bootstrap - depending on the method signature check, different files are loaded -> no `Deprecated` Suffixes are necessary, and no complaints because classes are extended with mismatching signatures. 🥳

jeromegamez:
Final comment before I stop spamming: I just rebased and split up the commits to make it easier to review one commit at a time instead of everything at once.

lcobucci:
@jeromegamez just reviewed here... great work in making things compatible with ancient PHP and tooling versions 👏

tests/Tests/Unit/MyMessage.php
```diff
@@ -4,10 +4,6 @@
 
 namespace Google\ApiCore\Tests\Unit;
 
-use Google\Protobuf\Internal\GPBType;
-use Google\Protobuf\Internal\RepeatedField;
-use Google\Protobuf\Internal\GPBUtil;
```
bshaffer, line 9: nit, but since this is generated, so we should probably not modify it.

tests/bootstrap.php
```diff
@@ -33,8 +33,21 @@
...
+use Google\ApiCore\Testing\ProtobufMessageComparator;
+use Grpc\Interceptor;
+
+// This is a long line, but better not set some temporary variables that get picked up by something later
+if (
+    (new \ReflectionClass(Interceptor::class))
+        ->getMethod('interceptUnaryUnary')
+        ->getParameters()[3]
+        ->getName() === 'metadata'
+) {
```
bshaffer, line 46: Nice work preserving these!

tests/Tests/Unit/GapicClientTraitTest.php
```diff
@@ -185,6 +183,7 @@ public function testStartOperationsCall()
         $message = new MockRequest();
         $operationsClient = $this->getMockBuilder(OperationsClient::class)
             ->disableOriginalConstructor()
+            ->setMethodsExcept(['validate'])
```
bshaffer, line 186: This line is causing PHP 5.5 tests to fail (the `setMethodsExcept` method doesn't exist in, presumably, PHPUnit 4.x). 

I imagine we could work around this somehow? By mocking `validate` to return null, or something along those lines?

I hate to spin wheels supporting PHP 5.5 though, so another option is to wait and we can drop it soon (within a week).

bshaffer (commented):
Wow, this looks great! You've done a really good job with a super complex task.

PHP 5.5 is failing, I added a comment regarding the failure. All the other comments I've made are very minor.

Thanks a lot again for your hard work on this!

bshaffer:
One more thing - we can add the new tests for PHP 7.4, 8.0 and 8.1 to [the `sync-repo-settings.yaml` file](https://github.com/googleapis/gax-php/blob/main/.github/sync-repo-settings.yaml#L7) so that they're required to merge

tests/Tests/Unit/GapicClientTraitTest.php
jeromegamez: Dropping support for PHP 5.5 would probably be the easiest, but we could also du a follow-up release. It would be nice not having to wait though (also the challenge 😄)

I will try to find PHP 5.5 somewhere so I can test it before pushing the branch just to find out it's still failing. 🤞

bshaffer: It would be great to get a work-around for this so it isn't blocked by dropping PHP 5.5 support! That is, assuming it's an easy fix 🤞

composer.json
```diff
@@ -16,8 +16,9 @@
         "google/common-protos": "^1.0||^2.0"
     },
     "require-dev": {
-        "phpunit/phpunit": "^4.8.36",
-        "squizlabs/php_codesniffer": "3.*"
+        "phpunit/phpunit": "^4.8.36||^5.5.7||^6.5.14||^7.5.20||^8.5.26",
```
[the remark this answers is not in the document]
bshaffer: More info here, because the gaps only exist with PHPUnit 4 (for PHP 5.5), PHPUnit 5 (for PHP 5.6, 7.0, and 7.1) and PHPUnit 8 (PHP 7.2+), we only need to support _those_ versions of `phpunit`

tests/Tests/Unit/GapicClientTraitTest.php
jeromegamez: Okay, I didn't get a hand on PHP 5.5, but I applied your suggestion, and it seems to work (locally). Let's see what GitHub Actions says when I push this. 🤞

jeromegamez:
Okay, third time's a charm! 🙏

jeromegamez:
Sigh. Okay, perhaps we should just wait until support for PHP 5.5 is dropped, that will make it a lot easier :/. I had the "cannot redeclare" errors while refactoring, that's why I added the `setMethodsExcept()` in the first place.

I now remember what the problem was - the `validate()` method is a static one, so it can only be mocked once. My PHP Inspections notified me about this, I should have listened 😄 

image

bshaffer:
@jeromegamez okay, we've merged the changes for dropping PHP 5.5, so if you can revert that last change for `setMethodsExcept`, we should be good to go!

jeromegamez:
@bshaffer Done! ✅

jeromegamez:
Rebased to be up-to-date with the current main branch. (This merge also serves as a gentle ping 😅)

.github/workflows/tests.yml
```diff
@@ -9,7 +9,7 @@ jobs:
         runs-on: ubuntu-latest
         strategy:
             matrix:
-                php: [ 5.6, "7.0", 7.1, 7.2, 7.3 ]
+                php: [ 5.6, 7.0, 7.1, 7.2, 7.3, 7.4, 8.0, 8.1 ]
```
bshaffer, line 12: nit, but the quotes are needed around `#.0` releases, otherwise the names display as the number without the decimal.
```suggestion
                php: [ 5.6, "7.0", 7.1, 7.2, 7.3, 7.4, "8.0", 8.1 ]
```

src/LongRunning/Gapic/OperationsGapicClient.php
```diff
@@ -45,11 +45,9 @@
 namespace Google\ApiCore\LongRunning\Gapic;
 
 use Google\ApiCore\ApiException;
-use Google\ApiCore\Call;
 use Google\ApiCore\GapicClientTrait;
-use Google\ApiCore\RetrySettings;
 use Google\ApiCore\RequestParamsHeaderDescriptor;
-use Google\ApiCore\Transport\TransportInterface;
+use Google\ApiCore\RetrySettings;
```
bshaffer, line 50: This is a generated class, so we _probably_ don't want to make the changes here. But we are removing this soon in favor of it's own package, so it's a pretty innocuous change.

src/Testing/MockRequest.php
```diff
@@ -33,7 +32,7 @@ class MockRequest extends \Google\Protobuf\Internal\Message
      * }
      */
     public function __construct($data = NULL) {
-        \GPBMetadata\ApiCore\Testing\Mocks::initOnce();
+        Mocks::initOnce();
```
bshaffer, line 35: Same here, this is a generated class so it'd be better to just ignore it, as these changes will be overwritten in a future update.

src/Transport/RestTransport.php
```diff
@@ -31,6 +31,7 @@
  */
 namespace Google\ApiCore\Transport;
 
+use BadMethodCallException;
```
bshaffer, line 34: Hmm, this import is not used (the reference to `BadMethodCallException` below is still absolute)

bshaffer (approved):

jeromegamez:
Thanks for merging! 🤗

PR #395, opened 2022-07-01
feat: support requesting numeric enum rest encoding

noahdietz:
Support the top-level REST descriptor config option `numericEnums` which enables inclusion of the `$alt=json;enum-encoding=int` system parameter that requests that enums be encoded as integers instead of strings. This is a Google System parameter. This is an API-wide configuration hence its spot at the same level as `interfaces`

This will be coupled with a generator change to include the REST descriptor config option in the generated output when a generator option is supplied.


noahdietz:
@bshaffer this is ready for review now.

src/RequestBuilder.php
```diff
@@ -107,6 +108,12 @@ public function build($path, Message $message, array $headers = [])
                 // We found a valid uriTemplate - now build and return the Request
 
                 list($body, $queryParams) = $this->constructBodyAndQueryParameters($message, $config);
+                
+                // Request enum fields be encoded as numbers rather than strings.
```
bshaffer, line 112: nit: I'm reading this sentence in a caveman voice

tests/Tests/Unit/RequestBuilderTest.php
```diff
@@ -378,6 +382,20 @@ public function testMethodWithOneOfInQueryString()
         $this->assertSame('some-value', $query['field1']);
     }
 
+    public function testMethodWithNumericEnumsQueryParam()
+    {
+        $message = (new MockRequestBody())
+            ->setName('')
+            ->setNumber(0);
```
bshaffer, line 389: I'm confused by your test example because neither of these seem like enums. Shouldn't we be testing a request or response message with enums?

I'm thinking that it's because this is specifically for the request, and not the response. If that's the case, we probably can drop the `name` and `number` logic, to make this clear.

noahdietz: I think I see your point: The fields on this request message aren't actually part of the test, so let's exclude them. That makes sense to me. Done.

src/RequestBuilder.php
noahdietz: haha oops, fixed.

src/RequestBuilder.php
```diff
@@ -109,7 +109,7 @@ public function build($path, Message $message, array $headers = [])
 
                 list($body, $queryParams) = $this->constructBodyAndQueryParameters($message, $config);
                 
-                // Request enum fields be encoded as numbers rather than strings.
+                // Request enum fields will be encoded as numbers rather than strings.
```
vchudnov-g, line 112: nit: If you're describing what this query param does, the previous wording was correct.
`Request enum fields be encoded as numbers rather than strings (in the response)`

If you're saying that, additionally, the request itself will have enums as numbers, this wording is fine, but it's probably misleading to say that here, since it's not related to the effects of the query param.

vchudnov-g (commented):
LGTM, but I'll PHP experts give the final approval

src/RequestBuilder.php
noahdietz: Added the `(in the response)` clarification.

bshaffer (approved):
Approved but one thing that surprises me about this is nothing is required client side to prepare for the response? Is that because we expect it to be handled automatically by `protobuf`? And would we benefit from any kind of mocking / test case for this scenario?

noahdietz:
> Is that because we expect it to be handled automatically by protobuf?

Yes, protobuf (de)serialization handles numbers or strings for enum fields. I tested it manually before, during the exploration phase of the encoding stuff. Most protobuf JSON serializers handle either string or number for enum fields smoothly.

PR #401, opened 2022-07-26
chore: move longrunning to standalone package

bshaffer:


composer.json
```diff
@@ -13,7 +13,8 @@
         "google/protobuf": "^3.12.2",
         "guzzlehttp/promises": "^1.3",
         "guzzlehttp/psr7": "^1.7.0||^2",
-        "google/common-protos": "^1.0||^2.0"
+        "google/common-protos": "^1.0||^2.0",
+        "google/longrunning": "dev-main"
```
bshaffer, line 17: This will be updated to `^0.1` once we make the first release

composer.json
```diff
@@ -13,7 +13,8 @@
         "google/protobuf": "^3.12.2",
         "guzzlehttp/promises": "^1.3",
         "guzzlehttp/psr7": "^1.7.0||^2",
-        "google/common-protos": "^1.0||^2.0"
+        "google/common-protos": "^1.0||^2.0",
```
bshaffer, line 16: While it's not absolutely required, it would be best to allow `^3.0` of `google/common-protos` here as well, as that version will not have any of the operations protos in it.

See https://github.com/googleapis/common-protos-php/pull/41

dwsupplee (approved):

dwsupplee:
It looks like there is a LongRunning test as well, should we remove that too?

bshaffer:
@dwsupplee good catch! It's been moved [into the php-longrunning package](https://github.com/googleapis/google-cloud-php/blob/main/LongRunning/tests/Unit/OperationsClientTest.php), so you're right it should have been removed, and I've now removed it.  Thank you!

PR #403, opened 2022-08-02
chore: add typehints

bshaffer:
- Adds compatible typehints where possoble
- Improves PHPDoc where it isn't possible
- Adds primitive typehints (PHP 7+)

Depends on #397


.github/workflows/tests.yml
```diff
@@ -9,14 +9,14 @@ jobs:
         runs-on: ubuntu-latest
         strategy:
             matrix:
-                php: [ 5.6, "7.0", 7.1, 7.2, 7.3, 7.4, "8.0", 8.1 ]
+                php: [ "7.0", 7.1, 7.2, 7.3, 7.4, "8.0", 8.1 ]
```
noahdietz, line 12: Nit: Any idea why some are wrapped in quotations and others aren't?

src/RequestBuilder.php
```diff
@@ -108,12 +108,12 @@ public function build($path, Message $message, array $headers = [])
                 // We found a valid uriTemplate - now build and return the Request
 
                 list($body, $queryParams) = $this->constructBodyAndQueryParameters($message, $config);
-                
+
                 // Request enum fields will be encoded as numbers rather than strings  (in the response).
                 if ($numericEnums) {
                     $queryParams['$alt'] = "json;enum-encoding=int";
                 }
-
```
noahdietz, line 116: Whitespace be gone!!

src/ResourceTemplate/AbsoluteResourceTemplate.php
```diff
@@ -61,10 +61,10 @@ class AbsoluteResourceTemplate implements ResourceTemplateInterface
 
     /**
      * AbsoluteResourceTemplate constructor.
-     * @param string $path
+     * @param string|null $path
```
noahdietz, line 64: So why `string|null` here, but not elsewhere? There are a few other spots that have a `string` param that is optional and defaults to `null`.

src/ResourceTemplate/RelativeResourceTemplate.php
```diff
@@ -272,12 +272,15 @@ public function match($path)
         return $collapsedBindings;
     }
 
-    private function matchException($path, $reason)
+    /**
+     * @param string|null $path
```
noahdietz, line 276: Missing one param here? `$reason`?

src/ResourceTemplate/RelativeResourceTemplate.php
```diff
@@ -272,12 +272,15 @@ public function match($path)
...
+     */
+    private function matchException($path, string $reason)
```
noahdietz, line 278: Can we add `string $path = null` here?

src/ResourceTemplate/RelativeResourceTemplate.php
```diff
@@ -272,12 +272,15 @@ public function match($path)
...
     {
         return new ValidationException("Could not match path '$path' to template '$this': $reason");
     }
 
-    private function renderingException($bindings, $reason)
+    private function renderingException(array $bindings, string $reason)
```
noahdietz, line 283: Nit: php doc?

.github/workflows/tests.yml
bshaffer: Yes, it's so the "7.0" isn't converted to "7" when converted to a string. We could wrap them all in quotes, but it's a little more concise this way.

noahdietz: That's fine as is, thanks for clarifying.

src/ResourceTemplate/RelativeResourceTemplate.php
bshaffer: We can't because declaring an optional parameter before a required parameter throws a warning in PHP 8.0

src/ResourceTemplate/AbsoluteResourceTemplate.php
bshaffer: I think you're right, and the optional `null` should be removed.

src/ResourceTemplate/RelativeResourceTemplate.php
bshaffer: I only added it because I couldn't typehint it. But since this is a private method, I've gone back and fixed it so the typehint is valid.

src/ResourceTemplate/RelativeResourceTemplate.php
bshaffer: I was able to remove support for `|null`, since this method is private.

noahdietz: Cool thanks

noahdietz (approved):

src/ResourceTemplate/RelativeResourceTemplate.php
bshaffer: My general rule with this PR was to only add `phpdoc` if I couldn't typehint the parameters properly

noahdietz: Yeah that's fair.

PR #415, opened 2022-08-22
chore: add phpspec prophecy to require-dev

bshaffer:
The latest versions of `phpunit` require `phpspec/prophecy` as a separate package


noahdietz (approved):
:shipit:

PR #418, opened 2022-08-23
feat: add startAsyncCall support

noahdietz:
Adds `startAsyncCall` to `GapicClientTrait` that initiates an API call via `startApiCall`, but directly returns the resulting `Promise` for the user to handle. Does not support Streaming RPCs today. To support async initialization of a PagedResponse, I added a `getPagedListResponseAsync` variant with all of the logic needed today, which is then called from the synchronous variant and waited upon. 

Also moves all of the "call config" validation into a helper used by both `startApiCall` and `startAsyncCall`.


src/GapicClientTrait.php
```diff
@@ -528,6 +528,86 @@ private static function defaultTransport()
...
+     *           override for the call.
+     * }
+     *
+     * @experimental
+     *
+     * @return PromiseInterface
+     */
+    private function startAsyncCall(
+        string $methodName,
+        string $interfaceName = null,
```
bshaffer, line 594: I envisioned the sendAsync method to only accept a Request object. Would we be able to include that logic in this method as well?

noahdietz: I'm not sure what you mean. How would we know which RPC to look up if we didn't have the methodName or optional interfaceName?

It only accepts the $request message as far as payload is concerned.

bshaffer: There's (almost always) a one-to-one map of Request type to RPC (there's one exception, in BigQuery, where `Table` is used for multiple RPCs). I thought we decided to derive it and allow the users to just pass the Request message, but maybe we decided to require the `methodName` also.

I am still not really sure what the `$interfaceName` is for, however. Is that something sent in by the GapicClients automatically? I don't think that should be exposed in `sendAsync`.

noahdietz: > (almost always)

For this reason, I don't think it would be good to use such "magic" tbh. I see it as kind of brittle.

You are right that we don't need interfaceName! removing it.

noahdietz: In our design doc, I included `methodName` and did not include `interfaceName`, fwiw.

noahdietz: I'm wondering now which `methodName` the user should be expected to provide to the GAPIC - `GetFoo` (rpc name) or `getFoo` (php client method name). I'd say the latter, but then we have UpperCamelCase (the first letter) to turn it into the RPC name to look up with in the descriptor. WDYT?

noahdietz: I decided to convert to the UpperCamelCase form to start, but let's discuss more.

noahdietz: We discussed this offline and `string methodName` is fine, we will force UpperCamel.

dwsupplee: We might need interfaceName for cases like the following RPC:

https://github.com/googleapis/google-cloud-php/blob/main/PubSub/src/V1/Gapic/PublisherGapicClient.php#L1053 - I believe we need it order to lookup the correct key here: https://github.com/googleapis/google-cloud-php/blob/main/PubSub/src/V1/resources/publisher_rest_client_config.php#L5

bshaffer: Oof, could we just make it so the client looks through every possible interface name in the descriptors instead?

noahdietz: hmm...idk...I think an individual GAPIC client only gets the descriptors for its service: https://github.com/googleapis/gax-php/blob/main/src/GapicClientTrait.php#L399-L400

Although idk how it would work today if the client didn't know to go an look in the "other service's" rest config.

dwsupplee: @noahdietz Is it correct that if we had mixins for LRO/IAM, etc., we'd only need a single interface for each client?

If so, what are your thoughts on us emulating what this would look like in the generator, so that when we eventually do migrate to mixins it wouldn't even necessitate a change in the generated output.

noahdietz: > Is it correct that if we had mixins for LRO/IAM, etc., we'd only need a single interface for each client?

No, because the mix-in services still have HTTP rules identified by their actual service name, and the generator respects that. 

Example [method](https://github.com/googleapis/google-cloud-php/blob/main/Datastream/src/V1/Gapic/DatastreamGapicClient.php#L2622-L2644) and [config](https://github.com/googleapis/google-cloud-php/blob/main/Datastream/src/V1/resources/datastream_rest_client_config.php#L306-L328).

So the descriptors **must** be available to GAX to look up...

noahdietz: I think we _will_ need to do what @bshaffer suggests in order to support mixins in a reasonable way...today, we just expose a mixin method on the client e.g. `getLocation` from `google.cloud.Locations`, we don't ask the user to provide the interface for that (the GAPIC is generated to know this). So I think we will need to have GAX `startAsyncCall` look at all other interfaces if the methodName is not present in the current interface.

noahdietz: Ok spoke with @dwsupplee offline and we will do the interface lookup in GAX if the requested method doesn't exist on the client's service descriptor. This _should_ be possible but I might need to make a few changes.

noahdietz: In looking at this some more, it looks like a service descriptor will get the mix-in rpc descriptors in its own descriptor config interface entry ([example](https://github.com/googleapis/google-cloud-php/blob/main/Datastream/src/V1/resources/datastream_descriptor_config.php#L166-L173)). This is good! But we will still need to lookup the interface name in order to include that for gRPC request building and HTTP binding lookup.

noahdietz: Ok fixed all of our troubles by using a new `interfaceOverride` in the descriptor config!

src/GapicClientTrait.php
```diff
@@ -528,6 +528,98 @@ private static function defaultTransport()
...
+        string $methodName,
+        Message $request,
+        array $optionalArgs = []
+    ) {
+        // Convert method name to the UpperCamelCase of RPC names from lowerCamelCase of GAPIC method names
+        // in order to find the method in the descriptor config.
+        $methodName = ucfirst($methodName);
+        $this->validateCallConfig($methodName);
+        
+        $method = $this->descriptors[$methodName];
```
bshaffer, line 601: Having the `validateCallConfig` method return the method descriptors would clean this up!

Also, nit, but make variable name `$methodDescriptors` for clarity?
```suggestion
        $methodDescriptors = $this->validateCallConfig($methodName);
```

noahdietz: For sure, sounds good! I will implement those changes.

src/GapicClientTrait.php
```diff
@@ -528,6 +528,98 @@ private static function defaultTransport()
...
+        }
+        
+        // LRO are either Standard LRO response type or custom, which are handled by
+        // startOperationCall, so no need to validate responseType for those callType.
+        if ($callType != Call::LONGRUNNING_CALL) {
+            if (!isset($method['responseType'])) {
+                throw new ValidationException("Requested method '$methodName' does not have a responseType " .
+                    "in descriptor configuration.");
+            }
+        }
```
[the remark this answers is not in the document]
noahdietz: Good call. Done!

src/GapicClientTrait.php
noahdietz: Done.

src/GapicClientTrait.php
```diff
@@ -528,6 +528,98 @@ private static function defaultTransport()
             : 'rest';
     }
 
+    private function validateCallConfig(string $methodName)
```
dwsupplee, line 531: Outside of checking for the method name, what scenarios do we expect to need to run this validation? Since we're generating the descriptor configs ourselves and they aren't user input, I'm just thinking through if we need to do this. It seems like we'd have bigger problems if these fields don't exist.

dwsupplee (approved):
LGTM, just a thought on the validation.

src/GapicClientTrait.php
noahdietz: For me, it is more about providing a better experience for those error scenarios than it is about preventing them. I'm not sure if it is entirely possible, but if an older version of a client uses a newer version of gax, and a manual layer there attempts to make use of these newer gax APIs, I think this exception is better than a null pointer or segfault.

dwsupplee: The vector seems unlikely, but the cost of doing these checks is cheap so if you prefer them I'm okay with it.

dwsupplee (approved):

bshaffer (approved):

PR #426, opened 2022-10-24
chore(docs): add docs gen for protobuf-php

bshaffer:
**UPDATE**: These are now published! https://protobuf.dev/reference/php/

Fixes https://github.com/protocolbuffers/protobuf/issues/10373

Adds a script to build docs for PHP Protobuf. We will have to run this locally and submit CLs manually to google3 when an update is desired.

The Protobuf team does not maintain the reference documentation for PHP at this time and rely on us (the Core Client Libraries team) to do so.

As per the instructions at the top of `dev/sh/build-protobuf-docs.sh`, the output should be copied over to the [protocolbuffers.github.io repo](https://github.com/protocolbuffers/protocolbuffers.github.io) as seen in https://github.com/protocolbuffers/protocolbuffers.github.io/pull/9


bshaffer:
This has been launched! https://protobuf.dev/reference/php/

PR #428, opened 2022-11-09
feat: add ResourceHelperTrait

noahdietz:
Introduces a new `ResourceHelperTrait` that supplies functionality for loading a resource name template map from a descriptor config, retrieving a `PathTemplate`, and parsing values using registered templates. This can be used by GAPIC clients to replace much of the resource template properties generated into every client.

Proposal is at http://go/php-resource-name-refactor.


src/ResourceHelperTrait.php
```diff
@@ -0,0 +1,92 @@
...
+{
+    private static $templateMap;
+
+    // Must be implemented by extendees to call loadTemplates.
+    private static function registerTemplates()
+    {
+        self::$templateMap = [];
+    }
+
+    private static function loadTemplates(string $configPath, string $serviceName)
```
bshaffer, line 47: let's add return typehints for everything!
```suggestion
    private static function loadTemplates(string $configPath, string $serviceName): void
```

src/ResourceHelperTrait.php
```diff
@@ -0,0 +1,92 @@
...
+namespace Google\ApiCore;
+
+use Google\ApiCore\ValidationException;
+
+trait ResourceHelperTrait
+{
+    private static $templateMap;
+
+    // Must be implemented by extendees to call loadTemplates.
+    private static function registerTemplates()
```
bshaffer, line 42: Is the point of this method to allow the GapicClients to lazy load the templates? Currently, `GapicClientTrait` loads teh descriptors in `setClientOptions`, which is called from the `__constructor`, so we could just load all the templates up front and then we wouldn't need this method.

bshaffer (commented):
preliminary review, looks good!

src/ResourceHelperTrait.php
noahdietz: Ok! Will do :)

bshaffer: [Nullable typehints](https://www.php.net/manual/en/migration71.new-features.php) were added in PHP 7.1, so this will definitely be supported in the 2.x branch

src/ResourceHelperTrait.php
bshaffer: Then, we could potentially be smarter and make it use values in `$options`, so we don't have to define `registerTemplates` on every GAPIC client:

```php
private static function registerTemplates()
{
        if (!is_null(self::$templateMap)) {
            return;
        }
        if (empty($this->options['descriptorsConfigPath']) || empty($this->options['serviceName'])) {
            return;
        }
        $descriptors = require($this->options['descriptorsConfigPath']);
        $templates = $descriptors['interfaces'][$this->options['serviceName']]['templateMap'] ?? [];
        self::$templateMap = [];
        foreach ($templates as $name => $template) {
             self::$templateMap[$name] = new PathTemplate($template);
        }
}
```

noahdietz: Sorry, I'm confused, we can't use `$this` in a `static` method, can we? There is no instance associated with a `static` method.

src/ResourceHelperTrait.php
noahdietz: Ok, then I will add a TODO to add that for `getTemplate`.

noahdietz: Actually refactored `loadTemplates` to not have a `return`, so let's see if that makes php 7 happy.

noahdietz: Nope, need to remove it

src/ResourceHelperTrait.php
bshaffer: You're right... we can't. I missed that it was static. The way we have it now, `descriptorsConfigPath` and `serviceName` are both configurable in the constructor. So if those values were configured manually, the paths defined `registerTemplates` would differ.

src/ResourceHelperTrait.php
bshaffer: We can add it after we drop support for PHP 7.0 and below, but hopefully that'll happen before we roll this out anyway.

src/ResourceHelperTrait.php
bshaffer: e.g.

```php
$foo = new FooClient(['descriptorsConfigPath' => '/path/to/my_descriptors.php', 'serviceName' => 'myServiceName']);
$templateName = $foo->templateName(); // This will still use the default descriptors and service name
```

bshaffer: The way we currently have it isn't dynamic either, so we could still implement it this way, and it wouldn't be breaking. It's just a bit strange since the API makes it seem configurable and it wouldn't be.

noahdietz: Yeah this is a fair point, but fwiw that's not a use case that we even support today for the existing resource helper support. It could be a cool thing to support in the future though!

I don't see why `serviceName` would ever change for a given GAPIC - it is generated based on a specific service. But I could definitely see a world where some veneer wants to extend the GAPIC and add more resource name mappings that aren't in the protos for whatever reason.

noahdietz: I could even see it being used as a backwards compatibility shim if a pattern or type changes. We could encode the broken pattern/mapping in the veneer layer and extend the descriptor config.

noahdietz: I think I liked it mostly because it opens us up to support a more config driven approach vs. hardcoded mapping like before. @dwsupplee gets credit for the idea.

noahdietz: Resolving for now.

dwsupplee:
I just set aside a block of time for tomorrow morning to review.

src/ResourceHelperTrait.php
```diff
@@ -0,0 +1,95 @@
...
+    private static function getTemplate(string $key)
+    {
+        // TODO: Add nullable return type reference once PHP 7.1 is minimum.
+        if (is_null(self::$templateMap)) {
+            self::registerTemplates();
+        }
+        return self::$templateMap[$key] ?? null;
+    }
+
+    private static function parse(string $formattedName, string $template = null): array
```
bshaffer, line 72: I checked, and `parse` does not exist on any client classes 👍

src/ResourceHelperTrait.php
```diff
@@ -0,0 +1,95 @@
...
+
+        $descriptors = require($configPath);
+        $templates = $descriptors['interfaces'][$serviceName]['templateMap'] ?? [];
+        self::$templateMap = [];
+        foreach ($templates as $name => $template) {
+            self::$templateMap[$name] = new PathTemplate($template);
+        }
+    }
+
+    private static function getTemplate(string $key)
```
bshaffer, line 63: The function name `getTemplate` exists on one gapic client:

 - [`Dataflow/src/V1beta3/Gapic/TemplatesServiceGapicClient.php`](https://github.com/googleapis/google-cloud-php/blob/main/Dataflow/src/V1beta3/Gapic/TemplatesServiceGapicClient.php#L321)

Note: this will not be a conflict once we use the `send` prefix, but we should think of a more explicit name in the meantime, e.g. `getPathTemplate` (I've confirmed this does not exist on any clients)

Note: This also means we would want to rename `loadTemplates` and `registerTemplates` to be consistent with the naming that we choose for `getTemplate`

bshaffer (changes requested):

tests/Tests/Unit/ResourceHelperTraitTest.php
```diff
@@ -0,0 +1,115 @@
...
+        ResourceHelperTraitStub::parseName('no/matching/pattern');
+    }
+    
+}
+
+class ResourceHelperTraitStub{
+    use ResourceHelperTrait;
+
+    public static $CONFIG_PATH = __DIR__ . '/testdata/test_service_descriptor_config.php';
+    public static $SERVICE_NAME = 'test.interface.v1.api';
```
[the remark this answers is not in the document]
bshaffer: I'm going to guess this was done in case we wanted to customize it to point to a different file or service name in another test. I'd rather it actually mimic the GAPICs though! (which means, using constants but not private ones...)

src/ResourceHelperTrait.php
noahdietz: w00t thank you

src/ResourceHelperTrait.php
noahdietz: Great call, I'll do that. Done.

tests/Tests/Unit/ResourceHelperTraitTest.php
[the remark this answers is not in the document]
noahdietz: For sure, sounds good. Done.

tests/Tests/Unit/ResourceHelperTraitTest.php
```diff
@@ -0,0 +1,115 @@
...
+    {
+        $this->expectException(ValidationException::class);
+        $this->expectExceptionMessage('Input did not match any known format. Input: no/matching/pattern');
+
+        ResourceHelperTraitStub::parseName('no/matching/pattern');
+    }
+    
+}
+
+class ResourceHelperTraitStub{
```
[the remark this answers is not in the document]
noahdietz: haha a nit I will happily accept. Done.

dwsupplee:
Sorry about the delay, I was out the last few days. Taking a look today.

src/ResourceHelperTrait.php
```diff
@@ -0,0 +1,95 @@
...
+namespace Google\ApiCore;
+
+use Google\ApiCore\ValidationException;
+
+trait ResourceHelperTrait
+{
+    private static $templateMap;
+
+    // Must be implemented by extendees to call loadPathTemplates.
+    private static function registerPathTemplates()
```
dwsupplee, line 42: The safety of having this marked abstract made a lot of sense. What are your thoughts on changing visibility to protected/public in order to achieve that? We can make sure we add annotations to the docblock about how it is internal/access should be considered private.

src/ResourceHelperTrait.php
```diff
@@ -0,0 +1,95 @@
...
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+namespace Google\ApiCore;
+
+use Google\ApiCore\ValidationException;
+
+trait ResourceHelperTrait
+{
+    private static $templateMap;
```
dwsupplee, line 39:
```suggestion
    /** @var array|null */
    private static $templateMap;
```

src/ResourceHelperTrait.php
```diff
@@ -0,0 +1,95 @@
...
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+namespace Google\ApiCore;
+
+use Google\ApiCore\ValidationException;
+
+trait ResourceHelperTrait
```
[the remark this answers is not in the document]
noahdietz: Yes, great call, thank you. Added!

src/ResourceHelperTrait.php
noahdietz: Sweet, thank you. Accepted.

src/ResourceHelperTrait.php
noahdietz: > The safety of having this marked abstract made a lot of sense

Yeah, I liked a lot as well. It just _made sense_ to me as `abstract`.

> What are your thoughts on changing visibility to protected/public in order to achieve that? We can make sure we add annotations to the docblock about how it is internal/access should be considered private.

My only worry is that it really isn't something we want users to see on the surface. It's an implementation detail. The only `public` method on `GapicClientTrait` is `close` which makes a lot of sense. The `protected` functions have `@access private` phpdoc statements, but IDK why they need to be `protected` other than for maybe veneer testing purposes? The other thing was that since this `trait` should be used by generated surfaces only(?), we will always implement the `registerPathTemplates` function (he says with 100% confidence that nothing will ever go wrong with that assumption 😝).

If you & @bshaffer think it's nothing to worry about, I can switch it back to `protected abstract`.

dwsupplee (approved):

bshaffer (approved):

PR #429, opened 2022-11-17
chore(docs): remove serviceAddress param from phpdoc

bshaffer:
Remove deprecated `$serviceAddress` param from `GapicClientTrait` phpdoc.

See https://github.com/googleapis/gapic-generator-php/pull/531

Note: the functionality of this parameter does not change, but we are removing it to be consistent in `GapicClientTrait` with our GAPIC clients, which will have the PHPDoc for this parameter removed as part of the above change in the generator.


noahdietz (approved):
Can you change the commit to say "remove serviceAddress param from doc"? haha or "chore(docs): remove serviceAddress param"

bshaffer:
> Can you change the commit to say "remove serviceAddress param from doc"? haha or "chore(docs): remove serviceAddress param"

How about both? Good catch!

PR #430, opened 2022-11-30
fix: message parameters in querystring

bshaffer:


src/RequestBuilder.php
```diff
@@ -201,7 +201,19 @@ private function constructBodyAndQueryParameters(Message $message, array $config
                 $requiredQueryParam = Serializer::toCamelCase($requiredQueryParam);
                 if (!array_key_exists($requiredQueryParam, $queryParams)) {
                     $getter = Serializer::getGetter($requiredQueryParam);
-                    $queryParams[$requiredQueryParam] = $message->$getter();
+                    $queryParamValue = $message->$getter();
+                    if ($queryParamValue instanceof Message) {
+                        $param = json_decode($queryParamValue->serializeToJsonString(), true);
+                        if (is_array($param)) {
+                            foreach ($param as $key => $value) {
+                                $queryParams[$requiredQueryParam . '.' . $key] = $value;
```
noahdietz, line 209: Can you add a comment that indicates this only supports one-level of nesting? i.e. it will not support enumerating each field of a message type field within this query param value (that is an instance of a Message).

src/RequestBuilder.php
```diff
@@ -201,7 +201,19 @@ private function constructBodyAndQueryParameters(Message $message, array $config
...
+                            }
+                        } else {
+                            $queryParams[$requiredQueryParam] = $param;
+                        }
+                    } else {
+                        $queryParams[$requiredQueryParam] = $queryParamValue;
+                    }
```
noahdietz, line 216: nit: I really want to optimize this code some how because they are the basically the same statement...but I don't have any clever ideas...do you?

src/RequestBuilder.php
bshaffer: Added!

src/RequestBuilder.php
bshaffer: We could use `continue`:

```php
            foreach ($config['queryParams'] as $requiredQueryParam) {
                $requiredQueryParam = Serializer::toCamelCase($requiredQueryParam);
                if (!array_key_exists($requiredQueryParam, $queryParams)) {
                    $getter = Serializer::getGetter($requiredQueryParam);
                    $queryParamValue = $message->$getter();
                    if ($queryParamValue instanceof Message) {
                        // Decode message for the query parameter.
                        $queryParamValue = json_decode($queryParamValue->serializeToJsonString(), true);
                        if (is_array($queryParamValue)) {
                            // If the message has properties, add them as nested querystring values.
                            // NOTE: If the nested property is also an object, it will be a JSON
                            // object. This means the nesting only happens at one level of depth.
                            foreach ($queryParamValue as $key => $value) {
                                $queryParams[$requiredQueryParam . '.' . $key] = $value;
                            }
                            continue;
                        }
                    }
                    $queryParams[$requiredQueryParam] = $queryParamValue;
                }
            }
```

noahdietz: oo yeah that great :)

bshaffer: Or split it up like this, which would allow for arrays to be passed in `$config['queryParams']`:
```php
            foreach ($config['queryParams'] as $requiredQueryParam) {
                $requiredQueryParam = Serializer::toCamelCase($requiredQueryParam);
                if (!array_key_exists($requiredQueryParam, $queryParams)) {
                    $getter = Serializer::getGetter($requiredQueryParam);
                    $queryParamValue = $message->$getter();
                    if ($queryParamValue instanceof Message) {
                        // Decode message for the query parameter.
                        $queryParamValue = json_decode($queryParamValue->serializeToJsonString(), true);
                    }
                    if (is_array($queryParamValue)) {
                        // If the message has properties, add them as nested querystring values.
                        // NOTE: If the nested property is also an object, it will be a JSON
                        // object. This means the nesting only happens at one level of depth.
                        foreach ($queryParamValue as $key => $value) {
                            $queryParams[$requiredQueryParam . '.' . $key] = $value;
                        }
                    } else {
                        $queryParams[$requiredQueryParam] = $queryParamValue;
                    }
                }
            }
```

noahdietz (approved):

PR #439, opened 2023-01-06
fix: ensure metadata return type is loaded into descriptor pool

dwsupplee:
There are a few rare cases where metadata return types for long running operations aren't being added into the descriptor pool when utilizing the REST transport. This fix ensures they are added by instantiating the metadata class, which triggers an `initOnce`call which loads the data into the associated pool.

### DIFF
--- a/tests/Tests/Unit/Transport/RestTransportTest.php
+++ b/tests/Tests/Unit/Transport/RestTransportTest.php
@@ -146,6 +146,9 @@ class RestTransportTest extends TestCase
             ->wait();
     }
 
+    /**
+     * @runInSeparateProcess
+     */
     public function testStartUnaryCallWithValidProtoNotLoadedInDescPool()
     {
         $endpoint = 'www.example.com';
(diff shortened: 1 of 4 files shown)


tests/Tests/Unit/Transport/RestTransportTest.php
```diff
@@ -144,6 +146,91 @@ public function testStartUnaryCallThrowsException()
...
+                'metadataReturnType' => DateTime::class
+            ])
+            ->wait();
+
+        $this->assertInstanceOf(Operation::class, $response);
+        $this->assertEquals(
+            $body['metadata']['@type'],
+            $response->getMetadata()->getTypeUrl()
+        );
+    }
```
bshaffer, line 193: It's unlikely, but I'm a little worried that this type could be accidentally loaded earlier. Could we mitigate this by adding a `@runInSeparateProcess` PHPUnit annotation?

bshaffer (commented):
One small request, otherwise this looks great!

tests/Tests/Unit/Transport/RestTransportTest.php
dwsupplee: Good call, added!

bshaffer (approved):

PR #441, opened 2023-01-24
fix: ensure cache is used in calls to ADC::onGCE

bshaffer:
Will require an update to the minimum `google/auth` version once https://github.com/googleapis/google-auth-library-php/pull/431 is merged

 **Note**: We will also need to cherry-pick this commit into [v1.19.0](https://github.com/googleapis/google-auth-library-php/releases/tag/v1.19.0) and tag `v1.19.1` as well, in order to maintain compatibility in PHP 7.0.
 - [x] tagged [v1.25.0](https://github.com/googleapis/google-auth-library-php/releases/tag/v1.25.0)
 - [x] tagged [v1.19.1](https://github.com/googleapis/google-auth-library-php/releases/tag/v1.19.1)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -7,7 +7,7 @@
     "license": "BSD-3-Clause",
     "require": {
         "php": ">=7.0",
-        "google/auth": "^1.18.0",
+        "google/auth": "dev-add-get-fetcher as 1.25.0",
         "google/grpc-gcp": "^0.2",
         "grpc/grpc": "^1.13",
         "google/protobuf": "^3.21.4",
--- a/composer.json
+++ b/composer.json
@@ -7,7 +7,7 @@
     "license": "BSD-3-Clause",
     "require": {
         "php": ">=7.0",
-        "google/auth": "dev-add-get-fetcher as 1.25.0",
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -7,7 +7,7 @@
     "license": "BSD-3-Clause",
     "require": {
         "php": ">=7.0",
-        "google/auth": "^1.18.0",
+        "google/auth": "dev-add-get-fetcher as 1.25.0",
```
bshaffer, line 10: once `google/auth` is tagged, this will need to be updated to 
```suggestion
        "google/auth": "1.19.1||^1.25.0",
```

composer.json
```diff
@@ -7,7 +7,7 @@
     "license": "BSD-3-Clause",
     "require": {
         "php": ">=7.0",
-        "google/auth": "^1.18.0",
+        "google/auth": "1.19.1||^1.25.0",
```
bshaffer, line 10: in `google/auth:1.20.0`, we dropped support for PHP 7.0, which is still supported by this library. For this reason, we had to configure the dependency like this.

noahdietz (approved):

PR #449, opened 2023-03-22
chore(tests): skip v2 client tests on php 7.3 and below

bshaffer:
PHPUnit will skip running for any test in a subnamespace `\Client\` when on PHP 7.3 or below (see `tests/Tests/Unit/Client/PretendGeneratedClientTest.php`)

Any test extending `GeneratorTest` will runs as expected even when the `phpunit-polyfill` library isn't installed (as this dependency is not installed for all our tests).


bshaffer:
We decided against this approach in favor of waiting for the drop of PHP 7.3

PR #450, opened 2023-03-31
feat: typesafety for client options (new surface only)

bshaffer:
Add classes to wrap our current typeless array for Client Options. There is a different "Options" class for each "layer" of options (e.g. nested array), so that we have typehinted classes with corresponding docs instead of untyped associative arrays. This will allow us to have consistent options across our various classes, including handwritten clients.

An overview of the design decisions

1. I implemented `ArrayAccess` in `OptionsTrait` so that the getters do not show up in the docs or IDEs. This is just a nice convenience factor, since the getters shouldn't be used or be considered very important. It also causes less churn for our internal code, since we were previously using array access.
1. I made all the setters public so they _do_ show up in IDEs and docs. These could be made private, as the most common API will be to just call the constructor with an array. But I thought it seemed nicer for the documentation and API.
1. I didn't add typehints for the Options classes public methods (e.g., customers could only use these classes in their code if they called `toArray()` before passing them anywhere). This is because I didn't want to widen (e.g. remove) the `array` typehints (as union types aren't available until PHP 8.0! 😡). If we decide to do it this way (we can always widen later), we should at least add a reference (e.g. `@see ClientOptions`), etc, to everywhere it's applicable, so users can find the docs.
1. I left all strings nullable. We could default to empty strings, but I feel like `null` says something different. I did not allow arrays/booleans to be nullable, because I don't think that distinction means anything (and it just makes the code look sloppier). 

cc @saranshdhingra and @noahdietz - This is a draft! I am not hard-set on any of these design decisions! I would still love your feedback!

 - [x] Only use in V2 client surfaces
 - [x] Add `CallOptions`
     - See [misc-config-options](https://github.com/grpc/grpc/tree/master/src/php#misc-config-options) and [group-grpc-args](https://grpc.github.io/grpc/core/group__grpc__arg__keys.html)
 - [x] Add tests
 - [x] Add better docs explaining its usage
 - [x] Get design approval 
 - [x] drop support for PHP 7.4


src/Options/OptionsTrait.php
```diff
@@ -0,0 +1,53 @@
...
+    public function offsetUnset($offset): void
+    {
+        throw new BadMethodCallException('Cannot unset options through array access. Use the setters instead');
+    }
+
+    public function toArray(): array
+    {
+        $arr = [];
+        /** @phpstan-ignore-next-line */
+        foreach ($this as $key => $value) {
```
bshaffer, line 48: Funny enough, using `(array) $this` does not have the same behavior, and the values of `$key` are prefixed with the classname, e.g `Classnamefoo` instead of `foo`.

src/Options/OptionsTrait.php
```diff
@@ -0,0 +1,53 @@
+<?php
+
+namespace Google\ApiCore\Options;
+
+use Google\ApiCore\ValidationException;
+use BadMethodCallException;
+
+trait OptionsTrait
+{
```
bshaffer, line 9: nit: maybe we move this to `Google\ApiCore\OptionsTrait`? That way the `Options` dir is only for client options.

src/GapicClientTrait.php
```diff
@@ -362,37 +364,29 @@ private function setClientOptions(array $options)
             'libName',
             'libVersion',
         ]);
-
-        $clientConfig = $options['clientConfig'];
-        if (is_string($clientConfig)) {
-            $clientConfig = json_decode(file_get_contents($clientConfig), true);
-        }
+        $options = new ClientOptions($options);
```
saranshdhingra, line 367: Probably we could do:
```php
$options = ($options instanceof ClientOptions ? $options : new ClientOptions($options);
```
This would allow the users to pass the object as is when initiating a new Client.

src/Options/OptionsTrait.php
saranshdhingra: Just a thought but I think this dir seems the right place, as it's usage is in this directory _only_.

Perhaps, `Google\ApiCore\Options\Traits` can work too.

src/GapicClientTrait.php
bshaffer: The reason I don't want to do this is that it would require us to lose the `array` typehint for all existing clients.

We could do this for V2 clients only, or we can wait until we drop support for PHP 7.4 and add union types (e.g. `array|ClientOptions`)

src/GapicClientTrait.php
```diff
@@ -785,6 +797,19 @@ private function configureCallConstructionOptions(string $methodName, array $opt
         ];
     }
 
+    /**
+     * @return array
+     */
+    private function configureCallOptions(array $optionalArgs): array
+    {
+        if ($this->isNewClientSurface()) {
+            // cast to ClientOptions for new surfaces only
```
noahdietz, line 806:
```suggestion
            // cast to CallOptions for new surfaces only
```

src/GapicClientTrait.php
```diff
@@ -785,6 +797,19 @@ private function configureCallConstructionOptions(string $methodName, array $opt
         ];
     }
 
+    /**
+     * @return array
+     */
+    private function configureCallOptions(array $optionalArgs): array
+    {
+        if ($this->isNewClientSurface()) {
```
noahdietz, line 805: Is this worth caching this result in a property on init if it will be invoked on every API call?

bshaffer: for sure

bshaffer: done

noahdietz (approved):

bshaffer:
Added `Do not merge` as I'd like to wait one week after we release #470 in order to mitigate churn

PR #453, opened 2023-04-11
chore: add spread operator, refactor tests

bshaffer:


tests/Tests/Unit/ArrayTraitTest.php
```diff
@@ -50,7 +50,7 @@ public function testPluck()
         $value = '123';
         $key = 'key';
         $array = [$key => $value];
-        $actualValue = $this->implementation->call('pluck', [$key, &$array]);
+        $actualValue = $this->implementation->pluck($key, $array);
```
vishwarajanand, line 53:
```suggestion
        $actualValue = $this->implementation->pluck($key, &$array);
```

tests/Tests/Unit/ArrayTraitTest.php
```diff
@@ -60,7 +60,7 @@ public function testPluckThrowsExceptionWithInvalidKey()
     {
         $array = [];
         $this->expectException(InvalidArgumentException::class);
-        $this->implementation->call('pluck', ['not_here', &$array]);
+        $this->implementation->pluck('not_here', $array);
```
vishwarajanand, line 63:
```suggestion
        $this->implementation->pluck('not_here', &$array);
```

tests/Tests/Unit/ArrayTraitTest.php
```diff
@@ -72,25 +72,25 @@ public function testPluckArray()
         ];
         $expectedArray = $array;
 
-       $actualValues = $this->implementation->call('pluckArray', [$keys, &$array]);
+       $actualValues = $this->implementation->pluckArray($keys, $array);
```
vishwarajanand, line 75:
```suggestion
       $actualValues = $this->implementation->pluckArray($keys, &$array);
```

vishwarajanand (approved):
Lgtm

tests/Tests/Unit/ArrayTraitTest.php
bshaffer: It's not necessary to supply the `&` on this side, since it's part of the `pluck` method signature already

bshaffer: More information: https://www.php.net/manual/en/language.references.pass.php

> Note: There is no reference sign on a function call - only on function definitions. Function definitions alone are enough to correctly pass the argument by reference.

And here as well: https://stackoverflow.com/questions/8971261/php-5-4-call-time-pass-by-reference-easy-fix-available

Call-time pass-by-reference was deprecated in PHP 5.3 and removed entirely in PHP 5.4

PR #457, opened 2023-04-26
feat: add support for quota project env var

bshaffer:
Alternative implementation to https://github.com/googleapis/google-auth-library-php/pull/452/files

This works:
```php
putenv('GOOGLE_CLOUD_QUOTA_PROJECT=my-quota-project');
// outputs "my-quota-project"
echo CredentialsWrapper::build()->getQuotaProject();
```
This does not:
```php
putenv('GOOGLE_CLOUD_QUOTA_PROJECT=my-quota-project');
// outputs value of "quota_project_id" in the JSON keyfile, or null if one doesn't exist
echo CredentialsWrapper::build(['keyFile' => $keyFilePath])->getQuotaProject(); 
```
But this does:
```php
putenv('GOOGLE_CLOUD_QUOTA_PROJECT=my-quota-project');
putenv('GOOGLE_APPLICATION_CREDENTIALS=' . $keyFilePath);
// outputs "my-quota-project" regardless of what is in the JSON keyfile
echo CredentialsWrapper::build()->getQuotaProject(); 
```

cc @TimurSadykov

### DIFF
--- a/src/CredentialsWrapper.php
+++ b/src/CredentialsWrapper.php
@@ -268,7 +268,7 @@ class CredentialsWrapper
                 $authHttpHandler,
                 $authCacheOptions,
                 $authCache,
-                $quotaProject,
+                $quotaProject ?: self::getQuotaProjectFromEnvVar(),
                 $defaultScopes
             );
         } catch (DomainException $ex) {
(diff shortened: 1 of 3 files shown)


bshaffer:
@sai-sunder-s the main reason is that the env var is `GOOGLE_CLOUD_`, so it doesn't seem like it should apply to generic auth rules.

sai-sunder-s:
Auth library in some languages already support `GOOGLE_CLOUD_` env vars. eg: `GOOGLE_CLOUD_PROJECT`.

`GOOGLE_CLOUD_QUOTA_PROJECT` is part of [AIP](https://google.aip.dev/auth/4110).

bshaffer:
That is not the case in PHP, we support `GOOGLE_CLOUD_PROJECT` only as part of our `google/cloud-core` package.

bshaffer:
@sai-sunder-s sounds good to me! I reopened https://github.com/googleapis/google-auth-library-php/pull/452

bshaffer:
Closing in favor of implementing in `google/auth`

PR #462, opened 2023-05-05
fix: ensure timeout set by RetryMiddleware is int not float

acoulton:
RetrySettings/RetryMiddleware supports a totalTimeoutMs option which caps the overall timeout for all retries. At the moment, when this is active it can result in RetryMiddleware calculating that the timeout for the next call should be a floating point value.

This is because the calculation includes the current time found by `microtime(true) * 1000`, which returns a floating point value.

When used in combination with the GRPC transport, the timeout value will eventually end up in \Grpc\AbstractCall::__construct(). That class checks that the value passes `is_numeric()` and then attempts to create `\Grpc\Timeval($timeout)`. However that class is typed to take an integer, and so PHP >= 8.1 gives a deprecation error like:

```
Implicit conversion from float 9999998.046875 to int loses precision
 in /path/to/vendor/grpc/grpc/src/lib/AbstractCall.php:55
```

Because the timeout is within an options array it is not strictly typed, but it would appear that it is always expected to be an integer.

This commit explicitly casts the calculated timeout to an int, after rounding down, which resolves the later PHP deprecation error.

### DIFF
--- a/src/Middleware/RetryMiddleware.php
+++ b/src/Middleware/RetryMiddleware.php
@@ -128,7 +128,7 @@ class RetryMiddleware
         }
 
         $delayMs = min($delayMs * $delayMult, $maxDelayMs);
-        $timeoutMs = min(
+        $timeoutMs = (int) min(
             $timeoutMs * $timeoutMult,
             $maxTimeoutMs,
             $deadlineMs - $this->getCurrentTimeMs()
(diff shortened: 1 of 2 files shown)


src/Middleware/RetryMiddleware.php
```diff
@@ -128,11 +128,11 @@ private function retry(Call $call, array $options, string $status)
         }
 
         $delayMs = min($delayMs * $delayMult, $maxDelayMs);
-        $timeoutMs = min(
+        $timeoutMs = (int) floor(min(
```
bshaffer, line 131: Why not just cast the result of `min` to an int? Why do we need the `floor` here? If you cast a float to an int, it already rounds down to the nearest int.

bshaffer (commented):
Thank you for this submission! One question but otherwise this looks great

src/Middleware/RetryMiddleware.php
acoulton: Just that it was the middle of the night & I'd forgotten that casting truncates rather than rounding :) Thanks for the reminder. I will amend.

PR #470, opened 2023-06-06
feat: sets api header "gcloud-php-new" and "gcloud-php-legacy" to differentiate new surface versions

bshaffer:
Adds an entry to `User-Agent` for `gcloud-php-new/[version]` and `gcloud-php-legacy/[version]`. This will have the result of populating the "client_library_name" (currently unused for GAPIC clients) so we can track the migration.

 - [x] verify this shows up in the backend

**Edit**: This shows up as expected, but it changes the client type to "veneer", which is not accurate, so until that is fixed, this PR is blocked.

 - [ ] verify this shows up in the backend without changing the client type to "veneer"


noahdietz (approved):

bshaffer:
We have to verify the data is showing up in the backend as expected before merging this

PR #473, opened 2023-06-16
feat: interface and methods for middleware stack

bshaffer:
Implementation for Option 2 of [go/php-interceptors-and-middlewares](https://goto.google.com/php-interceptors-and-middlewares):

```php
$client->addMiddleware(function (MiddlewareInterface $handler) {
    return new class ($handler) implements MiddlewareInterface {
        public function __construct(private MiddlewareInterface $handler) {
        }

        /**
         * Modify or observe the API call request and response.
         * The returned value must include the result of the next MiddlewareInterface invocation in the
         * chain.
         *
         * @param Call $call
         * @param array $options
         * @return PromiseInterface|ClientStream|ServerStream|BidiStream
         */
        public function __invoke(Call $call, array $options) {
            // do middleware stuff (pre-request)
            $response = ($this->handler)($call, $options);
            // do middleware stuff (post-request)
            return $response;
        }
    };
});
```

### DIFF
--- a/src/Middleware/MiddlewareInterface.php
+++ b/src/Middleware/MiddlewareInterface.php
@@ -1,0 +1,91 @@
+<?php
+/*
+ * Copyright 2023 Google LLC
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met:
+ *
+ *     * Redistributions of source code must retain the above copyright
+ * notice, this list of conditions and the following disclaimer.
+ *     * Redistributions in binary form must reproduce the above
+ * copyright notice, this list of conditions and the following disclaimer
+ * in the documentation and/or other materials provided with the
+ * distribution.
+ *     * Neither the name of Google Inc. nor the names of its
+ * contributors may be used to endorse or promote products derived from
+ * this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+namespace Google\ApiCore\Middleware;
+
+use Google\ApiCore\Call;
+use GuzzleHttp\Promise\PromiseInterface;
+use Google\ApiCore\ClientStream;
+use Google\ApiCore\ServerStream;
+use Google\ApiCore\BidiStream;
+
+/**
+ * Middlewares must take a MiddlewareInterface as their first constructor
+ * argument {@see Google\ApiCore\Middleware\ResponseMetadataMiddleware}, which
+ * represents the next middleware in the chain. This next middleware MUST be
+ * invoked by this MiddlewareInterface, and the result must be returned as part
+ * of the `__invoke` method implementation.
+ *
+ * To create your own middleware, first implement the interface, as well as pass the handler
+ * in through the constructor:
+ *
+ * ```
+ * use Google\ApiCore\Call;
+ * use Google\ApiCore\Middleware\MiddlewareInterface;
+ *
+ * class MyTestMiddleware implements MiddlewareInterface
+ * {
+ *     public function __construct(MiddlewareInterface $handler)
+ *      {
+ *.         $this->handler = $handler;
+ *      }
+ *      public function __invoke(Call $call, array $options)
+ *      {
+ *          echo "Logging info about the call: " . $call->getMethod();
+ *          return ($this->handler)($call, $options);
+ *      }
+ * }
+ * ```
+ *
+ * Next, add the middleware to any class implementing `GapicClientTrait` by passing in a
+ * callable which returns the new middleware:
+ *
+ * ```
+ * $client = new ExampleGoogleApiServiceClient();
+ * $client->addMiddleware(function (MiddlewareInterface $handler) {
+ *     return new MyTestMiddleware($handler);
+ * });
+ * ```
+ */
+interface MiddlewareInterface
+{
+    /**
+     * Modify or observe the API call request and response.
+     * The returned value must include the result of the next MiddlewareInterface invocation in the
+     * chain.
+     *
+     * @param Call $call
+     * @param array $options
+     * @return PromiseInterface|ClientStream|ServerStream|BidiStream
+     */
+    public function __invoke(Call $call, array $options);
+}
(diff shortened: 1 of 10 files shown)


src/Middleware/MiddlewareInterface.php
```diff
@@ -0,0 +1,49 @@
...
+
+namespace Google\ApiCore\Middleware;
+
+use Google\ApiCore\Call;
+use GuzzleHttp\Promise\PromiseInterface;
+use Google\ApiCore\ClientStream;
+use Google\ApiCore\ServerStream;
+use Google\ApiCore\BidiStream;
+
+interface MiddlewareInterface
```
noahdietz, line 41: I think the expectation of implementors is that they invoke the next middleware, is that correct? If so, I think we need some more explicit documentation about this. I don't see it in here that "a MiddlewareInterface implementation **must** call the MiddlewareInterface it is given". Not sure where it goes though.

bshaffer: Yes you are correct!  I struggled with this, because right now the handler is passed in the constructor (e.g. `__construct(MiddlewareInterface $nextHandler)`) and then called in `__invoke`, which is noted by the `PromiseInterface` typehint. 

The issue I ran into is: 
1. The constructors can (and should) differ. So I can't put that in the interface. 
2. The return types can _also_ be `BidiStream`, `ClientStream`, or `ServerStream`, so we can't typehint that either.

Would it be sufficient to expand the documentation to explain these things? I'll go ahead and make that change now.

bshaffer: See https://github.com/googleapis/gax-php/pull/473/commits/75a54947f82958ea875b37b09740f87a38b6b21f

noahdietz: SGTM thanks

src/Middleware/MiddlewareInterface.php
```diff
@@ -38,9 +38,19 @@
...
+/**
+ * Middlewares should take a MiddlewareInterface as a constructor argument
+ * {@see Google\ApiCore\Middleware\ResponseMetadataMiddleware} which represents
+ * the next middleware in the chain. This will be invoked as part of this
+ * middleware's invoke method, and the result returned.
+ */
 interface MiddlewareInterface
 {
     /**
+     * Modify or observe the request and response parameters.
```
noahdietz, line 50: Nit: IDK if it makes sense to be this specific though.
```suggestion
     * Modify or observe the API call request and response.
```

src/Middleware/MiddlewareInterface.php
```diff
@@ -38,9 +38,19 @@
...
+     * The returned value should be the result of the next middleware in the
```
noahdietz, line 51:
```suggestion
     * The returned value must include the result of the next MiddlewareInterface invocation in the
```

Stricter language with "must", and slightly looser language with "include the result", in case **this** MiddlewareInterface wants to modify it in some way.

src/Middleware/MiddlewareInterface.php
```diff
@@ -38,9 +38,19 @@
 use Google\ApiCore\ServerStream;
 use Google\ApiCore\BidiStream;
 
+/**
+ * Middlewares should take a MiddlewareInterface as a constructor argument
+ * {@see Google\ApiCore\Middleware\ResponseMetadataMiddleware} which represents
+ * the next middleware in the chain. This will be invoked as part of this
+ * middleware's invoke method, and the result returned.
```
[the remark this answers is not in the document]
bshaffer: Done in 915e692829987f6cacd35a9a1ba777f60d399e73

noahdietz (approved):
This looks great.

src/Middleware/MiddlewareInterface.php
```diff
@@ -0,0 +1,60 @@
...
+use Google\ApiCore\ClientStream;
+use Google\ApiCore\ServerStream;
+use Google\ApiCore\BidiStream;
+
+/**
+ * Middlewares must take a MiddlewareInterface as their first constructor
+ * argument {@see Google\ApiCore\Middleware\ResponseMetadataMiddleware}, which
+ * represents the next middleware in the chain. This next middleware MUST be
+ * invoked by this MiddlewareInterface, and the result must be returned as part
+ * of the `__invoke` method implementation.
```
bshaffer, line 46:
```suggestion
 * of the `__invoke` method implementation.
 *
 * To create your own middleware, first implement the interface, as well as pass the handler
 * in through the constructor:
 * ```
 * use Google\ApiCore\Call;
 * use Google\ApiCore\Middleware\MiddlewareInterface;
 * 
 * class MyTestMiddleware implements MiddlewareInterface
 * {
 *     public function __construct(MiddlewareInterface $handler)
 *      {
 *.         $this->handler = $handler;
 *      }
 *      public function __invoke(Call $call, array $options)
 *      {
 *         echo "Logging info about the call: " . $call->getMethod();
 *          return ($this->handler)($call, $options);
 *      }
 * }
 * ```
 * Next, add the middleware to any class implementing `GapicClientTrait` by passing in a 
 * callable which returns the new middleware:
 * 
 * ```
 * $client = new ExampleGoogleApiServiceClient();
 * $client->addMiddleware(function (MiddlewareInterface $handler) {
 *     return new MytestMiddleware($handler);
 * });
 * ```
```

src/Middleware/MiddlewareInterface.php
```diff
@@ -44,6 +44,37 @@
...
+ * ```
+ *
+ * Next, add the middleware to any class implementing `GapicClientTrait` by passing in a
+ * callable which returns the new middleware:
+ *
+ * ```
+ * $client = new ExampleGoogleApiServiceClient();
+ * $client->addMiddleware(function (MiddlewareInterface $handler) {
+ *     return new MyTestMiddleware($handler);
+ * });
```
noahdietz, line 76: Ok I know this is kind of a last minute suggestion, but...do you think it would make more sense to add support for this via Client options (at `new Client()` time)?

My thinking is as follows
- client configuration is already done there
- only supporting it there makes it more "concurrency" safe (i.e. a user can only add Middlewares at create and not after it's been created and potentially already in use e.g. what happens if a user has inflight calls, but they add a middleware?)

WDYT?

bshaffer: Hmm, that's interesting. So it would look something like:

```php
$client = new SomeGoogleApiServiceClient([
    'middlewares' => [
        function (MiddlewareInterface $handler) {
           return new MyMiddleware1($handler);
        },
        function (MiddlewareInterface $handler) {
           return new MyMiddleware2($handler);
        },
    ]
);
```

noahdietz: Or, given that they have the necessary constructor, just like
```php
$client = new SomeGoogleApiServiceClient([
    'middlewares' => [
        MyMiddleware::class,
        MyOtherMiddleware::class,
    ]
);
```
and we init them in the loop in `createCallStack` in the same way. That said, I'm more interested in what you think of it as a client setting than the actual way it looks as a value there.

The primary benefits I see are
- it is slightly less error prone compared to needing to call `addMiddleware` in the right spot
- it simplifies things in that users cannot get too wacky with Middleware initialization (since it happens on every call) when all that they give us is the `class` that fulfills our interface (think what might happen if they had some heavy code in the `function (MiddlewareInterface $handler): MiddlewareInterface` function they give to `addMiddleware`)

The primary drawbacks I see are
- my suggestion of giving only `::class` it is more strict/limiting, might not be bad though
- users are forced to create a new client when they want to add middlewares, might not be bad though

bshaffer: I like it... I think it makes sense as a constructor option. You wouldn't want a client to function with conditional middlewares

noahdietz: Sorry to give this late feedback!

bshaffer: One problem with this approach is that if the users aren't creating the middlewares, just supplying the class names, then theres no way for them to pass custom arguments to them. For instance, if I wanted to pass a  `GuzzleHttp\Client` to my middleware, which I could mock during testing (or provide different credentials depending on the environment), I wouldn't be able to.

So I think we should stick with the `callable` approach. WDYT?

noahdietz: > For instance, if I wanted to pass a GuzzleHttp\Client to my middleware, which I could mock during testing (or provide different credentials depending on the environment), I wouldn't be able to.

I called that out as a drawback above, but I also am not sure its really a loss. Do we really want users to be using some other client in the middleware? Mocking might be a good reason to do it though I wonder if spoofing our Transport provider is more complete. These middlewares are initialized at the start at every API call so the heavier the Middleware, the more it impacts performance.

I'm happy with the `callable` approach  though if you think it adds value without creating too much risk, I'm with you.

bshaffer: A better example would be for the LoggingMiddleware, you would want to use some class to do the logging - maybe an existing library. Configuring and passing it in per environment would be necessary, so it would need access to external configuration  somehow.

noahdietz: Great call. I agree that sounds necessary. Let's keep the `callable` strategy.

fiboknacky (approved):

PR #479, opened 2023-07-18
feat: add +s1 and +s2 to identify previous surfaces

bshaffer:
Replacement for https://github.com/googleapis/gax-php/pull/470

The above PR is blocked because it has the unwanted affect of changing the client type to "veneer". This change will not have this affect, while still allowing us to query for this data in a separate table.

 - [ ] verify this shows up in the backend


noahdietz (approved):

bshaffer:
This has been superseded by #470

PR #483, opened 2023-08-04
fix: deprecation notice while GapicClientTrait->setClientOptions

vishwarajanand:
Fix for the following deprecation notice:
```
PHP Deprecated:  Creation of dynamic property Google\Cloud\PubSub\V1\PublisherClient::$newClient 
is deprecated in /Users/vishwarajanand/github/google-cloud-php/PubSub/vendor/google/gax/src/GapicClientTrait.php 
on line 1077
```

### DIFF
--- a/src/GapicClientTrait.php
+++ b/src/GapicClientTrait.php
@@ -81,6 +81,7 @@ trait GapicClientTrait
         Call::SERVER_STREAMING_CALL => 'startServerStreamingCall',
     ];
     private bool $isNewClient;
+    private $newClient;
 
     /**
      * Initiates an orderly shutdown in which preexisting calls continue but new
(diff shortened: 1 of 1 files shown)


src/GapicClientTrait.php
```diff
@@ -81,6 +81,7 @@ trait GapicClientTrait
         Call::SERVER_STREAMING_CALL => 'startServerStreamingCall',
     ];
     private bool $isNewClient;
+    private $newClient;
```
ajupazhamayil, line 84: Just a doubt, where does this get instantiated? Or any setter for this in this trait?

bshaffer: Woops, this is a typo. But as you can see, above this is the variable `$isNewClient`, so what we want is to use `$isNewClient` where we use `$newClient`.

bshaffer (changes requested):

src/GapicClientTrait.php
bshaffer, line 84:
```suggestion
```

bshaffer (approved):
I made the correct fix in https://github.com/googleapis/gax-php/pull/483/commits/4bcb4ccb6bf90050bf77fd36d9df404873883b14

PR #452, opened 2023-04-08
chore: add property typehints

bshaffer:
- [x] Depends on #451

PR #489, opened 2023-09-20
feat: Add custom retries in gax

saranshdhingra:
Adds custom retries in gax.


src/Middleware/RetryMiddleware.php
```diff
@@ -86,12 +95,27 @@ public function __invoke(Call $call, array $options)
...
-                throw $e;
-            }
-
-            if (!in_array($e->getStatus(), $this->retrySettings->getRetryableCodes())) {
-                throw $e;
+            $customRetryFunc = $this->retrySettings->getRequestRetryFunction();
+            // If we have a custom retry function, and it returns false
+            // we throw the exception, otherwise we continue.
+            if (!is_null($customRetryFunc)) {
+                if (!$customRetryFunc($e, $options, $this->currentAttempts)) {
```
bshaffer, line 102: The function signature of the custom retry function is not documented anywhere. It seems like it should be something like `function (Exception $e, RetrySettings $options, int $retryAttempt): bool` ? 

For guidance, take a look at the [exponential backoff](https://github.com/googleapis/google-cloud-php/blob/main/Core/src/ExponentialBackoff.php#L100)

src/Middleware/RetryMiddleware.php
```diff
@@ -51,14 +51,23 @@ class RetryMiddleware
     /** @var float|null */
     private $deadlineMs;
 
+    /*
+     * The number of API calls(not retries) that have already been attempted.
+     * The first retry will have $currentAttempts set to 1.
+     * @var int
+     */
+    private $currentAttempts;
```
bshaffer, line 59: It seems like this should be named `$retryAttempt` instead of `$currentAttempts`? 

see [ExponentialBackoff](https://github.com/googleapis/google-cloud-php/blob/e9ed1a156a32a07bc6231001fa431d1c162f452f/Core/src/ExponentialBackoff.php#L100)

src/Middleware/RetryMiddleware.php
```diff
@@ -86,12 +95,27 @@ public function __invoke(Call $call, array $options)
...
+                // This is the default retry behaviour, i.e. we don't retry an ApiException
+                // and for other exception types, we only retry when the error code is in
+                // the list of retryable error codes.
+                if (!$e instanceof ApiException) {
+                    throw $e;
+                }
+
+                if (!in_array($e->getStatus(), $this->retrySettings->getRetryableCodes())) {
+                    throw $e;
+                }
```
bshaffer, line 118: Just a suggestion, but instead of `$customRetryFunc` we should just call it `retryFunction`, and have a private method `getDefaultRetryFunction()` with the following:

```php
return function (Exception $e, int $retryAttempt): bool {
    // This is the default retry behaviour, i.e. we don't retry an ApiException
    // and for other exception types, we only retry when the error code is in
    // the list of retryable error codes.
    if (!$e instanceof ApiException) {
        throw $e;
    }

    if (!in_array($e->getStatus(), $this->retrySettings->getRetryableCodes())) {
        throw $e;
    }
};
```

bshaffer (changes requested):
Just some small suggestions, but before we merge this we need to have a corresponding design doc.

src/Middleware/RetryMiddleware.php
saranshdhingra: Actually this included the original API attempt to reflect the number of total attempts to do the operation. I'll change it back to how it is in the ExponentialBackoff file.

src/Middleware/RetryMiddleware.php
saranshdhingra: Thanks for the suggestion.

src/Middleware/RetryMiddleware.php
saranshdhingra: Thanks for the suggestion. Incorporated.

src/Middleware/RetryMiddleware.php
```diff
@@ -51,14 +51,23 @@ class RetryMiddleware
     /** @var float|null */
     private $deadlineMs;
 
+    /*
+     * The number of retries that have already been attempted.
+     * The first retry will have $retryAttempts set to 0.
+     * @var int
+     */
+    private $retryAttempts;
```
bshaffer, line 59: We can use typehints! I have a [PR for the rest of the library](https://github.com/googleapis/gax-php/pull/452), but as we add new ones I think it's good to add them: 
```suggestion
    private int $retryAttempts;
```

src/RetrySettings.php
```diff
@@ -217,6 +217,13 @@ class RetrySettings
 
     private $noRetriesRpcTimeoutMillis;
 
+    /**
+     * When set, this function will be used to evaluate if the retry should
+     * take place or not. The callable will have the following signature:
+     * function (Exception $e, array $options, int $retryAttempt): bool
+     */
+    private $requestRetryFunction;
```
bshaffer, line 225: nit: rename all these to `retryFunction`
```suggestion
    private $retryFunction;
```

src/Middleware/RetryMiddleware.php
```diff
@@ -51,14 +51,23 @@ class RetryMiddleware
     /** @var float|null */
     private $deadlineMs;
 
+    /*
+     * The number of retries that have already been attempted.
+     * The first retry will have $retryAttempts set to 0.
+     * @var int
+     */
+    private int $retryAttempts;
```
vishwarajanand, line 59: We don't have other precedents of adding type hints in php7.4 style in this repo. We can either choose to keep `private int` or the `@var int`, both seems redundant.

```suggestion
    private $retryAttempts;
```

vishwarajanand (approved):

src/Middleware/RetryMiddleware.php
```diff
@@ -51,14 +51,24 @@ class RetryMiddleware
     /** @var float|null */
     private $deadlineMs;
 
+    /*
+     * The number of retries that have already been attempted.
+     * The first retry will have $retryAttempts set to 0.
+     *
+     * @experimental
```
bshaffer, line 58: There's no real need to mark a private variable as experimental (since it's not visible outside the code anyway). However, marking the `$retryAttemtps` variable in the constructor argument as experimental is okay.
```suggestion
```

src/Middleware/RetryMiddleware.php
```diff
@@ -155,4 +167,26 @@ protected function getCurrentTimeMs()
...
         return microtime(true) * 1000.0;
     }
+
+    /**
+     * This is the default retry behaviour.
+     */
+    private function getRetryFunction()
+    {
+        return $this->retrySettings->getRetryFunction() ??
+            function (\Exception $e, array $options, int $retryAttempt): bool {
```
bshaffer, line 177: We are passing in `$retryAttempts` but not using it? I am assuming this is because we are going to pass in custom retry functions in handwritten libraries which support the `$retries` argument? But if we do that, we might as well just add a `maxRetries` option to `RetrySettings` and support it as well.  What's the intention here?

src/RetrySettings.php
```diff
@@ -489,6 +505,11 @@ public function getTotalTimeoutMillis()
         return $this->totalTimeoutMillis;
     }
 
+    public function getRetryFunction()
```
bshaffer, line 508: Same as above - let's mark this as `@experimental`, as well as in the options array (although tags aren't possible there, so just write "This is Experimental" or something similar), and remove it from the private variable.

tests/Tests/Unit/Middleware/RetryMiddlewareTest.php
```diff
@@ -275,4 +275,103 @@ public function testNoRetryLogicalTimeout()
...
+    {
+        $call = $this->getMockBuilder(Call::class)
+            ->disableOriginalConstructor()
+            ->getMock();
+
+        $maxAttempts = 3;
+        $retrySettings = RetrySettings::constructDefault()
+            ->with([
+                'retriesEnabled' => true,
+                'retryFunction' => function ($ex, $options, $attempts) use($maxAttempts) {
```
bshaffer, line 289: nit:
```suggestion
                'retryFunction' => function ($ex, $options, $attempts) use ($maxAttempts) {
```

tests/Tests/Unit/Middleware/RetryMiddlewareTest.php
```diff
@@ -275,4 +275,103 @@ public function testNoRetryLogicalTimeout()
...
+            ->with([
+                'retriesEnabled' => false,
+                'retryFunction' => function ($ex, $options, $attempts) {
+                    // This should not run as retriesEnabled is false
+                    $this->fail('Custom retry function shouldn\'t have run.');
+                    return true;
+                }
+            ]);
+
+        $handler = function(Call $call, $options) {
```
bshaffer, line 329: not:
```suggestion
        $handler = function (Call $call, $options) {
```

tests/Tests/Unit/Middleware/RetryMiddlewareTest.php
```diff
@@ -275,4 +275,103 @@ public function testNoRetryLogicalTimeout()
...
+                'retriesEnabled' => true,
+                'totalTimeoutMillis' => 1,
+                'retryFunction' => function ($ex, $options, $attempts) {
+                    usleep(900);
+                    return true;
+                }
+            ]);
+
+        $callCount = 0;
+        $handler = function(Call $call, $options) use(&$callCount) {
```
bshaffer, line 357: nit: 
```suggestion
        $handler = function (Call $call, $options) use(&$callCount) {
```

tests/Tests/Unit/Middleware/RetryMiddlewareTest.php
```diff
@@ -275,4 +275,103 @@ public function testNoRetryLogicalTimeout()
...
+            return new Promise(function () use(&$callCount) {
+                ++$callCount;
+                throw new ApiException('Call count: ' . $callCount, 0, '');
+            });
+        };
+        $middleware = new RetryMiddleware($handler, $retrySettings);
+
+        // $this->expectException(ApiException::class);
+
```
bshaffer, line 366: remove outdated comment
```suggestion
```

bshaffer (changes requested):
Looking good, just a few minor changes (and a question)

src/Middleware/RetryMiddleware.php
saranshdhingra: So, I remember we were doing that in the PR where we deprecated the connection classes and then we moved to making the properties as experimental.

Hence I followed the same here as otherwise the method gets marked as experimental, right?

bshaffer: Yes, but the difference here is that the property is private

src/Middleware/RetryMiddleware.php
bshaffer: I disagree here. The reason the precedent is to not have typehints is because we previously were supporting PHP versions which didn't support typehints, so adding them would have thrown a PHP Syntax Error. However, we've now dropped support for those versions of PHP, and adding typehints is the _new_ precedent. I even have a PR which adds typehints to all properties (https://github.com/googleapis/gax-php/pull/452). 

So, when adding new properties, we absolutely want to add typehints.

src/Middleware/RetryMiddleware.php
```diff
@@ -86,14 +94,17 @@ public function __invoke(Call $call, array $options)
...
-                throw $e;
-            }
+            $retryFunction = $this->getRetryFunction();
 
-            if (!in_array($e->getStatus(), $this->retrySettings->getRetryableCodes())) {
+            // If the number of retries has surpassed the max allowed retries,
+            // or if the retry function returns false then throw the
+            // exception as we normally would.
+            if ($this->retryAttempts >= $this->retrySettings->getMaxRetries() ||
+                !$retryFunction($e, $options)) {
```
bshaffer, line 103: when `maxRetries` is `0`, this represents infinite retries. So this logic needs to be updated for that case. 

Also, we can keep these as two separate cases, for readability
```suggestion
            if (0 !== $this->retrySettings->getMaxRetries()
                && $this->retryAttempts >= $this->retrySettings->getMaxRetries()
            ) {
                throw $e;
            }
            if (!$retryFunction($e, $options)) {
                throw $e;
            }
```

src/RetrySettings.php
```diff
@@ -203,6 +205,8 @@ class RetrySettings
 {
     use ValidationTrait;
 
+    const DEFAULT_MAX_RETRIES = 3;
```
bshaffer, line 208: Default should be `0` (unlimited retries) rather than `3`, which will change existing behavior.

src/Middleware/RetryMiddleware.php
saranshdhingra: Update on this. We finally did add the maxRetries as we saw many uses for that. This is however considered experimental for now.

src/RetrySettings.php
```diff
@@ -224,6 +224,7 @@ class RetrySettings
     /**
      * The number of maximum retries an operation can do.
      * This doesn't include the original API call.
+     * Setting this 0 means no limit.
```
bshaffer, line 227: nit
```suggestion
     * Setting this to 0 means no limit.
```

src/Middleware/RetryMiddleware.php
```diff
@@ -96,11 +96,17 @@ public function __invoke(Call $call, array $options)
         return $nextHandler($call, $options)->then(null, function ($e) use ($call, $options) {
             $retryFunction = $this->getRetryFunction();
 
-            // If the number of retries has surpassed the max allowed retries,
-            // or if the retry function returns false then throw the
+            // If the number of retries has surpassed the max allowed retries
+            // throw the exception as we normally would.
```
bshaffer, line 100: nit
```suggestion
            // then throw the exception as we normally would.
```

bshaffer (approved):

PR #490, opened 2023-09-20
feat: TPC with late binding

bshaffer:
CLOSED in favor of two separate PRs:

 - **LATE BINDING**: https://github.com/googleapis/gax-php/pull/491
 - **TPC**: https://github.com/googleapis/gax-php/pull/492


composer.json
```diff
@@ -7,7 +7,7 @@
     "license": "BSD-3-Clause",
     "require": {
         "php": ">=7.4",
-        "google/auth": "^1.25.0",
+        "google/auth": "dev-add-universe-domain-phase1 as 1.31.0",
```
bshaffer, line 10: For testing only. TODO: remove this

src/GapicClientTrait.php
```diff
@@ -250,12 +264,37 @@ private function buildClientOptions(array $options)
 
         // mTLS: If no apiEndpoint has been supplied by the user, and either
         // GOOGLE_API_USE_MTLS_ENDPOINT tells us to, or mTLS is available, use the mTLS endpoint.
-        if ($options['apiEndpoint'] === $defaultOptions['apiEndpoint']
-            && $this->shouldUseMtlsEndpoint($options)
-        ) {
-            $options['apiEndpoint'] = self::determineMtlsEndpoint($options['apiEndpoint']);
+        if (is_null($apiEndpoint) && $this->shouldUseMtlsEndpoint($options)) {
```
bshaffer, line 267: **NOTE** - we previously checked if a user supplied the API endpoint by seeing if it was different from the default. Now we are checking to see if it's supplied in the options array. This requires consideration.

In this case, if a user supplied the endpoint directly, even if it was the same as the default, this would mean it should be used instead of the mTLS endpoint, so while this is a change of behavior, to me it seems like a bug fix.  However, if we want to be extra safe, we can revert this.

src/GapicClientTrait.php
```diff
@@ -250,12 +264,37 @@ private function buildClientOptions(array $options)
...
 
+        $this->apiEndpointFunction = Closure::fromCallable(function () use ($apiEndpoint, $defaultOptions) {
+            // If a custom endpoint is set, or we are using an MTLS endpoint, return it
+            if ($apiEndpoint) {
+                return $apiEndpoint;
+            }
+
+            // If no API endpoint is set, derive the endpoint from the service address template and the
+            // universe domain fetched from the credentials.
+            if (defined('self::SERVICE_ADDRESS_TEMPLATE')) {
```
bshaffer, line 279: This is TDB but the idea is this will be a new constant on GAPIC clients for constructing the API endpoint from the universe domain.

PR #491, opened 2023-09-26
feat: late binding for transport later

bshaffer:
add support for late binding to avoid early calls to metadata server


src/GapicClientTrait.php
```diff
@@ -779,6 +810,10 @@ private function createCallStack(array $callConstructionOptions)
         }
         $callStack = function (Call $call, array $options) {
             $startCallMethod = $this->transportCallMethods[$call->getCallType()];
+            if (!$this->transport) {
+                // Create transport at call-time to allow for lazy initialization
+                $this->transport = $this->buildTransport();
+            }
             return $this->transport->$startCallMethod($call, $options);
```
vishwarajanand, line 817:
```suggestion
            return $this->getTransport()->$startCallMethod($call, $options);
```

bshaffer: I forgot that this was done intentionally to prevent breaking changes in the event that someone overrode the getTransport method in a client subclass.

bshaffer:
I am closing this PR, as late binding for transports is no longer required.

PR #492, opened 2023-09-26
feat: TPC with late binding

bshaffer:
Adds support for TPC by determining the API endpoint using the new `SERVICE_ADDRESS_TEMPLATE` constant in GAPIC clients and the "universe domain" from the auth credentials. 

Requirements
 - [ ] depends on https://github.com/googleapis/google-auth-library-php/pull/477
 - [ ] depends on https://github.com/googleapis/gax-php/pull/491
 - [ ] depends on https://github.com/googleapis/gapic-generator-php/pull/658

**NOTE**: It may be better to abstract some of this out so that the Handwritten clients can take advantage as well. For instance, in an `ApiEndpointFactory`.


bshaffer:
Late Binding is no longer required, will close this and submit one without it.

PR #498, opened 2023-10-17
chore: permissions for docs.yml workflow

bshaffer:
Adds write permission to repository contents. Satisfies security while not breaking the job

https://github.com/googleapis/gax-php/pull/454

See https://docs.github.com/en/actions/using-jobs/assigning- permissions-to-jobs


noahdietz (approved):

bshaffer:
cc @vishwarajanand

PR #500, opened 2023-11-07
chore(ci): add workflow for running google/cloud tests before releases

bshaffer:
- [x] depends on https://github.com/googleapis/google-cloud-php/pull/6773

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -27,7 +27,7 @@ jobs:
       run: |
         cd google-cloud-php
         composer install -q -d dev
-        dev/google-cloud update-deps google/gax 'dev-run-google-cloud-php-tests as 1.200.0' --add=dev
+        dev/google-cloud update-deps google/gax 'dev-main as 1.200.0' --add=dev
     - name: Run google/cloud package tests
       run: |
         cd google-cloud-php
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -8,8 +8,7 @@ jobs:
   release-suite:
     runs-on: ubuntu-latest
     name: Run googleapis/google-cloud-php tests against latest version
-    # @TODO: Uncomment before merge
-    # if: github.event.pull_request.user.login == 'release-please[bot]'
+    if: github.event.pull_request.user.login == 'release-please[bot]'
     steps:
     - uses: actions/checkout@v4
     - name: Clone googleapis/google-cloud-php
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,7 +7,7 @@ permissions:
 jobs:
   release-suite:
     runs-on: ubuntu-latest
-    name: Run googleapis/google-cloud-php tests against latest version
+    name: Run google-cloud-php tests
     if: github.event.pull_request.user.login == 'release-please[bot]'
     steps:
     - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


src/CredentialsWrapper.php
```diff
@@ -54,7 +54,7 @@ class CredentialsWrapper
     use ValidationTrait;
 
     /** @var FetchAuthTokenInterface $credentialsFetcher */
-    private FetchAuthTokenInterface $credentialsFetcher;
+    private ?FetchAuthTokenInterface $credentialsFetcher = null;
```
vishwarajanand, line 57: This may've been added by mistake?

.github/workflows/release.yml
```diff
@@ -0,0 +1,36 @@
...
+      uses: shivammathur/setup-php@v2
+      with:
+        php-version: '8.1'
+        extensions: grpc
+    - name: Configure GAX dev-main
+      # @TODO change version to be "dev-main" before we merge
+      run: |
+        cd google-cloud-php
+        composer install -q -d dev
+        dev/google-cloud update-deps google/gax 'dev-run-google-cloud-php-tests as 1.200.0' --add=dev
```
vishwarajanand, line 32: Should this also update auth? maybe this need to do 2 `update-devs` command?

src/CredentialsWrapper.php
bshaffer: No, it was required in order for the tests to pass - this was throwing an error when running these tests (see [this failing test](https://github.com/googleapis/gax-php/actions/runs/6789914294/job/18458150731#step:6:1569))

.github/workflows/release.yml
bshaffer: I think we should have a separate PR to add `auth` as a presubmit check for the auth library. E.g., GAX has a presubmit check to run the google-cloud-php library tests against itself before each release, and the auth library has a presubmit check to run against the google-cloud-php library before each release. 

The auth library could ALSO run the presubmit tests against THIS library before each release (because it's upstream to both libraries), but we wouldn't want to test auth in this check, since it'd make more sense to have the auth library handle it, and doing it here also would be redundant.

.github/workflows/release.yml
```diff
@@ -0,0 +1,34 @@
...
```
bshaffer, line 30: Depends on https://github.com/googleapis/gax-php/pull/503
```suggestion
        dev/google-cloud update-deps google/gax 'dev-main as 1.200.0' --add=dev
```

vishwarajanand:
Changes looks good to me

vishwarajanand (approved):

.github/workflows/release.yml
```diff
@@ -0,0 +1,34 @@
...
+  pull_request:
+  workflow_dispatch:
+permissions:
+  contents: read
+jobs:
+  release-suite:
+    runs-on: ubuntu-latest
+    name: Run googleapis/google-cloud-php tests against latest version
+    # @TODO: Uncomment before merge
+    # if: github.event.pull_request.user.login == 'release-please[bot]'
```
bshaffer, line 12:
```suggestion
    if: github.event.pull_request.user.login == 'release-please[bot]'
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,33 @@
+name: Release Pre-Check
+on:
+  pull_request:
+  workflow_dispatch:
+permissions:
+  contents: read
+jobs:
+  release-suite:
+    runs-on: ubuntu-latest
+    name: Run googleapis/google-cloud-php tests against latest version
```
bshaffer, line 10:
```suggestion
    name: Run google-cloud-php tests
```

PR #502, opened 2023-11-14
feat: add support for universe domain

bshaffer:
- [x] CL-R1 Client Libraries must allow the end-user to configure the endpoint override alongside the universe
 - [x] CL-R2  Client Libraries must use the universe_domain variable/property/method name
 - [x] CL-R3 Client Libraries must use the universe_domain to form the service endpoint and should deprecate the hardcoded endpoint values
 - [x] CL-R5 Client Libraries must use the universe_domain from the credentials to verify that the credentials can be used with the configured universe domain
     - CL-R5.1 If the end-user has not configured the universe_domain explicitly, the Client Libraries must use GDU as the configured universe domain.
     - CL-R5.2 The client libraries must not use the endpoint to validate the universe domain
  - [x] ~~CL-R6 Client Libraries should surface either the credentials or the universe_domain to the end-user~~
      - _this is not currently supported (there are no "getters" on the client classes). I personally don't want to add support_
      - ~~CL-R6.1 Client Libraries should surface the service endpoint to the end-user~~
          - _this is not currently supported (there are no "getters" on the client classes). I personally don't want to add support for it unless it's a hard requirement_
  - [x] CL-R7 Client Libraries must allow the end-user to configure the universe domain explicitly
  - [x] ~~CL-R8 Client Libraries must update the documentation if they allow the end-user to pass the authentication token/JWT instead of credentials~~ - _N/A_
  - [x] CL-R9 Libraries must avoid leaking the authentication tokens and user data to the GDU 
  - [x] CL-R10: Configuring mTLS with a universe domain other than GDU must throw an exception


composer.json
```diff
@@ -7,7 +7,7 @@
     "license": "BSD-3-Clause",
     "require": {
         "php": ">=7.4",
-        "google/auth": "^1.25.0",
+        "google/auth": "dev-add-universe-domain-phase2 as 1.31.0",
```
bshaffer, line 10: This is here until we merge https://github.com/googleapis/google-auth-library-php/pull/477, at which point we need to update it to the corresponding auth library release

bshaffer, line 10: We can check against the main branch before the release
```suggestion
        "google/auth": "dev-main as 1.33.0",
```

src/CredentialsWrapper.php
```diff
@@ -232,6 +254,26 @@ public function getAuthorizationHeaderCallback($audience = null)
         };
     }
 
+    /**
+     * Verify that the expected universe domain matches the universe domain from the credentials.
+     */
+    private function checkUniverseDomain()
```
vishwarajanand, line 260: We don't seem to have a way to differentiate UD passed from credentials file or was user supplied to check for the following scenario:

If a user passes in the universeDomain client option WITH custom credentials which do not implement GetUniverseDomainInterface, the check would not occur, increasing the chance of credentials not honoring UD config.

Do you think we should add this in comments?

bshaffer: Sorry for the confusion. This sentence was in consideration of a _different_ implementation that was rejected.

```
If a user passes in the universeDomain client option WITH custom credentials which do not implement GetUniverseDomainInterface, the check would not occur, increasing the chance of credentials not honoring UD config.
```

I will add a test for the scenario above, to ensure that it's not the case.

src/GapicClientTrait.php
```diff
@@ -195,44 +201,31 @@ private function buildClientOptions(array $options)
...
+                $apiEndpoint = $options['apiEndpoint'];
+            }
 
-        // serviceAddress is now deprecated and acts as an alias for apiEndpoint
-        if (isset($options['serviceAddress'])) {
-            $options['apiEndpoint'] = $this->pluck('serviceAddress', $options, false);
+            // serviceAddress is now deprecated and acts as an alias for apiEndpoint
+            if (isset($options['serviceAddress'])) {
+                $apiEndpoint = $this->pluck('serviceAddress', $options, false);
+            }
```
vishwarajanand, line 216: It seems the order is incorrect. The `$apiEndpoint` should first look at `serviceAddress` from options and then the `apiEndpoint` not the other way around.

src/Options/ClientOptions.php
```diff
@@ -175,6 +179,7 @@ private function fromArray(array $arr): void
         $this->setLibVersion($arr['libVersion'] ?? null);
         $this->setGapicVersion($arr['gapicVersion'] ?? null);
         $this->setClientCertSource($arr['clientCertSource'] ?? null);
+        $this->setUniverseDomain($arr['universeDomain'] ?? null);
```
vishwarajanand, line 182: I do not see value in not initializing UD by null here.
Like if we know that the default value is gonna be `GetUniverseDomainInterface::DEFAULT_UNIVERSE_DOMAIN` lets use it?

bshaffer:
Dang, I tried renaming the branch to keep the PR open, but github still closed it :-/

src/CredentialsWrapper.php
bshaffer: Looks like [I already added a test for this!](https://github.com/googleapis/gax-php/pull/502/files#diff-d7d06c054fa4a4bc3263a0062e0a341cac26d42d6958662fb79f699a973de523R262)

src/GapicClientTrait.php
bshaffer: I agree, but unfortunately, the way we have it implemented is _currently_ in this order, and changing that could potentially break the existing behavior.

src/Options/ClientOptions.php
bshaffer: This is up for debate, but the reason this choice was made is that I wanted the `ClientOptions` to be dumb. If a parameter isn't supplied, it's `null`. Always. They are options, and options are optional, so they're nullable. The client option represent what the user supplied, not what the default values should be. This is important because a user could supply `null`, or they could supply `googleapis.com`, and if we set the default, we wouldn't be able to know if they supplied it or not. Now, that may not matter (and probably doesn't), but that was the rationale here.

Furthermore, I didn't want default values to be defined in `ClientOptions`, but rather by the classes which use them. So in this case, the default universe domain is being defined by `GapicClientTrait`, which I think makes more sense.

vishwarajanand (approved):
Lgtm

composer.json
```diff
@@ -7,7 +7,7 @@
     "license": "BSD-3-Clause",
     "require": {
         "php": ">=7.4",
-        "google/auth": "^1.25.0",
+        "google/auth": "dev-main as 1.33.0",
```
bshaffer, line 10:
```suggestion
        "google/auth": "^1.33.0",
```

composer.json
```diff
@@ -7,7 +7,7 @@
     "license": "BSD-3-Clause",
     "require": {
         "php": ">=7.4",
-        "google/auth": "^1.25.0",
+        "google/auth": "^1.33.0",
```
bshaffer, line 10: We will want to update this to rely on `google/auth:1.34.0` once it's released

PR #520, opened 2024-01-23
feat: allow setting of universe domain in environment variable

bshaffer:


src/ClientOptionsTrait.php
```diff
@@ -178,6 +177,11 @@ private function buildClientOptions(array $options)
             $apiEndpoint = self::determineMtlsEndpoint($options['apiEndpoint']);
         }
 
+        // If the user has not supplied a universe domain, use the environment variable if set.
+        // Otherwise, use the default ("googleapis.com").
+        $options['universeDomain'] ??= getenv('GOOGLE_CLOUD_UNIVERSE_DOMAIN')
```
noahdietz, line 182: TIL this, `??=`, null coalescing variant! I love these types of quality of life features PHP has.

src/ClientOptionsTrait.php
```diff
@@ -178,6 +177,11 @@ private function buildClientOptions(array $options)
...
+            ?: GetUniverseDomainInterface::DEFAULT_UNIVERSE_DOMAIN;
```
noahdietz, line 183:
```suggestion
            ?? GetUniverseDomainInterface::DEFAULT_UNIVERSE_DOMAIN;
```
Should we use null coalescing here as well, b.c the ternary in use rn doesn't have a "truthy" value defined? We'd want the `getenv` return value to be used if it is set, right?

noahdietz (approved):

src/ClientOptionsTrait.php
bshaffer: I love it too! I wanted to write an RFC for it and then found out it was already implemented :D

src/ClientOptionsTrait.php
bshaffer: Because `getenv` returns `false` if it doesn't exist, which means we HAVE to use `?:` because `??` is only false for `null` values. This also means we'd get the default universe domain if someone set the env var to an empty string (e.g. `GOOGLE_CLOUD_UNIVERSE_DOMAIN=`), which I think is also what we want.

noahdietz: Oh I see I see, thanks. That `false` nuance got me. Nice stuff.

PR #523, opened 2024-01-30
feat: implement ProjectIdProviderInterface in CredentialsWrapper

bshaffer:


src/CredentialsWrapper.php
```diff
@@ -202,6 +203,14 @@ public function getQuotaProject()
         return null;
     }
 
+    public function getProjectId(?callable $httpHandler = null): ?string
```
bshaffer, line 206:
```suggestion
    public function getProjectId(callable $httpHandler = null): ?string
```

saranshdhingra (approved):

PR #530, opened 2024-02-14
fix: more Options fixes

bshaffer:


src/Options/TransportOptions.php
```diff
@@ -69,7 +69,7 @@ public function __construct(array $options)
     private function fromArray(array $arr): void
     {
         $this->setGrpc(new GrpcTransportOptions($arr['grpc'] ?? []));
-        $this->setGrpcFallback(new GrpcFallbackTransportOptions($arr['grpc-fallack'] ?? []));
```
noahdietz, line 72: ouch

noahdietz (approved):

src/Options/TransportOptions.php
bshaffer: yep... I guess it's a good thing nobody's using the grpc fallback transport 🤦

PR #534, opened 2024-02-16
fix: universe domain check for grpc transport

### LINKED ISSUE #533: segfault or memory exceeded for gRPC and invalid universe domain
A strange phenomenon is happening when the `universeDomain` is set to a domain that doesn't exist, and the `grpc.so` extension is enabled. If we use a valid universe domain, the call is made as expected. If we use a valid universe domain but one that doesn't match the credentials, we get a universe domain mismatch error as expected. _However_, if the configured universe domain doesn't exist (for example, `doesnt-exist.com`), the process hangs for a very long time before either segfaulting or timing out with "memory exceeded"

```php
$kms = new KeyManagementServiceClient([
    'universeDomain' => 'doesnt-exist.com',
]);

// this should throw a universe domain mismatch error, but instead of hangs forever
$keyRings = $kms->listKeyRings($kms->locationName($project, $location));
```

This is for grpc 1.61.0
```
php -i | grep grpc
grpc
grpc support => enabled
grpc module version => 1.61.0
grpc.enable_fork_support => 0 => 0
grpc.grpc_trace => no value => no value
grpc.grpc_verbosity => no value => no value
grpc.log_filename => no value => no value
grpc.poll_strategy => no value => no value
```

bshaffer:
fixes #533

Checks the universe domain _before_ invoking the grpc libraries. This ensures that a segfault doesn't happen when the universe domain is invalid. 

1. Is this the best way to do this?
2. Should this be addressed in grpc core itself?

### DIFF
--- a/src/Transport/GrpcTransport.php
+++ b/src/Transport/GrpcTransport.php
@@ -159,6 +159,8 @@ class GrpcTransport extends BaseStub implements TransportInterface
      */
     public function startBidiStreamingCall(Call $call, array $options)
     {
+        $this->verifyUniverseDomain($options);
+
         return new BidiStream(
             $this->_bidiRequest(
                 '/' . $call->getMethod(),
@@ -175,6 +177,9 @@ class GrpcTransport extends BaseStub implements TransportInterface
      */
     public function startClientStreamingCall(Call $call, array $options)
     {
+
+        $this->verifyUniverseDomain($options);
+
         return new ClientStream(
             $this->_clientStreamRequest(
                 '/' . $call->getMethod(),
@@ -191,6 +196,8 @@ class GrpcTransport extends BaseStub implements TransportInterface
      */
     public function startServerStreamingCall(Call $call, array $options)
     {
+        $this->verifyUniverseDomain($options);
+
         $message = $call->getMessage();
 
         if (!$message) {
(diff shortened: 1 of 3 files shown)


src/Transport/GrpcTransport.php
```diff
@@ -159,6 +159,8 @@ public static function build(string $apiEndpoint, array $config = [])
      */
     public function startBidiStreamingCall(Call $call, array $options)
     {
+        $this->verifyUniverseDomain($options);
```
noahdietz, line 162: Clarification: Does this need to happen for every call because credentials can come through the call options? Or is there a singular place this could be verified and not on every call?

noahdietz (approved):
Some tests are failing that should be fixed first it seems

src/Transport/GrpcTransport.php
bshaffer: Good question - this was a specific choice because I wanted to avoid adding a network call for the credentials to get the network domain before a call was made - to avoid adding latency where it previously didn't exist. 

As a result, the universe domain is checked at call time. But it's only checked once per request (see [CredentialsWrapper::checkUniverseDomain](https://github.com/googleapis/gax-php/blob/dfca60f46fed8d9a4d9d2ece7286f18d53d3ebad/src/CredentialsWrapper.php#L277) and never again.

noahdietz: awesome, sgtm, good choice!

PR #543, opened 2024-02-27
feat: Auto Populate fields configured for auto population in Rpc Request Message

yash30201:
See b/318607731


tests/Tests/Unit/testdata/test_service_descriptor_config.php
```diff
@@ -11,6 +11,9 @@
                     'requestPageSizeSetMethod' => 'setPageSize',
                     'responsePageTokenGetMethod' => 'getNextPageToken',
                 ],
+                'autoPopulatedFields' => [
+                    'pageToken' => \Google\Api\FieldInfo\Format::UUID4,
+                ]
```
yash30201, line 16: It's been decided that we'll use enum value instead of a string constant like `UUID4`. Ref [comment](https://github.com/googleapis/gapic-generator-php/pull/686#discussion_r1504934386)

composer.json
```diff
@@ -14,7 +14,8 @@
         "guzzlehttp/promises": "^1.4||^2.0",
         "guzzlehttp/psr7": "^2.0",
         "google/common-protos": "^4.4",
-        "google/longrunning": "~0.2"
+        "google/longrunning": "~0.2",
+        "ramsey/uuid": "^4.0"
```
bshaffer, line 18: This dep requires `ext-json`, but apparently so does guzzle, so it shouldn't introduce any dependency errors

src/GapicClientTrait.php
```diff
@@ -633,6 +636,12 @@ private function createCallStack(array $callConstructionOptions)
         $callStack = new CredentialsWrapperMiddleware($callStack, $this->credentialsWrapper);
         $callStack = new FixedHeaderMiddleware($callStack, $fixedHeaders, true);
         $callStack = new RetryMiddleware($callStack, $callConstructionOptions['retrySettings']);
+        if (!empty($callConstructionOptions['autoPopulationSettings'])) {
```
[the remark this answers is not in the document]
bshaffer: Yes, it would be better to just add the middleware like this:

```php
$callstack = new RequestAutoPopulationMiddleware(
    $callstack,
    $callConstructionOptions['autoPopulationSettings']
);
```

Then just have the middleware do a no-op if the settings are empty.

Also, we can ensure `autoPopulationSettings` will always be set (see below).

src/GapicClientTrait.php
```diff
@@ -674,9 +684,12 @@ private function configureCallConstructionOptions(string $methodName, array $opt
...
         }
-        return [
-            'retrySettings' => $retrySettings,
-        ];
+        $result = ['retrySettings' => $retrySettings];
+
+        if (!empty($autoPopulatedFields)) {
+            $result['autoPopulationSettings'] = $autoPopulatedFields;
+        }
+        return $result;
```
bshaffer, line 692: this seems complicated, why not just 
```suggestion
        return [
            'retrySettings' => $retrySettings,
            'autoPopulationSettings' => $autoPopulatedFields,
        ];
```
I don't think there's an advantage between an empty `autoPopulationSettings` and not setting it at all

src/Middleware/RequestAutoPopulationMiddleware.php
```diff
@@ -0,0 +1,92 @@
...
+                switch ($valueType) {
+                    case \Google\Api\FieldInfo\Format::UUID4:
+                        $request->$setFieldName(Uuid::uuid4()->toString());
+                        break;
+                    default:
+                        // Do nothing
+                }
+            }
+        }
+        $call = $call->withMessage($request);
```
bshaffer, line 86: We probably want to wrap this in `if ($this-autoPopulationSettings)` so that executing this with empty settings is a more efficient no-op

bshaffer (commented):
minor suggestions

src/GapicClientTrait.php
yash30201: Doing this would break the `testConfigureCallConstructionOptionsAcceptsRetryObjectOrArray` [test]( https://github.com/googleapis/gax-php/blob/main/tests/Tests/Unit/GapicClientTraitTest.php#L138-L163). The fix would be to just add the `"autoPopulationSettings" => []` entry in [expected options](https://github.com/googleapis/gax-php/blob/main/tests/Tests/Unit/GapicClientTraitTest.php#L143-L146).

Should we do that instead?

src/Middleware/RequestAutoPopulationMiddleware.php
yash30201: I've opened a PR showing this refactoring https://github.com/googleapis/gax-php/pull/545.

yash30201: Merged

tests/Tests/Unit/Middleware/RequestAutoPopulationMiddlewareTest.php
```diff
@@ -0,0 +1,72 @@
...
+
+namespace Google\ApiCore\Tests\Unit\Middleware;
+
+use Google\ApiCore\Call;
+use Google\ApiCore\Middleware\RequestAutoPopulationMiddleware;
+use Google\ApiCore\Testing\MockRequest;
+use PHPUnit\Framework\TestCase;
+use Ramsey\Uuid\Uuid;
+
+class RequestAutoPopulationMiddlewareTest extends TestCase
```
bshaffer, line 41: Let's mark this class as internal
```suggestion
/**
 * @internal
 */ 
class RequestAutoPopulationMiddlewareTest extends TestCase
```

src/GapicClientTrait.php
bshaffer: Yes, we shouldn't write code to match existing tests, but rather update the tests when we make the desired changes to the code. 

LGTM

noahdietz (commented):
Nice job @yash30201 this is pretty cool.

LGTM but let's have @bshaffer stamp

tests/Tests/Unit/Middleware/RequestAutoPopulationMiddlewareTest.php
bshaffer: Woops, this was supposed to be for `RequestAutoPopulationMiddleware`, not for the test. And that is already [marked as internal](https://github.com/googleapis/gax-php/pull/543/files#diff-402e3db24cb32a8f2d6424b029ea9136f0a2380180735a5bd3e6861c94caaf49R430)

src/Middleware/RequestAutoPopulationMiddleware.php
```diff
@@ -0,0 +1,96 @@
...
+
+        if (empty($this->autoPopulationSettings)) {
+            return $next($call, $options);
+        }
+
+        $request = $call->getMessage();
+        foreach ($this->autoPopulationSettings as $fieldName => $valueType) {
+            $getFieldName = 'get' . ucwords($fieldName);
+            // Populate the field if it's not already set by user
+            if (empty($request->$getFieldName())) {
```
bshaffer, line 79: I believe it'd be better to use `$request->hasFieldName()` rather than `empty($request->getFieldName())`.

src/Middleware/RequestAutoPopulationMiddleware.php
```diff
@@ -0,0 +1,96 @@
...
+                $setFieldName = 'set' . ucwords($fieldName);
+                switch ($valueType) {
+                    case \Google\Api\FieldInfo\Format::UUID4:
+                        $request->$setFieldName(Uuid::uuid4()->toString());
+                        break;
+                    default:
+                        // Do nothing
```
bshaffer, line 86: I think we should throw an exception for invalid values, wdyt?

src/Middleware/RequestAutoPopulationMiddleware.php
```diff
@@ -0,0 +1,96 @@
...
+    {
+        $next = $this->nextHandler;
+
+        if (empty($this->autoPopulationSettings)) {
+            return $next($call, $options);
+        }
+
+        $request = $call->getMessage();
+        foreach ($this->autoPopulationSettings as $fieldName => $valueType) {
+            $getFieldName = 'get' . ucwords($fieldName);
```
bshaffer, line 77: This assumes that `$fieldName` from the descriptor is alligator-case, but proto fields (and protobuf message fields in protobuf) are snake-case. I just want to confirm this is guaranteed to be alligator-case.

I believe so, because of [this line](https://github.com/googleapis/gapic-generator-php/pull/686/files#diff-f7058175f24ff4dac2b95e78a8585c7d5ee2ae06e0fb5342dfbfefafc7088398R552)

yash30201: Yes you are correct. It has been made sure in generator that $fieldName would be always in camel-casing.

src/Middleware/RequestAutoPopulationMiddleware.php
yash30201: This only case where I see this happening is when all the the below happens at the same time:
1) Some new enum get's added to the `google.api.field_info.Format`
2) This library's minimum version of common-protos-php isn't updated.
3) The new field_info is used for autopopulation

So we would always want to catch this error!!!

src/Middleware/RequestAutoPopulationMiddleware.php
bshaffer: After much discussion and investigation, we decided this is the right way to implement this, and we should add a clarifying comment such as: "We use a getter instead of a hazzer here because there's no need to differentiate between isset and an empty default value. Even if a field is explicitly set to an empty string, we want to autopopulate it"

src/Middleware/RequestAutoPopulationMiddleware.php
yash30201: Added the logic

src/Middleware/RequestAutoPopulationMiddleware.php
yash30201: Added the comment

src/Middleware/RequestAutoPopulationMiddleware.php
```diff
@@ -0,0 +1,103 @@
...
+                $setFieldName = 'set' . ucwords($fieldName);
+                switch ($valueType) {
+                    case \Google\Api\FieldInfo\Format::UUID4:
+                        $request->$setFieldName(Uuid::uuid4()->toString());
+                        break;
+                    default:
+                        throw new \UnexpectedValueException(sprintf(
+                            "Value type %s::%s not supported for auto population of the field %s",
+                            \Google\Api\FieldInfo\Format::class,
+                            \Google\Api\FieldInfo\Format::name($valueType),
```
bshaffer, line 91: nit, import this so we don't have to use absolute paths

yash30201: Done 🚀

bshaffer (approved):

PR #544, opened 2024-02-28
feat: Make the default authHttpHandler in CredentialsWrapper null

saranshdhingra:
Currently the [v2 samples for Pubsub](https://github.com/GoogleCloudPlatform/php-docs-samples/pull/1967) are failing due to an internal attempt to `serialize` the `PubSubClient` in the `BatchPublisher`.
This is because the default `authHttpHandler` contains a closure. With this PR we make the default handler as `null`.

### DIFF
--- a/src/CredentialsWrapper.php
+++ b/src/CredentialsWrapper.php
@@ -81,7 +81,7 @@ class CredentialsWrapper implements ProjectIdProviderInterface
         string $universeDomain = GetUniverseDomainInterface::DEFAULT_UNIVERSE_DOMAIN
     ) {
         $this->credentialsFetcher = $credentialsFetcher;
-        $this->authHttpHandler = $authHttpHandler ?: self::buildHttpHandlerFactory();
+        $this->authHttpHandler = $authHttpHandler;
         if (empty($universeDomain)) {
             throw new ValidationException('The universe domain cannot be empty');
         }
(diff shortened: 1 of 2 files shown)


yash30201 (approved):
LGTM.

This makes sense as `ApplicationDefaultCredentials` in auth library creates its own httphandlerif null is supplied.

src/CredentialsWrapper.php
```diff
@@ -141,7 +141,7 @@ public static function build(
         ];
 
         $keyFile = $args['keyFile'];
-        $authHttpHandler = $args['authHttpHandler'] ?: self::buildHttpHandlerFactory();
+        $authHttpHandler = $args['authHttpHandler'];
```
bshaffer, line 144: nit: let's simplify this and just make the $authHttpHandler supplied to the loader and `CredentialWrapper` come directly from `$args`

tests/Tests/Unit/CredentialsWrapperTest.php
```diff
@@ -83,7 +83,7 @@ public function buildDataWithoutExplicitKeyFile()
         $appDefaultCreds = getenv('GOOGLE_APPLICATION_CREDENTIALS');
         putenv('GOOGLE_APPLICATION_CREDENTIALS=' . __DIR__ . '/testdata/json-key-file.json');
         $scopes = ['myscope'];
-        $defaultAuthHttpHandler = HttpHandlerFactory::build();
+        $defaultAuthHttpHandler = null;
```
bshaffer, line 86: It defaults to `null`, so why pass this in at all? Seems better to get rid of it and all instances of it being passed in to simplify things.

bshaffer (commented):
Let's add a test that verifies the `CredentialsWrapper` class is able to be serialized by default:

```php
public function testSerializeCredentialsWrapper()
{
    $credentialsWrapper = CredentialsWrapper::build(['credentials' => new ServiceAccoutnCredentials($testJson)]);
    $foo = serialize($credentialsWrapper);
    $this->assertIsString($foo);
}
```

bshaffer (approved):

PR #553, opened 2024-04-18
feat: Add the api header to the GapicClientTrait

Hectorhammett:
This adds the API version header to the `GapicClientTrait` trait for the new api version control.

### DIFF
--- a/src/GapicClientTrait.php
+++ b/src/GapicClientTrait.php
@@ -629,6 +629,13 @@ trait GapicClientTrait
                 'X-Goog-User-Project' => [$quotaProject]
             ];
         }
+        
+        if (isset($this->apiVersion)) {
+            $fixedHeaders += [
+                'X-Goog-Api-Version' => [$this->apiVersion]
+            ];
+        }
+
         $callStack = function (Call $call, array $options) {
             $startCallMethod = $this->transportCallMethods[$call->getCallType()];
             return $this->transport->$startCallMethod($call, $options);
--- a/tests/Tests/Unit/GapicClientTraitTest.php
+++ b/tests/Tests/Unit/GapicClientTraitTest.php
@@ -135,6 +135,71 @@ class GapicClientTraitTest extends TestCase
         );
     }
 
+    public function testVersionedHeadersOverwriteBehavion()
+    {
+        $unaryDescriptors = [
+            'callType' => Call::UNARY_CALL,
+            'responseType' => 'decodeType',
+            'headerParams' => [
+                [
+                    'fieldAccessors' => ['getName'],
+                    'keyName' => 'name'
+                ]
+            ]
+        ];
+        $request = new MockRequestBody(['name' => 'foos/123/bars/456']);
+        $header = AgentHeader::buildAgentHeader([
+            'libName' => 'gccl',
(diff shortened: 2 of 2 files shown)


tests/Tests/Unit/GapicClientTraitTest.php
```diff
@@ -135,6 +135,71 @@ public function testHeadersOverwriteBehavior()
         );
     }
 
+    public function testVersionedHeadersOverwriteBehavion()
```
Hectorhammett, line 138: I am just testing for the `VersionedStubGapicClient` as the expected behaviour is that if there is no Api version, it should not send the header, which is tested already by the other tests. We could add one specific test if we want to make sure that that is the case.

src/GapicClientTrait.php
```diff
@@ -629,6 +629,13 @@ private function createCallStack(array $callConstructionOptions)
                 'X-Goog-User-Project' => [$quotaProject]
             ];
         }
+        
+        if (isset($this->apiVersion)) {
```
noahdietz, line 633: Should we also check !empty?

bshaffer: Just a reminder that `isset` will return `false` when `$apiVersion` is not defined, not initialized, and when it's equal to `null`. It will return `true` when `$apiVersion` is `false` or an empty string. `!empty` will return `false` in all of these cases.

The only valid difference I could see would be when the proto defines `google.api.api_version` as an empty string:

```
google.api.api_version = ""
```

 In this case, would we want to skip it or include it as an empty header? If the former, then what we have here is correct. If the latter, then we'd want to use `!empty`.

Hectorhammett: Hmmm I can't think of a case where the `apiVersion` is either a `null`. It can be an empty string if you write an empty api version on the proto I guess?

If so, I still think there is value in sending the value itself even if is empty. Imagine someone forgets adding an API version and then they test the client. I think seeing the empty value on the header is more valuable than not sending it at all. But is just my train of thought.

Perhaps you have a specific case in mind? I am still not an expert on the generation side of things 😄

Hectorhammett: > Just a reminder that `isset` will return `false` when `$apiVersion` is not defined, not initialized, and when it's equal to `null`. It will return `true` when `$apiVersion` is `false` or an empty string. `!empty` will return `false` in all of these cases.
> 
> The only valid difference I could see would be when the proto defines `google.api.api_version` as an empty string:
> 
> ```
> google.api.api_version = ""
> ```
> 
> In this case, would we want to skip it or include it as an empty header? If the former, then what we have here is correct. If the latter, then we'd want to use `!empty`.

I wrote an answer before you pressed "comment" haha.

So yeah in my previous comment, I added a point where I see value in sending an empty string. That empty string is not a correct version, but I think there is value in sending whatever is in there for visibility? Don't think that obfuscating that mistake would be helpful.

bshaffer: I agree, I think `isset` in here works as well or better than `empty` 👍

bshaffer (approved):

noahdietz (approved):

PR #554, opened 2024-04-19
feat: Add a custom encoder in Serializer

saranshdhingra:
While instantiating a Serializer we can pass a custom encoder as the last argument.
When present, a proto will be converted into an array using the custom function(encoder) supplied instead of the `encodeMessage` method in the `Serializer` class.

Essentially this provides a way to short-circuit the `encodeMessage` method.


tests/Tests/Unit/SerializerTest.php
```diff
@@ -291,4 +291,50 @@ public function testSpecialEncodingDecodingByFieldType()
...
+            'red' => 100,
+            'blue' => 100
+        ];
+
+        return  [
+            [
+                [
+                    Color::class => function ($message) use($modified) {
+                        return $modified;
+                    }
```
bshaffer, line 325: nit, but can we make this serializer do something with the message and fields? This example is a little too contrived for me. Something like this:

```php
Color::class => function (Color $message) {
    return ['red' => $message->getRed() * 2, 'blue' => $message->getBlue() * 2];
}
```

tests/Tests/Unit/SerializerTest.php
```diff
@@ -291,4 +291,50 @@ public function testSpecialEncodingDecodingByFieldType()
...
+                ],
+                new Color($original),
+                $modified
+            ],
+            [
+                // When no custom encoder is supplied, the encodeMessage will return the data
```
bshaffer, line 331: another nit, but in addition to supplying NO encoder, I'd like to see a test where an encoder is supplied for a _different_ message class, to illustrate that only matching classes are affected.

bshaffer (approved):
two minor requests, otherwise this looks great!

tests/Tests/Unit/SerializerTest.php
saranshdhingra: Yeah, good idea.

Issue #548, opened 2024-03-13
CredentialsWrapper doesn't use authHttpClient when updating meta data


stefanzweifel:
Hi there!

I'm in a Laravel PHP project where I would like to replace the default HttpClient used by the [google-ads SDK](https://github.com/googleads/google-ads-php) and gax-php to be the [Laravel Http Client](https://laravel.com/docs/11.x/http-client).

I noticed that `CredentialsWrapper` doesn't pass `$this->authHttpHandler` to `$this->credentialsFetcher->updateMetaData()` [here](https://github.com/googleapis/gax-php/blob/main/src/CredentialsWrapper.php#L253).

I didn't understand how `CredentialsWrapper` is generated (through protobuf or manually), so I didn't send in a pull request.

Passing `$this->authHttpHandler` to the `updateMetadata` method did the trick for me.

```diff
 // Call updateMetadata to take advantage of self-signed JWTs
 if ($this->credentialsFetcher instanceof UpdateMetadataInterface) {
-    return $this->credentialsFetcher->updateMetadata([], $audience);
+    return $this->credentialsFetcher->updateMetadata([], $audience, $this->authHttpHandler);
 }
```

#### Environment details

  - OS: macOS
  - PHP version: 8.2
  - Package name and version: v1.30.0

#### Steps to reproduce

  1. Create a custom HttpHandler class/closure 
  2. Add the custom HttpHandler to the client options
  3. When making a request to a Google Ads API endpoint, the request to fetch the OAuth Token is done using the default `Guzzle7HttpHandler` instead of the custom HttpHandler.

#### Code example

Here's a short code example using both the Google Ads SDK and gax-php.
The `print_r` statement in the custom HttpHandler will only output the request URL to `https://googleads.googleapis.com/*` but not to `https://oauth2.googleapis.com/token`.

```php
<?php

use Google\Ads\GoogleAds\Lib\OAuth2TokenBuilder;
use Google\Ads\GoogleAds\Lib\V16\GoogleAdsClientBuilder;
use Google\Ads\GoogleAds\V16\Services\Client\GoogleAdsServiceClient;
use Google\Ads\GoogleAds\V16\Services\SearchGoogleAdsRequest;
use GuzzleHttp\Promise\PromiseInterface;
use Psr\Http\Message\RequestInterface;

$OAuth2Credential = (new OAuth2TokenBuilder())
    ->withClientId($clientId)
    ->withClientSecret($clientSecret)
[...]

PR #555, opened 2024-04-22
Fixed CredentialsWrapper not passing authHandler to credentials fetcher

norberttech:
Whole description of this issue is available [here](https://github.com/googleapis/gax-php/issues/548)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -40,8 +40,6 @@ jobs:
                 max_attempts: 3
                 command: composer update ${{ matrix.composerflags }}
             - name: Run script
-              env:
-                GOOGLE_APPLICATION_CREDENTIALS: ${{github.workspace}}/tests/Tests/Unit/testdata/json-key-file.json
               run: vendor/bin/phpunit
 
     style:
--- a/composer.json
+++ b/composer.json
@@ -20,7 +20,9 @@
     "require-dev": {
         "phpunit/phpunit": "^9.6",
         "squizlabs/php_codesniffer": "3.*",
-        "phpspec/prophecy-phpunit": "^2.1"
+        "phpspec/prophecy-phpunit": "^2.1",
+        "ext-protobuf": "*",
+        "ext-grpc": "*"
     },
     "conflict": {
         "ext-protobuf": "<3.7.0"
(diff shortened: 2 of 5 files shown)


composer.json
```diff
@@ -20,7 +20,9 @@
     "require-dev": {
         "phpunit/phpunit": "^9.6",
         "squizlabs/php_codesniffer": "3.*",
-        "phpspec/prophecy-phpunit": "^2.1"
+        "phpspec/prophecy-phpunit": "^2.1",
+        "ext-protobuf": "*",
+        "ext-grpc": "*"
```
norberttech, line 25: without those two it seems to be impossible to run test suite, also it's anyway required by github workflows

.github/workflows/tests.yml
```diff
@@ -40,8 +40,6 @@ jobs:
                 max_attempts: 3
                 command: composer update ${{ matrix.composerflags }}
             - name: Run script
-              env:
-                GOOGLE_APPLICATION_CREDENTIALS: ${{github.workspace}}/tests/Tests/Unit/testdata/json-key-file.json
```
norberttech, line 44: same, without checking workflow it's pretty much impossible to figure out what's wrong so I just moved it to phpunit.xml.dist

norberttech:
> Thanks for your pull request! It looks like this may be your first contribution to a Google open source project. Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).
> 
> View this [failed invocation](https://github.com/googleapis/gax-php/pull/555/checks?check_run_id=24096567273) of the CLA check for more information.
> 
> For the most up to date status, view the checks section at the bottom of the pull request.

lol that's a new level of corporation world bullshit but yeah, here you go: 

image

norberttech:
> I'm not able to understand why it's necessary to add the grpc and protobuf extensions mandatory in the require-dev. I tried to fix this and PR #557 works without any of the configuration changes.

It's not necessary, but without it running locally `vendor/bin/phpunit` is failing, so it should be added to the composer.json required-dev section or the tests that are relaying on those two extensions should be marked as skipped when extensions are not present

composer.json
bshaffer: These test changes are really useful! Can we get them in a separate PR?

norberttech: do you prefer to keep extensions as required-dev or maybe detect if the extension is available and mark tests that rely on it as skipped?

bshaffer: I think the latter provided a better user experience

norberttech: I think so too, will try to propose something later

PR #556, opened 2024-04-22
chore(tests): add phpstan as a dev dependency

norberttech:
For those folks that would like to run phpstan before they open a PR I moved phpstan to lib dev dependencies as it's not creating any conflicts.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -73,6 +73,4 @@ jobs:
           php-version: '8.2'
       - name: Run Script
         run: |
-          composer install
-          composer global require phpstan/phpstan
-          ~/.composer/vendor/bin/phpstan analyse
+          vendor/bin/phpstan analyse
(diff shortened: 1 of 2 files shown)


bshaffer:
The tests are failing with:

```
Invalid workflow file: .github/workflows/tests.yml#L70
The workflow is not valid. .github/workflows/tests.yml (Line: 70, Col: 14): Unexpected value ''
```

See https://github.com/googleapis/gax-php/actions/runs/8788059561

.github/workflows/tests.yml
```diff
@@ -68,9 +68,12 @@ jobs:
       - name: Setup PHP
         uses: shivammathur/setup-php@v2
         with:
```
bshaffer, line 70:
```suggestion
        with:
          php-version: '8.2'
```

bshaffer (approved):

PR #558, opened 2024-04-22
fix: failing tests by moving GOOGLE_APPLICATION_CREDENTIALS to phpunit.xml.dist

norberttech:
The solution mentioned [here](https://github.com/googleapis/gax-php/pull/555#discussion_r1575000105) is already implemented, when I first time tried to execute those tests they were failing not due to missing extensions, but missing env var. 
I moved that env var to phpunit.xml.dist so now the whole testsuite should pass regardless if the extensions are installed or not.


norberttech:
ping @bshaffer

bshaffer (approved):

PR #560, opened 2024-04-23
feat: Add Retry class

ajupazhamayil:
Alternative for `Retry` and `ExponentialBackoff` classes in `Google\Cloud\Core`

### DIFF
--- a/src/Retry.php
+++ b/src/Retry.php
@@ -1,0 +1,85 @@
+<?php
+/*
+ * Copyright 2024 Google LLC
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are
+ * met:
+ *
+ *     * Redistributions of source code must retain the above copyright
+ * notice, this list of conditions and the following disclaimer.
+ *     * Redistributions in binary form must reproduce the above
+ * copyright notice, this list of conditions and the following disclaimer
+ * in the documentation and/or other materials provided with the
+ * distribution.
+ *     * Neither the name of Google Inc. nor the names of its
+ * contributors may be used to endorse or promote products derived from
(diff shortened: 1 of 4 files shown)


src/Retry.php
```diff
@@ -0,0 +1,85 @@
...
+     *
+     * @param callable $call The callable to execute.
+     * @param array $arguments [optional]
+     * @return mixed
+     * @throws \Exception The last exception caught while retrying.
+     */
+    public function execute(callable $call, $arguments = []): mixed
+    {
+        $callStack = fn($method, $args) => call_user_func_array($method, $args);
+        $callStack = new RetryMiddleware($callStack, $this->retrySettings);
```
saranshdhingra, line 66: From what I remember in our discussions, we wanted the RetryMiddleware class to use this new "Retrier" class and not the other way round?

src/Retry.php
```diff
@@ -0,0 +1,85 @@
...
+namespace Google\ApiCore;
+
+use Google\ApiCore\Middleware\RetryMiddleware;
+
+/**
+ * This is a helper class to achieve block level retries.
+ *
+ * @internal
+ */
+class Retry
```
saranshdhingra, line 42: Should the name be "Retrier" ?

src/Retry.php
ajupazhamayil: The [RetryMiddleware](https://github.com/googleapis/gax-php/blob/main/src/Middleware/RetryMiddleware.php) logic only contains how to retry in a middleware way (`RetryMiddleware` is the `nextHandler` in case of any failures). Moving this logic to a `Retrier` class/trait would make the `RetryMiddleware` just a caller to the newly created class. Thats why this implementation proposed.

I am open to change it to the discussed way.

PR #561, opened 2024-04-23
feat: Add Retrier class

ajupazhamayil:


src/Middleware/RetryMiddleware.php
```diff
@@ -44,25 +45,26 @@ class RetryMiddleware implements MiddlewareInterface
...
+     * @param int $deadlineMillis TODO: Deprecate if possible.
+     * @param int $retryAttempts TODO: Deprecate if possible.
      */
-    private int $retryAttempts;
-
     public function __construct(
         callable $nextHandler,
         RetrySettings $retrySettings,
-        $deadlineMs = null,
+        $deadlineMillis = null,
```
bshaffer, line 59: Why change the name? I'd rather not change this unless it's necessary, as with named parameters it's possible that it will break something

bshaffer (commented):
I cannot track down anywhere the `delay` config settings are used to delay the retry function. This is something we can look into adding. I'd call it a bug that it's missing, but since it hasn't been implemented yet it's more like a feature.

src/Middleware/RetryMiddleware.php
ajupazhamayil: I wanted a uniformity between `RetryMiddleware` and `RetrySettings`. If it will break something, I will remove this change. Thank you for the comment.

src/ValidationTrait.php
```diff
@@ -56,6 +56,26 @@ public static function validateNotNull(array $arr, array $requiredKeys)
         return self::validateImpl($arr, $requiredKeys, false);
     }
 
+    /**
+     * @param array $arr Associative array
```
saranshdhingra, line 60: A description of what this function does will be helpful in the docbloc.

src/RetrySettings.php
```diff
@@ -264,37 +281,72 @@ class RetrySettings
...
             'initialRpcTimeoutMillis',
             'rpcTimeoutMultiplier',
             'maxRpcTimeoutMillis',
-            'totalTimeoutMillis',
-            'retryableCodes'
-        ]);
-        $this->initialRetryDelayMillis = $settings['initialRetryDelayMillis'];
-        $this->retryDelayMultiplier = $settings['retryDelayMultiplier'];
+        ];
+        $this->validateAllKeysFromOneOf(
```
saranshdhingra, line 320: Do we mean `self:: validateAllKeysFromOneOf`. I see it's a static function.

src/RetrySettings.php
```diff
@@ -405,6 +461,9 @@ public function with(array $settings)
             'noRetriesRpcTimeoutMillis' => $this->getNoRetriesRpcTimeoutMillis(),
             'maxRetries' => $this->getMaxRetries(),
             'retryFunction' => $this->getRetryFunction(),
+            'retryDelayFunction' => $this->getRetryDelayFunction(),
```
saranshdhingra, line 464: I thought we had the general agreement that we want to remove the delayFunction and only give option to add floats?

src/Retrier.php
```diff
@@ -0,0 +1,172 @@
...
+    /**
+     * Execute the callable with the retry logic.
+     *
+     * @param callable $call
+     * @param array $options
+     *
+     * @return mixed
+     * @throws \Exception
+     */
+    public function execute(callable $call, array $options)
```
saranshdhingra, line 68: This is certainly something new, but do we want to add the ability to change the `$call` or `$options` after a failed attempt.

We had this ability in Core where we allowed changing of headers using `retryListeners`.
Although we don't need this functionality right now, but I just wanted to initiate a discussion on this.

src/RetrySettings.php
ajupazhamayil: What if the delay needs to be retrieved from the server exception? Thats a usecase for the Spanner

src/RetrySettings.php
```diff
@@ -235,25 +233,44 @@ class RetrySettings
...
+    /**
+     * The deadline in milliseconds.
+     */
+    private ?float $deadlineMillis;
+
+    /*
+     * The number of retries that have already been attempted.
+     * The original API call will have $retryAttempts set to 0.
+     */
+    private int $retryAttempts;
```
bshaffer, line 254: This object represents the settings for retrying, or _HOW_ the retrying will be performed.

`$retryAttempts` is related to the actual retrying taking place. This should be handled by the _Retrier_, and NOT the RetrySettings.

src/Middleware/RetryMiddleware.php
```diff
@@ -74,121 +76,91 @@ public function __construct(
...
      * @param Call $call
      * @param array $options
-     * @param string $status
+     * @param ApiException $exception
      *
      * @return PromiseInterface
      * @throws ApiException
      */
-    private function retry(Call $call, array $options, string $status)
+    private function retry(Call $call, array $options, ApiException $exception)
```
bshaffer, line 113: Shouldn't this entire function live in the `Retrier`? That would also remove the need for so many public methods in the Retrier.

bshaffer (commented):
I feel like the design here is slightly off - there's still too much logic happening in `RetryMiddleware` - it should be as thin as possible. Also, having the `$retryAttempts` as part of `RetrySettings` is not what we want (think of `RetrySettings` as an object that should be able to be passed around to different RPC calls without unexpected behavior).

src/Middleware/RetryMiddleware.php
ajupazhamayil: @bshaffer This function uses `GuzzleHttp\Promise\PromiseInterface`. We wanted the `Retrier` to be as generic as possible. Thats why I did not move this functionality to the `Retrier` class.

src/Middleware/RetryMiddleware.php
```diff
@@ -74,121 +76,91 @@ public function __construct(
...
         if (!isset($options['timeoutMillis'])) {
             // default to "noRetriesRpcTimeoutMillis" when retries are disabled, otherwise use "initialRpcTimeoutMillis"
-            if (!$this->retrySettings->retriesEnabled() && $this->retrySettings->getNoRetriesRpcTimeoutMillis() > 0) {
-                $options['timeoutMillis'] = $this->retrySettings->getNoRetriesRpcTimeoutMillis();
-            } elseif ($this->retrySettings->getInitialRpcTimeoutMillis() > 0) {
-                $options['timeoutMillis'] = $this->retrySettings->getInitialRpcTimeoutMillis();
+            if (!$retrySettings->retriesEnabled() && $retrySettings->getNoRetriesRpcTimeoutMillis() > 0) {
+                $options['timeoutMillis'] = $retrySettings->getNoRetriesRpcTimeoutMillis();
+            } elseif ($retrySettings->getInitialRpcTimeoutMillis() > 0) {
+                $options['timeoutMillis'] = $retrySettings->getInitialRpcTimeoutMillis();
```
bshaffer, line 86: After discussing with @saranshdhingra, I think having a `parameterUpdateFunction` as part of either `Retrier` or or `RetrySettings` is the way to handle this - we need to update this on every call, but only for calls made through `RetryMiddleware`, so it should be generic. 

I have a WIP here: https://github.com/googleapis/gax-php/commit/6a940361119b0b5bcd682a729f11d4be50e3d3ba

src/RetrySettings.php
```diff
@@ -235,25 +233,44 @@ class RetrySettings
...
+     * and $retryDelayMultiplier.
+     * The callable will have the following signature:
+     * function (int attempts, Exception $e): int (delay in milliseconds)
+     */
+    private ?Closure $retryDelayFunction;
+
+    /**
+     * The deadline in milliseconds.
+     */
+    private ?float $deadlineMillis;
```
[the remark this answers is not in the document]
ajupazhamayil: I have followed this from RetryMiddleware here: https://github.com/googleapis/gax-php/blob/b4e80d6974781f84532d651f61d10bbb92717fe4/src/Middleware/RetryMiddleware.php#L48

src/RetrySettings.php
```diff
@@ -535,6 +594,48 @@ public function getRetryFunction()
...
     }
 
+    /**
+     * @experimental
+     */
+    public function getRetryDelayMillis(\Exception $exception)
+    {
+        if ($this->retryDelayFunction) {
+            return min(
+                call_user_func($this->retryDelayFunction, $this->retryAttempts, $exception),
```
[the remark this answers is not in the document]
ajupazhamayil: Yes, sometimes the exception contains the delay. See here: https://github.com/googleapis/google-cloud-php/blob/076287b03e801bfaff4f43809de8c4bfc7c9e5d6/Spanner/src/Database.php#L916

src/Retrier.php
```diff
@@ -0,0 +1,172 @@
...
+            );
+        }
+    }
+
+    /**
+     * @param float $currentTimeMillis
+     *
+     * @return float
+     */
+    public function calculateRetryDeadlineMillis(float $currentTimeMillis)
```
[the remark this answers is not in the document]
ajupazhamayil: This made public to use it by the `RetryMiddleware`

src/Retrier.php
```diff
@@ -0,0 +1,172 @@
...
+        return $retryFunction($exception, $options);
+    }
+
+    /**
+     * @param float $currentTimeMillis
+     *
+     * @return void
+     * @throws ApiException
+     */
+    public function checkDeadlineExceeded(float $currentTimeMillis)
```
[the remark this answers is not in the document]
ajupazhamayil: This made public to use it by the `RetryMiddleware`

ajupazhamayil:
We have decided not to go with `gax` `Retrier` logic, Instead we will continue to use the existing `core`'s Retry logic. Hence closing this.

PR #563, opened 2024-04-26
fix: Making backwardsCompatibilityMode an optional boolean

iammichiel:
Making the $backwardsCompatibilityMode an optional `bool`. 

The value of  `$backwardsCompatibilityMode` is not initialized on first call to the `isBackwardsCompatibilityMode()` function. 

Throwing a PHP exception when first called : 
```
Typed property Google\Cloud\Tasks\V2\Gapic\CloudTasksGapicClient::$backwardsCompatibilityMode must not be accessed before initialization
```

### DIFF
--- a/src/GapicClientTrait.php
+++ b/src/GapicClientTrait.php
@@ -80,7 +80,7 @@ trait GapicClientTrait
         Call::CLIENT_STREAMING_CALL => 'startClientStreamingCall',
         Call::SERVER_STREAMING_CALL => 'startServerStreamingCall',
     ];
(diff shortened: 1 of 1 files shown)


bshaffer:
Hi @iammichiel 
Can you verify you're running the latest version of GAX (`v1.32`)? 

```php
require 'vendor/autoload.php';

class TestClient
{
    use Google\ApiCore\GapicClientTrait;

    public function check() 
    {
        return $this->isBackwardsCompatibilityMode();
    }
}


$client = new TestClient();
var_dump($client->check());
```

PR #568, opened 2024-05-16
RFC: Retrier

bshaffer:
This is one way we could handle moving the logic in `RetryMiddleware` to a `Retrier` class, to make it generally available as a callable retrier in other classes.

After discussing with @ajupazhamayil and @saranshdhingra , we are going to continue using `ExponentialBackoff` in `google/cloud-core` instead of moving to a combined approach, as this PR illustrates a lot of complexity that makes this approach less optimal.


bshaffer:
Closing because we decided this option is too complicated

PR #569, opened 2024-05-17
feat: add support for new surface custom operations clients

bshaffer:
See https://github.com/googleapis/gapic-generator-php/pull/712. Adds 3 options to the `OperationsResponse` class:

 - `$getOperationRequest`
 - `$cancelOperationRequest`
 - `$deleteOperationRequest`

These are used to form the request to the V2 Operations client

**Pros**
- Does not use `ReflectionMethod` to determine first argument Request class, which is a hack

**Cons**
- Requires a change to the existing `longRunning` descriptors

**Alternatives Considered**
We could add a `ReflectionMethod` to determine the request class of the Operation methods (see https://github.com/googleapis/gax-php/pull/570)


src/OperationResponse.php
```diff
@@ -495,4 +514,10 @@ private function hasProtoResponse()
     {
         return !is_null($this->lastProtoResponse);
     }
+
+    private function isNewSurfaceOperationsClient(): bool
+    {
+        return !$this->operationsClient instanceof LegacyOperationsClient
+            && false !== strpos(get_class($this->operationsClient), '\\Client\\');
```
noahdietz, line 521: Personally, I would flip this condition so that it reads better instead of being a double negative "not false".
```suggestion
            && true === strpos(get_class($this->operationsClient), '\\Client\\');
```

Altogether it confirms "this client is not instance of Legacy client and has 'Client' namespce"

noahdietz, line 521: Nit: maybe make the magic value `'\\Client\\'` into a private constant `NEW_CLIENT_NAMESPACE` for readability 🤷🏻‍♂️

noahdietz (approved):
Generally LGTM

src/OperationResponse.php
bshaffer: makes sense to me, good call

bshaffer: Good point

bshaffer: Oh, I forgot that we actually have a very good reason for doing this 😛 

The `strpos` function returns `false` to mean the needle string was not found in the haystack string. So we HAVE to use `false !==`. We could potentially do `is_int(strpos(get_class($this->operationsClient), '\\Client\\');` instead, but I'm not sure that's more readable...

noahdietz: Ah ok that makes sense. Good to leave it as is. Thanks!

noahdietz (approved):

PR #570, opened 2024-05-20
draft: Support operations for compute v2 with Reflection

bshaffer:
Alternative implementation to #569 which uses `ReflectionMethod` instead of `Request` class configuration

**Pros**

 - does not require a change to the existing longRunning descriptors
 - fixing LRO for Compute does not require regenerating and re-releasing the compute client

**Cons**

 - Uses `ReflectionMethod` to determine first argument Request class

**Alternatives Considered**
We could add the options getRequestClass, cancelRequestClass, and deleteRequestClass to the descriptors via gapic-generator-php, which would be more efficient (no calls to ReflectionMethod) and better. See #569  and https://github.com/googleapis/gapic-generator-php/pull/712

### DIFF
--- a/src/OperationResponse.php
+++ b/src/OperationResponse.php
@@ -269,10 +269,7 @@ class OperationResponse
         if ($this->operationsClient instanceof OperationsClient) {
             $requestClass = GetOperationRequest::class;
         } elseif ($this->isNewSurfaceOperationsClient()) {
-            $requestClass = (new \ReflectionMethod(
-                $this->operationsClient,
-                $this->getOperationMethod)
-            )->getParameters()[0]->getType()->getName();
+            $requestClass = $this->getRequestClass($this->getOperationMethod);
         }
 
         $this->lastProtoResponse = $this->operationsCall($this->getOperationMethod, $requestClass);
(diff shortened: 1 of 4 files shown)


src/OperationResponse.php
```diff
@@ -456,12 +472,21 @@ public function getMetadata()
...
+     */
+    private function operationsCall(string $method, ?string $requestClass)
     {
-        $firstArgument = $this->operationsClient instanceof OperationsClient
-            ? $requestClass::build($this->getName())
-            : $this->getName();
-        $args = array_merge([$firstArgument], $this->additionalArgs);
+        $args = array_merge([$this->getName()], array_values($this->additionalArgs));
+        if ($requestClass) {
+            $request = call_user_func_array($requestClass . '::build', $args);
```
noahdietz, line 486: This assumes the request class has a builder method generated from the `google.api.method_signature`, right?

bshaffer: Yes it does! And I've confirmed all our existing Request classes have this. However, if something were introduced which didn't support this (or someone supplied the wrong Request class), this would throw a PHP Fatal Error.

it may be a good idea to add an exception here when `build` doesn't exist on the request class, rather than potentially throwing a PHP fatal error.

bshaffer: Done in 66e4a77e (for the other PR)

bshaffer:
Closing because I don't like this option

PR #571, opened 2024-05-21
feat: add support for new surface operations client

bshaffer:
add a new method `createNewSurfaceOperationsClient` (private) while we migrate the previous classes over. 

This is because we accidentally shipped GA "new surface" clients which use the previous surface's operations clients. We don't want to break GA clients, so we can only change the client classes for clients which are `NEW_SURFACE_ONLY` (e.g. they do not contain legacy clients).

### DIFF
--- a/src/GapicClientTrait.php
+++ b/src/GapicClientTrait.php
@@ -32,7 +32,7 @@
 
 namespace Google\ApiCore;
 
-use Google\ApiCore\LongRunning\OperationsClient;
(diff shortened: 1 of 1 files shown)


bshaffer:
I think this direction is not a good idea, and it inflates an already-inflated `GapicClientTrait`.

Better to generate the logic directly in the GAPIC client.

bshaffer:
Closing in favor of https://github.com/googleapis/gapic-generator-php/pull/714

PR #574, opened 2024-07-10
feat: Logging prototype

Hectorhammett:


src/Middleware/LoggerMiddleware.php
```diff
@@ -0,0 +1,76 @@
...
+            'timestamp' => date(DATE_RFC3339),
+            'jsonPayload' => [
+                'serviceName' => $this->serviceName
+            ]
+        ];
+
+        $this->logger->info(json_encode($infoEvent));
+
+        // We could use a Then here to calculate latency if we want as well the retry.
+        return $nextHandler($call, $options);
```
Hectorhammett, line 73: We need to be able to log retries. This automatically logs when we retry from the `RetryMiddleware` but it is not specifically logging the attempt `number`. How could we pass that one along?

src/Transport/GrpcTransport.php
```diff
@@ -254,6 +257,18 @@ function () use ($unaryCall, $options, &$promise) {
...
     }
 
+    /**
+     * Sets a PSR-3 LoggerInterface
+     * 
+     * @param LoggerInterface $logger
+     *
+     * @return void
+     */
+    public function setLogger(LoggerInterface $logger)
```
Hectorhammett, line 267: This is missing on the GRPC fallback transport.

src/Transport/RestTransport.php
```diff
@@ -109,24 +111,34 @@ public static function build(string $apiEndpoint, string $restConfigPath, array
...
+        $requestLog = null;
         $headers = self::buildCommonHeaders($options);
+        $request =  $this->requestBuilder->build(
+            $call->getMethod(),
+            $call->getMessage(),
+            $headers
+        );
+
+        if ($this->logger) {
+            $requestLog = $this->logRequest($request);
```
Hectorhammett, line 123: I am using this to have access to the `timestamp` and calculate the latency. But this is something that we can move to the `LoggerMiddleware` using a `then` inside that middleware.

src/Transport/RestTransport.php
```diff
@@ -264,4 +288,66 @@ private function getCallOptions(array $options)
...
+            'severity' => 'INFO', //Perhaps have something like Logger::debug
+            'jsonPayload' => [
+                'response.status' => $response->getStatusCode(),
+                'latency' => strtotime($startTime) - strtotime($timestamp)
+            ]
+        ];
+
+        $logger->info(json_encode($infoEvent));
+
+        return [$debugEvent, $infoEvent];
```
Hectorhammett, line 351: This is not needed, it was set like this for consistency.

src/Transport/RestTransport.php
```diff
@@ -264,4 +288,66 @@ private function getCallOptions(array $options)
 
         return $callOptions;
     }
+
+    /**
+     * @param RequestInterface $request
+     *
+     * @return array
+     */
+    private function logRequest(RequestInterface $request)
```
bshaffer, line 297: This logic could go in [HttpUnaryTransportTrait](https://github.com/googleapis/gax-php/blob/main/src/Transport/HttpUnaryTransportTrait.php) so that it's also available for [GrpcFallbackTransport](https://github.com/googleapis/gax-php/blob/main/src/Transport/GrpcFallbackTransport.php)

src/Transport/RestTransport.php
```diff
@@ -109,24 +111,34 @@ public static function build(string $apiEndpoint, string $restConfigPath, array
...
         )->then(
-            function (ResponseInterface $response) use ($call, $options) {
+            function (ResponseInterface $response) use ($call, $options, $requestLog) {
                 $decodeType = $call->getDecodeType();
                 /** @var Message $return */
                 $return = new $decodeType;
                 $body = (string) $response->getBody();
 
+                if ($this->logger) {
+                    $this->logResponse($response, $requestLog['timestamp']);
```
bshaffer, line 139: Why not just pass the timestamp in, instead of passing in the whole `$requestLog`? Likewise, why not just have `requestLog` return the timestamp instead of the whole log? Or even better, get the timestamp in `startUnaryCall`, and pass the timestamp into `requestLog` instead of getting it back from that function.

src/Middleware/LoggerMiddleware.php
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    public function __invoke(Call $call, array $options)
+    {
+        $nextHandler = $this->nextHandler;
+
+        $infoEvent = [
+            'timestamp' => date(DATE_RFC3339),
+            'jsonPayload' => [
+                'serviceName' => $this->serviceName
```
bshaffer, line 66: The `$call` has the method name - we should be logging that here as well!

bshaffer (commented):
Still needed:

 - add `logger` to `ClientOptions` 
 - add support for environment variable `GOOGLE_CLOUD_SDK_DEBUG_LOGGING` (per @Hectorhammett this should be in [ClientOptionsTrait::buildClientOptions](https://github.com/googleapis/gax-php/blob/28aa3e95969a75b278606a88448992a6396a119e/src/ClientOptionsTrait.php#L83))
 - gRPC logging

Also I think generally speaking, anything we log in the `LoggerMiddleware` instead of in the Transports themselves, we should, to keep things in one place (and make them transport-agnostic)

src/Transport/RestTransport.php
Hectorhammett: Comment for future reference:
We decided to change the logic for the latency to be calculated on the `loggerMiddleware`

src/Middleware/LoggerMiddleware.php
Hectorhammett: Do we want to log the `methodName` on the structured log? Is not on the docs but I absolutely see the benefit of adding it here 100%.

src/ClientOptionsTrait.php
```diff
@@ -115,6 +118,22 @@ private function buildClientOptions(array $options)
         // Keep track of the API Endpoint
         $apiEndpoint = $options['apiEndpoint'] ?? null;
 
+         // If the user provides a logger or the GOOGLE_SDK_DEBUG_LOGGING environment variable is set
+        // we attach a logger to the options array
+        // If we have the logger option SET to null, it should trump the environment variable AKA explicitely
+        // turning off logging
+        if (getenv('GOOGLE_SDK_DEBUG_LOGGING') ||
```
bshaffer, line 125: This should check the value of `GOOGLE_SDK_DEBUG_LOGGING`. The way it is now, if it was set to `false`, or `bamboozle`, it would still evaluate to true. Better to make this something like

```php
if ('1' === getenv('GOOGLE_SDK_DEBUG_LOGGING')
```

src/ClientOptionsTrait.php
```diff
@@ -115,6 +118,22 @@ private function buildClientOptions(array $options)
...
+        (array_key_exists('logger', $options) && $options['logger'] !== null)) {
+            // If we have the logger set to other than null and is not a LoggerInterface throw and exception.
+            if (isset($options['logger']) && !$options['logger'] instanceof LoggerInterface) {
+                throw new ValidationException(
+                    'the "Logger" option should be PSR-3 LoggerInterface compatible'
```
bshaffer, line 130: the option is lowercase
```suggestion
                    'the "logger" option should be PSR-3 LoggerInterface compatible'
```

src/ClientOptionsTrait.php
```diff
@@ -115,6 +118,22 @@ private function buildClientOptions(array $options)
...
+                );
+            } elseif (getenv('GOOGLE_SDK_DEBUG_LOGGING')) {
+                $options['logger'] = new Logger();
+            }
```
bshaffer, line 134: This logic would be a lot more readable like this

```php
if (!array_key_exists('logger', $options) && '1' === getenv('GOOGLE_SDK_DEBUG_LOGGING')) {
    $options['logger'] = new Logger();
}
// If the logger is supplied,  and is not a LoggerInterface throw and exception.
if (isset($options['logger']) && !$options['logger'] instanceof LoggerInterface) {
        'the "logger" option should be PSR-3 LoggerInterface compatible'
    );
}
```

If you want to keep the `elseif` in there, that's fine too, since we don't _really_ need to perform the `instanceof` check for the default logger... although I don't think it's necessary, and I think this way is more readable.

src/ClientOptionsTrait.php
```diff
@@ -115,6 +118,22 @@ private function buildClientOptions(array $options)
         // Keep track of the API Endpoint
         $apiEndpoint = $options['apiEndpoint'] ?? null;
 
+         // If the user provides a logger or the GOOGLE_SDK_DEBUG_LOGGING environment variable is set
```
bshaffer, line 121:
```suggestion
        // If the user provides a logger or the GOOGLE_SDK_DEBUG_LOGGING environment variable is set
```

src/ClientOptionsTrait.php
```diff
@@ -124,9 +143,19 @@ private function buildClientOptions(array $options)
         if (isset($options['transportConfig']['grpc'])) {
             $options['transportConfig']['grpc'] += $defaultOptions['transportConfig']['grpc'];
             $options['transportConfig']['grpc']['stubOpts'] += $defaultOptions['transportConfig']['grpc']['stubOpts'];
+            $options['transportConfig']['grpc']['logger'] = null;
+            
+            if (isset($options['logger'])) {
+                $options['transportConfig']['grpc']['logger'] = $options['logger'];
+            }
```
bshaffer, line 150: One line instead of 4 is a win.
```suggestion
            $options['transportConfig']['grpc']['logger'] = $options['logger'] ?? null;
```

src/ClientOptionsTrait.php
```diff
@@ -124,9 +143,19 @@ private function buildClientOptions(array $options)
...
         }
         if (isset($options['transportConfig']['rest'])) {
             $options['transportConfig']['rest'] += $defaultOptions['transportConfig']['rest'];
+            $options['transportConfig']['rest']['logger'] = null;
+
+            if (isset($options['logger'])) {
+                $options['transportConfig']['rest']['logger'] = $options['logger'];
+            }
```
bshaffer, line 158:
```suggestion
            $options['transportConfig']['rest']['logger'] = $options['logger'] ?? null;
```

src/GapicClientTrait.php
```diff
@@ -293,6 +296,10 @@ private function setClientOptions(array $options)
             $options['universeDomain']
         );
 
+        if ($options['logger']) {
+            $this->logger = $options['logger'];
+        }
```
bshaffer, line 301:
```suggestion
        $this->logger = $options['logger'] ?? null;
```

src/Middleware/LoggerMiddleware.php
```diff
@@ -0,0 +1,108 @@
...
+        ];
+        
+        // If retry is set and is bigger than 0, we add it to the log.
+        if (isset($options['retryAttempt']) && $options['retryAttempt'] > 0) {
+            $infoEvent['jsonPayload']['retryAttempt'] = $options['retryAttempt'];
+        }
+
+        $this->logger->info(json_encode($infoEvent));
+
+        return $nextHandler($call, $options)->then(function ($response) use ($startTime) {
```
bshaffer, line 78: nit: typehit `$response`

src/Middleware/LoggerMiddleware.php
```diff
@@ -0,0 +1,108 @@
...
+                'severity' => 'DEBUG',
+                'jsonPayload' => [
+                    'latency' => strtotime($endTime) - strtotime($endTime)
+                ]
+            ];
+
+            $this->logger->debug(json_encode($debugEvent));
+
+            return $response;
+        }, function ($exception) use ($startTime) {
```
bshaffer, line 92: nit: typehint `$exception`

src/Transport/GrpcFallbackTransport.php
```diff
@@ -90,7 +95,8 @@ public static function build(string $apiEndpoint, array $config = [])
         ];
         list($baseUri, $port) = self::normalizeServiceAddress($apiEndpoint);
         $httpHandler = $config['httpHandler'] ?: self::buildHttpHandlerAsync();
-        $transport = new GrpcFallbackTransport("$baseUri:$port", $httpHandler);
+        $logger = isset($config['logger']) ? $config['logger'] : null;
```
bshaffer, line 98: Use more succinct ternary/null-coalescing:
```suggestion
        $logger = $config['logger'] ?? null;
```
... or do it inline, as I suggest below

src/Transport/GrpcFallbackTransport.php
```diff
@@ -90,7 +95,8 @@ public static function build(string $apiEndpoint, array $config = [])
...
+        $transport = new GrpcFallbackTransport("$baseUri:$port", $httpHandler, $logger);
```
bshaffer, line 99: consider keeping this simple
```suggestion
        $transport = new GrpcFallbackTransport("$baseUri:$port", $httpHandler, $config['logger'] ?? null);
```

src/Transport/GrpcTransport.php
```diff
@@ -143,8 +153,9 @@ public static function build(string $apiEndpoint, array $config = [])
                 "instead got: " . print_r($channel, true)
             );
         }
+        $logger = isset($config['logger']) ? $config['logger'] : null;
```
bshaffer, line 156: Use more succinct ternary/null-coalescing:
```suggestion
        $logger = $config['logger'] ?? null;
```
... or do it inline, as I suggest below

src/Transport/GrpcTransport.php
```diff
@@ -143,8 +153,9 @@ public static function build(string $apiEndpoint, array $config = [])
...
         try {
-            return new GrpcTransport($host, $stubOpts, $channel, $config['interceptors']);
+            return new GrpcTransport($host, $stubOpts, $channel, $config['interceptors'], $logger);
```
bshaffer, line 158: Consider keeping this simple
```suggestion
            return new GrpcTransport($host, $stubOpts, $channel, $config['interceptors'], $config['logger'] ?? null);
```

src/Transport/GrpcTransport.php
```diff
@@ -224,12 +235,17 @@ public function startServerStreamingCall(Call $call, array $options)
     public function startUnaryCall(Call $call, array $options)
     {
         $this->verifyUniverseDomain($options);
+        $headers = isset($options['headers']) ? $options['headers'] : [];
```
bshaffer, line 238:
```suggestion
        $headers = $options['headers'] ?? [];
```

src/Transport/RestTransport.php
```diff
@@ -88,16 +93,21 @@ public function __construct(
...
         $config += [
             'httpHandler'  => null,
             'clientCertSource' => null,
         ];
         list($baseUri, $port) = self::normalizeServiceAddress($apiEndpoint);
         $requestBuilder = new RequestBuilder("$baseUri:$port", $restConfigPath);
         $httpHandler = $config['httpHandler'] ?: self::buildHttpHandlerAsync();
-        $transport = new RestTransport($requestBuilder, $httpHandler);
+        $logger = isset($config['logger']) ? $config['logger'] : null;
+        $transport = new RestTransport($requestBuilder, $httpHandler, $logger);
```
bshaffer, line 110:
```suggestion
        $transport = new RestTransport($requestBuilder, $httpHandler, $config['logger'] ?? null);
```

src/Transport/GrpcTransport.php
```diff
@@ -283,4 +307,66 @@ private static function loadClientCertSource(callable $clientCertSource)
...
+
+    /**
+     * @param mixed $response
+     * @param mixed $status
+     *
+     * @return void
+     */
+    private function logResponse(mixed $response, mixed $status)
+    {
+        $logger = $this->logger;
```
bshaffer, line 343: Why do it this way ?

src/Transport/HttpUnaryTransportTrait.php
```diff
@@ -167,4 +169,60 @@ private static function loadClientCertSource(callable $clientCertSource)
...
     }
+
+    /**
+     * @param RequestInterface $request
+     *
+     * @return void
+     */
+    private function logRequest(RequestInterface $request)
+    {
+        $logger = $this->logger;
```
bshaffer, line 180: why do it this way?

src/Transport/HttpUnaryTransportTrait.php
```diff
@@ -167,4 +169,60 @@ private static function loadClientCertSource(callable $clientCertSource)
...
+     * @return void
+     */
+    private function logResponse(ResponseInterface $response)
+    {
+        $logger = $this->logger;
+        $timestamp = date(DATE_RFC3339);
+        
+        $debugEvent = [
+            'timestamp' => $timestamp,
+            'severity' => 'DEBUG', //Perhaps have something like Logger::debug
```
bshaffer, line 209: You could use `Psr\Log\LogLevel::DEBUG`, but that's lowercase, so this works for me

src/Transport/GrpcTransport.php
```diff
@@ -283,4 +307,66 @@ private static function loadClientCertSource(callable $clientCertSource)
...
+        $timestamp = date(DATE_RFC3339);
+
+        // In the case we have a $status->code != Code::OK
+        // from the request, we log the status only and we do not have
+        // a response.
```
bshaffer, line 348: I am not crazy about the fact that we log two events, the info AND the debug event, separately. Wouldn't it be better to just log one event, such as:

```php
            $debugEvent = [
                'timestamp' => $timestamp,
                'severity' => 'DEBUG',
                'jsonPayload' => [
                    'response.headers' => $response ? $status->metadata : null,
                    'response.payload' => $response ?? null,
                    'response.status' => $status->code
                ]
            ];
```

The way you have it, one is `INFO` and one is `DEBUG`, so we'd only get to use one, but I assume that's not an important part of the requirements.

src/Transport/GrpcTransport.php
```diff
@@ -283,4 +307,66 @@ private static function loadClientCertSource(callable $clientCertSource)
...
+        if ($response) {
+            $debugEvent = [
+                'timestamp' => $timestamp,
+                'severity' => 'DEBUG',
+                'jsonPayload' => [
+                    'response.headers' => $status->metadata,
+                    'response.payload' => $response,
```
bshaffer, line 355: If the `$response` is of type `mixed`, it makes me nervous just shoving it into `json_encode` like this. Can we not narrow it down more than this? For instance, if it's a `ResponseInterface`, we may want to log `(string) $response->getBody()`.

Additionally, if this is a protobuf `Message` class, then `json_encode` just shows us `{}`:

```php
$message = new Google\Cloud\Logging\V2\Link(); // any message
$message->setName('Link Name');
var_dump(json_encode($message));
// string(2) "{}"
```

But if we use `serializeToJsonString`, we get to see the data:

```php
$message = new Google\Cloud\Logging\V2\Link(); // any message
$message->setName('Link Name');
var_dump(json_encode($message->serializeToJsonString()));
// string(16) "{"name":"brent"}"
```

So, I think we want to do more handling with this object than just putting everything in `json_encode`

bshaffer (commented):
This looks great! I have a handful of important suggestions / observations for the `logResponse` method, but otherwise my suggestions are mostly just PHP code style / best practice suggestions.

src/ClientOptionsTrait.php
Hectorhammett: Ahhh is going to be interpreted as a `string`! You're right, let me fix that.

src/ClientOptionsTrait.php
Hectorhammett: But what happens with you set the option `logger` to `nuil`. That would throw a `ValidationException`. This is still in the works right? But in the document I proposed that if you pass a `null` you are specifically disabling the logger for this client.

This is done for the following example in mind:
What if you have 3 different clients and two of them are having issues and you want to not log on one?
* You could enable the flag and pass the `null` option to the client you know is fine
* You have the flag disabled and pass loggers to the clients you want

Maybe we could instead of null pass a `false`? Or maybe another mechanism? As this is absolutely still in the design phase. We could have this discussion in the document :)

src/Transport/GrpcFallbackTransport.php
Hectorhammett: If `logger` is not set, that would raise a warning, wouldn't it? Wouldn't crash, but it would throw a warning.

src/Transport/GrpcTransport.php
Hectorhammett: No specific reason. This can be easily changed to `$this->logger-><level>()` below.

src/Transport/GrpcTransport.php
Hectorhammett: I get what you mean and I agree.

But the current requirements for the project states that `status` is an `info` level log:
image

bshaffer: This is maybe something you should confirm with the group then (or maybe you already have)... that we want a specific logging event just for `response.status`. If so, then LGTM

src/Transport/GrpcTransport.php
bshaffer: Seems strange to me to set a local variable that we don't need

src/ClientOptionsTrait.php
bshaffer: > But what happens with you set the option logger to nuil. That would throw a ValidationException. This is still in the works right?

No, because in that case, `isset($options['logger'])` would return `false`



```

File: src/Transport/GrpcTransport.php (hunk 26 above)
Comment (bshaffer) [reply src/Transport/GrpcTransport.php:348]: one issue with having multiple events is when we have multiple requests showing up, we can't be sure of which `response.status` is connected to which `response.header` and `response.payload`. We need to either log these things in the same event, or have a trace ID which ties them together.

File: src/ClientOptionsTrait.php (hunk 10 above)
Comment (Hectorhammett) [reply src/ClientOptionsTrait.php:134]: You are correct. The behaviour I was thinking of `isset` is the one from` array_key_exists`.

File: src/Transport/HttpUnaryTransportTrait.php (hunk 25 above)
Comment (Hectorhammett) [reply src/Transport/HttpUnaryTransportTrait.php:209]: Yeah I also thought of it, but the lower case is what made me doubt this.

File: src/Transport/GrpcTransport.php (hunk 26 above)
Comment (Hectorhammett) [reply src/Transport/GrpcTransport.php:348]: There is a `traceId` field but it is meant for open telemetry. After yesterday's meeting we stablished that we can in fact switch logs levels to different levels.

File: src/Transport/GrpcTransport.php (hunk 27 above)
Comment (bshaffer) [reply src/Transport/GrpcTransport.php:352]: I think `Message` would be appropriate, then. Looking at that code, it assumes a `mergeFromString` function exists. So it COULD be a generic object (e.g. duck typing), but I think we can assume it's a `Message`.  

From the [link you provided](https://github.com/grpc/grpc/blob/master/src/php/lib/Grpc/AbstractCall.php#L127), at the very least it must be an `object` or `null`. So I'd recommend doing something like this (to be safe): 

```php
private function logResponse(object|null $response, mixed $status)
{
    // .. 
    if ($response) {
        $responseToSerialize = $response instanceof Message
            ? $response->serializeToJsonString()
            : get_class($response);
    // ..
```

That way in some case where it isn't a `Message` (which we don't expect, but which is technically possible), we can at least handle that case gracefully.

src/Logging/Logger.php
```diff
@@ -0,0 +1,118 @@
...
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+namespace Google\ApiCore;
```
Hectorhammett, line 33: In the new design, this class will live inside the `AUTH` library

src/ClientOptionsTrait.php
```diff
@@ -115,6 +119,20 @@ private function buildClientOptions(array $options)
         // Keep track of the API Endpoint
         $apiEndpoint = $options['apiEndpoint'] ?? null;
 
+        // If the user provides a logger or the GOOGLE_SDK_DEBUG_LOGGING environment variable is set
+        // we attach a logger to the options array
+        // If we have the logger option SET to null, it should trump the environment variable AKA explicitely
+        // turning off logging
+        if (!array_key_exists('logger', $options)) {
+            $options['logger'] = ApplicationDefaultCredentials::getDefaultLogger();
```
Hectorhammett, line 127: This is not working in the prototype as this depends on the Auth library changes

composer.json
```diff
@@ -15,7 +15,8 @@
         "guzzlehttp/psr7": "^2.0",
         "google/common-protos": "^4.4",
         "google/longrunning": "~0.4",
-        "ramsey/uuid": "^4.0"
+        "ramsey/uuid": "^4.0",
+        "psr/log": "^3.0"
```
bshaffer, line 19: do the following to make this work in the tests here

```
composer require "google/auth:dev-logger-prototype as 1.43.0" 
```

src/ClientOptionsTrait.php
```diff
@@ -115,6 +119,20 @@ private function buildClientOptions(array $options)
         // Keep track of the API Endpoint
         $apiEndpoint = $options['apiEndpoint'] ?? null;
 
+        // If the user provides a logger or the GOOGLE_SDK_DEBUG_LOGGING environment variable is set
+        // we attach a logger to the options array
+        // If we have the logger option SET to null, it should trump the environment variable AKA explicitely
```
bshaffer, line 124: this looks good, but in the PHPDoc for `ClientOptions` (and here) we should reword this to be slightly more clear, something like
```
setting the "logging" option to `null` will explicitly turn logging off (e.g. a default logger will not be fetched from the environment)
```

src/Transport/GrpcFallbackTransport.php
```diff
@@ -103,15 +103,18 @@ public static function build(string $apiEndpoint, array $config = [])
     public function startUnaryCall(Call $call, array $options)
     {
         $httpHandler = $this->httpHandler;
+        $request = $this->buildRequest($call, $options);
```
bshaffer, line 106: nit: this probably wasn't necessary

src/Transport/RestTransport.php
```diff
@@ -262,6 +262,10 @@ private function getCallOptions(array $options)
             $callOptions['key'] = $key;
         }
 
+        if (isset($options['retryAttempt'])) {
+            $callOptions['retryAttempt'] = $options['retryAttempt'];
+        }
```
bshaffer, line 267: Should this be added to `GrpcFallbackTransport::getCallOptions` as well?

PR #580, opened 2024-08-30
feat: Add support for  to use a  object with a union type

Hectorhammett:


src/Options/ClientOptions.php
```diff
@@ -61,37 +61,37 @@ class ClientOptions implements ArrayAccess
 {
     use OptionsTrait;
 
-    private ?string $apiEndpoint;
+    public ?string $apiEndpoint;
```
Hectorhammett, line 64: This is a highly opinionated change. My believe is that the options objects should be mostly structural and have easy of use for the user and the developers.

If you are in need or wanting to use this object, using an IDE and start typing `$options->` should give a quick auto completion that helps the user understand the different options. We should document these values if we decided to continue with this philosophy.

Due this being highly opinionated though, I am willing to return this to `private`.

src/ClientOptionsTrait.php
```diff
@@ -80,8 +81,20 @@ private static function getClientDefaults()
...
 
-    private function buildClientOptions(array $options)
+    private function buildClientOptions(array|ClientOptions $options)
     {
+        if ($options instanceof ClientOptions) {
+            $options = $options->toArray();
+
+            // The constructor of ClientOptions construct creates a `TransportOptions`.
+            // We call `toArray` on it and nested items that uses the `optionsTrait` to keep
+            // the same behaviour as if it was just an array.
```
bshaffer, line 91: We could also make `toArray` call `toArray` on those items.... it would be a breaking change, so I think it's a bad idea. It's a shame we didn't do that before! Maybe in a 2.0 version...

src/Options/ClientOptions.php
bshaffer: The only concern I have here is that if there's some logic being done in the setters (which is the case for a few of these), it would be lost in setting these manually, which may lead to some confusing behavior.

bshaffer (commented):
One thing I think we should look into is how to make our reference documentation look better 

The first 5 methods in https://googleapis.github.io/gax-php/main/Google/ApiCore/Options/ClientOptions.html are not helpful, the top code example does not really show how to use the class, and the first method that is useful ([__construct](https://googleapis.github.io/gax-php/main/Google/ApiCore/Options/ClientOptions.html#method___construct)) is a completely unformatted mess.

I'm not sure what impact it would have making the options public, and if that would help or hurt, but we should look into that also.

bshaffer:
One thought on this - if we are unhappy with the `ClientOptions` interface, as we haven't exposed this class directly for use (nothing in our libraries accept OR return this class), if we want to make a change to it, now's the time.

I would suggest, just for safety, that we mark any changing methods as `@deprecated` for a month or so before doing this, just to be safe. But if we want to turn these into POPOs (which I think is a great idea), now's the time.

src/Options/ClientOptions.php
bshaffer: I think if we DID do this, we should rewrite the class and just make it a POPO.

bshaffer: After sitting on this for almost a year, I don't think we should make the properties public. My reasoning here is that
 1. you can access them already with array access (e.g. `$clientOptions['apiEndpoint']`)
 2. they do work in the setters (e.g. `setCredentialsFile` loads the file contents)

tests/Tests/Unit/ClientOptionsTraitTest.php
```diff
@@ -545,6 +546,16 @@ public function testBuildClientOptionsTwice()
...
     }
 
+    public function testBuildClientOptionsWithClientOptions()
+    {
+        $client = new StubClientOptionsClient();
+        $clientOptions = new ClientOptions([]);
+        $clientOptions->setApiEndpoint('TestEndpoint.com');
+        $builtOptions = $client->buildClientOptions($clientOptions);
+
+        $this->assertEquals($clientOptions->apiEndpoint, $builtOptions['apiEndpoint']);
```
bshaffer, line 556:
```suggestion
        $this->assertEquals($clientOptions['apiEndpoint'], $builtOptions['apiEndpoint']);
```

Hectorhammett: I know that we are trying to keep things compatible and such, but it breaks my heard to use array notations vs `$variable->property` </3

PR #582, opened 2024-09-03
Add `InsecureRequestBuilder` for emulator

pulzarraider:
fixes https://github.com/googleapis/google-cloud-php/issues/7187

[According to documentation](https://cloud.google.com/pubsub/docs/emulator), the Pub/Sub emulator only supports unsecure (HTTP) connections.  After updating `google/cloud-pubsub` to version 2 there was no option to use the emulator with REST transport  (without gRPC PHP extension). 
While the fix in https://github.com/googleapis/google-cloud-php/pull/7588 provided some assistance, it still required a significant amount of code to modify the client to generate URLs with the `http` protocol.

This merge request introduces `InsecureRequestBuilder` which automatically generates URLs with the http scheme, simplifying the process.

### DIFF
--- a/src/GapicClientTrait.php
+++ b/src/GapicClientTrait.php
@@ -218,6 +218,8 @@ trait GapicClientTrait
      */
     private function setClientOptions(array $options)
     {
+        $hasEmulator = $options['hasEmulator'] ?? false;
+
         // serviceAddress is now deprecated and acts as an alias for apiEndpoint
         if (isset($options['serviceAddress'])) {
             $options['apiEndpoint'] = $this->pluck('serviceAddress', $options, false);
@@ -300,7 +302,8 @@ trait GapicClientTrait
                 $options['apiEndpoint'],
                 $transport,
                 $options['transportConfig'],
-                $options['clientCertSource']
+                $options['clientCertSource'],
+                $hasEmulator
             );
     }
 
@@ -309,6 +312,7 @@ trait GapicClientTrait
      * @param string $transport
      * @param TransportOptions|array $transportConfig
      * @param callable $clientCertSource
+     * @param bool $hasEmulator
      * @return TransportInterface
      * @throws ValidationException
      */
@@ -316,7 +320,8 @@ trait GapicClientTrait
         string $apiEndpoint,
         $transport,
         $transportConfig,
-        callable $clientCertSource = null
+        callable $clientCertSource = null,
+        bool $hasEmulator = false
     ) {
         if (!is_string($transport)) {
             throw new ValidationException(
@@ -359,6 +364,8 @@ trait GapicClientTrait
                     );
                 }
                 $restConfigPath = $configForSpecifiedTransport['restClientConfigPath'];
+                $configForSpecifiedTransport['hasEmulator'] = $hasEmulator;
+
                 return RestTransport::build($apiEndpoint, $restConfigPath, $configForSpecifiedTransport);
             default:
                 throw new ValidationException(
--- a/src/InsecureRequestBuilder.php
+++ b/src/InsecureRequestBuilder.php
@@ -1,0 +1,45 @@
+<?php
+
+namespace Google\ApiCore;
+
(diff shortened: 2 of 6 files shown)


Dnd-Mafer:
This is exactly what I'm looking for !
Any update on when it will be merged ?

bshaffer:
@Dnd-Mafer @pulzarraider thanks for your contributions! I'm taking a look at this. This looks great, but I want to be sure we've looked at other ways this could be achieved as well (e.g. adding an option like `scheme` or `isSecure` to `CredentialsWrapper`) before merging. I'll try to take a look soon.

pulzarraider:
@bshaffer is there’s anything I can clarify or adjust to help this merge request move forward?

bshaffer:
@pulzarraider you can see the tests are failing. Those will need to be fixed before we can merge this.

src/GapicClientTrait.php
```diff
@@ -300,7 +302,8 @@ private function setClientOptions(array $options)
                 $options['apiEndpoint'],
                 $transport,
                 $options['transportConfig'],
-                $options['clientCertSource']
+                $options['clientCertSource'],
+                $hasEmulator
```
bshaffer, line 306: nit: just pass the above assertion here, no need for a local variable:
```suggestion
                $options['hasEmulator'] ?? false
```

src/InsecureRequestBuilder.php
```diff
@@ -0,0 +1,45 @@
...
+    /**
+     * @param string $baseUri
+     * @param string $restConfigPath
+     * @throws ValidationException
+     */
+    public function __construct(string $baseUri, string $restConfigPath)
+    {
+        parent::__construct($baseUri, $restConfigPath);
+    }
+
```
bshaffer, line 22: this is the same as the parent, so why define it?
```suggestion
```

src/InsecureRequestBuilder.php
```diff
@@ -0,0 +1,45 @@
...
+     */
+    protected function buildUri(string $path, array $queryParams)
+    {
+        $uri = Utils::uriFor(
+            sprintf(
+                'http://%s%s',
+                $this->baseUri,
+                $path
+            )
+        );
```
bshaffer, line 36: nit: simplify this
```suggestion
        $uri = Utils::uriFor(sprintf(
            'http://%s%s',
            $this->baseUri,
            $path
        ));
```

bshaffer (commented):
This looks great! I just have a few style nits (suggestions), but as soon as this passes we can merge it.

Thanks for your contribution

.github/workflows/docs.yml
```diff
@@ -19,5 +19,5 @@ jobs:
         uses: GoogleCloudPlatform/php-tools/.github/workflows/doctum.yml@main
         with:
           title: "Google API Core (GAX) Reference Documentation"
-          default_version: ${{ inputs.tag || github.head_ref || github.ref_name }}
+          default_version: ${{ inputs.tag || github.head_ref || github.event.pull_request.head.repo.full_name }}
```
bshaffer, line 22:
```suggestion
          default_version: ${{ inputs.tag || github.head_ref || github.ref_name }}
```

src/GapicClientTrait.php
```diff
@@ -218,6 +218,8 @@ protected function getCredentialsWrapper()
      */
     private function setClientOptions(array $options)
     {
+        $hasEmulator = $options['hasEmulator'] ?? false;
+
```
bshaffer, line 222:
```suggestion
```

bshaffer (approved):

pulzarraider:
Thank you @bshaffer

Dnd-Mafer:
❤️

bshaffer:
Looks like there's a bug that needs fixing here

https://github.com/googleapis/gax-php/commit/8d03965a341a481f57cd33abe42eefaddb245d4e#r148984939

PR #585, opened 2024-09-11
feat: add logging to the supported transports

Hectorhammett:
related: https://github.com/googleapis/google-auth-library-php/pull/578

Adds logging capabilities to the Gapic Clients by adding logging into the supported transports (Http, Grpc and GrpcFallback).


composer.json
```diff
@@ -7,7 +7,7 @@
     "license": "BSD-3-Clause",
     "require": {
         "php": "^8.0",
-        "google/auth": "^1.34.0",
+        "google/auth": "dev-http-logging as 1.3",
```
Hectorhammett, line 10: This needs to be reverted after the review.

bshaffer (changes requested):
A few things I've observed when debugging this:

1. As we are logging the response code separately, it should go _before_ the rest of the response log
2. We do NOT want to log all the client confiugration we are currently logging (see [this example](https://gist.githubusercontent.com/bshaffer/1d8434654aa7322f9e6347c1c03a9f3b/raw/eb27f92b2897e38fca954a64e2b4e09203058554/gistfile1.txt)). We should only log the configuration supplied by the developer
3. The order of the logs ideally would match the order of the requests being made. Right now, the request to the services is logged first, followed by the auth request. In reality, the auth request is made first. If this is really hard to implement, the way we have it is okay, but it's worth considering.

src/BidiStream.php
```diff
@@ -69,6 +70,7 @@ public function __construct(
             $this->resourcesGetMethod = $streamingDescriptor['resourcesGetMethod'];
         }
         $this->logger = $logger;
+        $this->identifier = spl_object_id($this) . rand(1000,9999);
```
bshaffer, line 73: using the `.` will convert both ints to a string. I think that's okay, but we'll need to change the type of `$identifier` to be `string`

Hectorhammett: This has been updated 🙏 We are calculating the `requestId` different now.

src/ClientOptionsTrait.php
```diff
@@ -317,4 +351,25 @@ private function isBackwardsCompatibilityMode(): bool
...
+        if (!$logger) {
+            return;
+        }
+
+        $configurationLog = [
+            'timestamp' => date(DATE_RFC3339),
+            'severity' => strtoupper(LogLevel::DEBUG),
+            'jsonPayload' => [
+                'suppliedConfiguration' => $options
+            ]
```
bshaffer, line 370: Let's add `processId` here as well, so it can be at least loosely tied to the request/response logs

src/ClientOptionsTrait.php
```diff
@@ -317,4 +351,26 @@ private function isBackwardsCompatibilityMode(): bool
...
+        if (!$logger) {
+            return;
+        }
+
+        $configurationLog = [
+            'timestamp' => date(DATE_RFC3339),
+            'severity' => strtoupper(LogLevel::DEBUG),
+            'processId' => getmypid(),
+            'jsonPayload' => [
+                'suppliedConfiguration' => $options
```
bshaffer, line 370: Can we make this something else? Such as `clientConfiguration`? as it is now, we don't know what configuration is being supplied. If you want to be super explicit, we could make it `userSuppliedClientConfiguration`.

Another issue is that there's no indication of what client this is. Would it make sense to have the classname or API endpoint in there as well? Something like:

```php
'jsonPayload' => [
    'clientName' => $this->serviceName, // could also be "get_class($this)"
    'clientConfiguration' => $options,
]
```

src/BidiStream.php
```diff
@@ -144,6 +168,22 @@ public function read()
                 throw ApiException::createFromStdClass($status);
             }
         }
+
+        if ($this->logger) {
+            $responseEvent = new RpcLogEvent();
+
+            $responseEvent->headers = (is_null($result)) ? $this->call->getMetadata() : null;
+            $responseEvent->status = (is_null($result)) ? $status->code : null;
```
bshaffer, line 176:
```suggestion
            $responseEvent->headers = $this->call->getMetadata();
            $responseEvent->status = $status->code ?? null;
```

src/BidiStream.php
```diff
@@ -144,6 +168,22 @@ public function read()
...
+            $responseEvent->processId = (int) getmypid();
+            $responseEvent->requestId = crc32((string) spl_object_id($this) . getmypid());
+
+            if ($result && $result instanceof Message) {
```
bshaffer, line 180: not necessary to check if `$result` exists first - if result is null this will be false already
```suggestion
            if ($result instanceof Message) {
```

src/ClientOptionsTrait.php
```diff
@@ -115,18 +120,45 @@ private function buildClientOptions(array $options)
...
+
         // Merge defaults into $options starting from top level
         // variables, then going into deeper nesting, so that
         // we will not encounter missing keys
         $options += $defaultOptions;
+
+        // If logger is explicitly set to false, logging is disabled
+        if ($options['logger'] !== false) {
+            $options['logger'] = $options['logger'] ?? ApplicationDefaultCredentials::getDefaultLogger();
+        }
```
bshaffer, line 134:
```suggestion
        $options['logger'] ??= ApplicationDefaultCredentials::getDefaultLogger();
```

src/ClientOptionsTrait.php
```diff
@@ -115,18 +120,45 @@ private function buildClientOptions(array $options)
...
+
+        if ($options['logger'] !== null && $options['logger'] !== false && !$options['logger'] instanceof LoggerInterface) {
+            throw new ValidationException(
+                'The "logger" option in the options array should be PSR-3 LoggerInterface compatible'
+            );
+        }
```
bshaffer, line 140:
```suggestion
        if (
            $options['logger'] !== null
            && $options['logger'] !== false
            && !$options['logger'] instanceof LoggerInterface
        ) {
            throw new ValidationException(
                'The "logger" option in the options array should be PSR-3 LoggerInterface compatible'
            );
        }
```

src/ClientOptionsTrait.php
```diff
@@ -115,18 +120,45 @@ private function buildClientOptions(array $options)
...
+
+        // Log the user supplied configuration.
+        $this->logConfiguration($options['logger'], $clientSuppliedOptions);
+
+        if (isset($options['logger'])) {
+            $options['credentialsConfig']['authHttpHandler'] = HttpHandlerFactory::build(logger: $options['logger']);
```
bshaffer, line 146:
```suggestion
            $options['credentialsConfig']['authHttpHandler'] = HttpHandlerFactory::build(
                logger: $options['logger']
            );
```

src/ClientStream.php
```diff
@@ -61,6 +71,16 @@ public function __construct(// @phpstan-ignore-line
      */
     public function write($request)
     {
+        if ($this->logger && $request instanceof Message) {
```
bshaffer, line 74: add a comment to explain this (or log the string `$request`)
```suggestion
        // In some cases, $request can be a string
        if ($this->logger && $request instanceof Message) {
```

src/ClientStream.php
```diff
@@ -74,6 +94,21 @@ public function readResponse()
...
         if ($status->code == Code::OK) {
+            if ($this->logger) {
+                $responseEvent = new RpcLogEvent();
+
+                $responseEvent->headers = $status->metadata;
+                $responseEvent->status = $status->code;
+                $responseEvent->processId = (int) getmypid();
+                $responseEvent->requestId = crc32((string) spl_object_id($this) . getmypid());
+
+                if ($response && $response instanceof Message) {
```
bshaffer, line 105:
```suggestion
                if ($response instanceof Message) {
```

src/ServerStream.php
```diff
@@ -31,28 +31,40 @@
...
      * @param ServerStreamingCallInterface $serverStreamingCall The server streaming call object
      * @param array $streamingDescriptor
+     * @param null|LoggerInterface $logger A PSR-3 compliant logger.
      */
-    public function __construct($serverStreamingCall, array $streamingDescriptor = [])
-    {
+    public function __construct(
+        $serverStreamingCall,
+        array $streamingDescriptor = [],
+        null|LoggerInterface $logger = null
```
bshaffer, line 61: can take advantage of declaring properties in php
```suggestion
        private null|LoggerInterface $logger = null
```

src/Transport/GrpcFallbackTransport.php
```diff
@@ -89,7 +89,7 @@ public static function build(string $apiEndpoint, array $config = [])
             'clientCertSource' => null,
         ];
```
bshaffer, line 90:
```suggestion
            'logger' => null,
        ];
```

src/Transport/GrpcFallbackTransport.php
```diff
@@ -89,7 +89,7 @@ public static function build(string $apiEndpoint, array $config = [])
...
         list($baseUri, $port) = self::normalizeServiceAddress($apiEndpoint);
-        $httpHandler = $config['httpHandler'] ?: self::buildHttpHandlerAsync();
+        $httpHandler = $config['httpHandler'] ?: self::buildHttpHandlerAsync(logger: $config['logger'] ?? null);
```
bshaffer, line 92:
```suggestion
        $httpHandler = $config['httpHandler'] ?: self::buildHttpHandlerAsync(logger: $config['logger']);
```

src/Transport/GrpcTransport.php
```diff
@@ -144,7 +157,10 @@ public static function build(string $apiEndpoint, array $config = [])
             );
         }
         try {
-            return new GrpcTransport($host, $stubOpts, $channel, $config['interceptors']);
+            if (isset($config['logger']) && $config['logger'] === false) {
+                $config['logger'] = null;
+            }
+            return new GrpcTransport($host, $stubOpts, $channel, $config['interceptors'], $config['logger'] ?? null);
```
bshaffer, line 163: Add `'logger' => null` to the `$config +=` array above, and then this can be simpler.
```suggestion
            if ($config['logger'] === false) {
                $config['logger'] = null;
            }
            return new GrpcTransport($host, $stubOpts, $channel, $config['interceptors'], $config['logger']);
```

src/Transport/RestTransport.php
```diff
@@ -100,7 +100,7 @@ public static function build(string $apiEndpoint, string $restConfigPath, array
         $requestBuilder = $config['hasEmulator']
             ? new InsecureRequestBuilder("$baseUri:$port", $restConfigPath)
             : new RequestBuilder("$baseUri:$port", $restConfigPath);
-        $httpHandler = $config['httpHandler'] ?: self::buildHttpHandlerAsync();
+        $httpHandler = $config['httpHandler'] ?: self::buildHttpHandlerAsync($config['logger'] ?? null);
```
bshaffer, line 103: add `logger => null` to the `$config` array above
```suggestion
        $httpHandler = $config['httpHandler'] ?: self::buildHttpHandlerAsync($config['logger']);
```

src/Transport/HttpUnaryTransportTrait.php
```diff
@@ -126,10 +127,10 @@ private static function buildCommonHeaders(array $options)
      * @return callable
      * @throws ValidationException
      */
-    private static function buildHttpHandlerAsync()
+    private static function buildHttpHandlerAsync(null|false|LoggerInterface $logger = null)
     {
         try {
-            return [HttpHandlerFactory::build(), 'async'];
+            return [HttpHandlerFactory::build(null, $logger), 'async'];
```
bshaffer, line 133:
```suggestion
            return [HttpHandlerFactory::build(logger: $logger), 'async'];
```

tests/Unit/Middleware/RetryMiddlewareTest.php
```diff
@@ -489,3 +529,30 @@ public function testUnlimitedMaxRetries()
         $middleware($call->reveal(), [])->wait();
     }
 }
+
+class MockHandler implements MiddlewareInterface
```
bshaffer, line 533: Can this be removed?

tests/Unit/Middleware/RetryMiddlewareTest.php
```diff
@@ -384,6 +385,45 @@ public function testMaxRetries()
         $middleware($call->reveal(), [])->wait();
     }
 
+    public function testRetriesAreIncludedInTheOptionsArray()
+    {
+        $call = $this->getMockBuilder(Call::class)
+            ->disableOriginalConstructor()
+            ->getMock();
```
bshaffer, line 392: can we make this prophecy instead?
```suggestion
        $call = $this->prophecy(Call::class)->reveal();
```

tests/Unit/Transport/GrpcTransportTest.php
```diff
@@ -436,6 +439,17 @@ public function testClientCertSourceOptionInvalid()
...
+    // {
+    //     // $logger = $this->prophesize(StdOutLogger::class);
+    //     // $logger->debug(Argument::cetera())
+    //     //     ->shouldBeCalledTimes(2);
+    //     // $logger->info(Argument::cetera())
+    //     //     ->shouldBeCalledTimes(1);
+    //     // $logger->log(Argument::cetera())
+    //     //     ->shouldBeCalled();
+    // }
+
```
bshaffer, line 452: remove!
```suggestion
```

bshaffer (approved):

PR #593, opened 2024-11-12
feat: Add emulator on `ClientOptions`

ubrize:
This pull request introduces the `hasEmulator` property to the `ClientOptions` class, finalizing support for the Pub/Sub emulator.

Previously, the `InsecureRequestBuilder` (introduced recently in [#582](https://github.com/googleapis/gax-php/pull/582)) was ignored when using Pub/Sub V2 because the `$options` array in `GapicClientTrait` is transformed into a `ClientOptions` object ([line 253](https://github.com/googleapis/gax-php/blob/main/src/GapicClientTrait.php#L253)), causing the `hasEmulator` value to be discarded. As a result, `hasEmulator` always defaulted to `false` ([line 317](https://github.com/googleapis/gax-php/blob/main/src/GapicClientTrait.php#L317)).

### DIFF
--- a/src/Options/ClientOptions.php
+++ b/src/Options/ClientOptions.php
@@ -96,6 +96,8 @@ class ClientOptions implements ArrayAccess
 
     private ?string $apiKey;
 
+    private bool $hasEmulator;
+
     /**
      * @param array $options {
      *     @type string $apiEndpoint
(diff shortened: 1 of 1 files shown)


bshaffer:
Thank you for your contribution! 

I don't think this is the way we want to support this, especially since only a select few of our client libraries support emulators. 

We can potentially revert the change made [here](https://github.com/googleapis/gax-php/commit/8d03965a341a481f57cd33abe42eefaddb245d4e#r148984939), or add this logic to [EmulatorTrait](https://github.com/googleapis/google-cloud-php/blob/main/Core/src/EmulatorTrait.php), which gets added to GAPIC clients which support emulators.

bshaffer:
After thinking about this more, I feel like the intended design is to use environment variables, rather than a `useEmulator` client option. Therefore I think we should put the logic to see if the parameter exists before casting to `ClientOptions` in `GapicClientTrait`. I'll create a PR shortly!

bshaffer:
Done in https://github.com/googleapis/gax-php/pull/594! If you get a chance, please take a look and verify that this resolves your issues.

ubrize:
Thank you for the update. I can confirm that https://github.com/googleapis/gax-php/pull/594 solves the issue.

PR #594, opened 2024-11-14
fix: ensure hasEmulator client option is passed to createTransport

bshaffer:
fixes bad edit made in  https://github.com/googleapis/gax-php/commit/8d03965a341a481f57cd33abe42eefaddb245d4e#r148984939 as part of https://github.com/googleapis/gax-php/pull/582

alternative implementation of https://github.com/googleapis/gax-php/pull/593, but does not add `hasEmulator` to `ClientOptions` (because we don't want to expose `hasEmulator` as a client option)

### DIFF
--- a/src/GapicClientTrait.php
+++ b/src/GapicClientTrait.php
@@ -243,7 +243,7 @@ trait GapicClientTrait
         // internally to determine if the client is running in emulator mode.
         // Therefore, we need to remove it from the $options array before
         // creating the ClientOptions.
-        $hasEmulator = $this->pluck('hasEmulator', $options, false);
+        $hasEmulator = $this->pluck('hasEmulator', $options, false) ?: false;
         if ($this->isBackwardsCompatibilityMode()) {
             if (is_string($options['clientConfig'])) {
                 // perform validation for V1 surfaces which is done in the
--- a/src/GapicClientTrait.php
+++ b/src/GapicClientTrait.php
@@ -243,7 +243,7 @@ trait GapicClientTrait
         // internally to determine if the client is running in emulator mode.
         // Therefore, we need to remove it from the $options array before
         // creating the ClientOptions.
-        $hasEmulator = $this->pluck('hasEmulator', $options, false) ?: false;
+        $hasEmulator = $this->pluck('hasEmulator', $options, false) ?? false;
         if ($this->isBackwardsCompatibilityMode()) {
             if (is_string($options['clientConfig'])) {
                 // perform validation for V1 surfaces which is done in the
--- a/src/GapicClientTrait.php
+++ b/src/GapicClientTrait.php
@@ -238,6 +238,12 @@ trait GapicClientTrait
             'libName',
             'libVersion',
         ]);
+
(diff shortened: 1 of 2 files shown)


src/GapicClientTrait.php
```diff
@@ -238,6 +238,12 @@ private function setClientOptions(array $options)
             'libName',
             'libVersion',
         ]);
+
+        // "hasEmulator" is not a supported Client Option, but is used
+        // internally to determine if the client is running in emulator mode.
+        // Therefore, we need to remove it from the $options array before
+        // creating the ClientOptions.
+        $hasEmulator = (bool) $this->pluck('hasEmulator', $options, false);
```
Hectorhammett, line 246: Is this casting problematic? I see that `pluck` has a `false` parameter, I am assuming that this is a default return value if is not found?

Would this not return a value nor throw an exception in any other cases?

bshaffer: the `false` is (not intuitively) to say that the parameter isn't required. Without that, it would throw an exception if the key is not present in the array.

This parameter is only to be set by us, so it isn't super important to cast it. But the `bool` cast would never throw a type error, so if someone passed an object, stream, etc into `hasEmulator`, it would result in a `bool`. 

```php
var_dump((bool) new stdClass());
// bool(true)
function foo(bool $bar)
{
    var_dump($bar);
}

foo('123');
// bool(true)
foo(new stdClass());
// Uncaught TypeError: foo(): Argument #1 ($bar) must be of type bool, stdClass given
```

I could see how a type error would be preferred, so we could remove this cast so the typehinting is enforced:

```suggestion
        $hasEmulator = $this->pluck('hasEmulator', $options, false);
```

src/GapicClientTrait.php
```diff
@@ -238,6 +238,12 @@ private function setClientOptions(array $options)
             'libName',
             'libVersion',
         ]);
+
+        // "hasEmulator" is not a supported Client Option, but is used
+        // internally to determine if the client is running in emulator mode.
+        // Therefore, we need to remove it from the $options array before
+        // creating the ClientOptions.
+        $hasEmulator = $this->pluck('hasEmulator', $options, false);
```
bshaffer, line 246:
```suggestion
        $hasEmulator = $this->pluck('hasEmulator', $options, false) ?: null;
```

src/GapicClientTrait.php
```diff
@@ -238,6 +238,12 @@ private function setClientOptions(array $options)
             'libName',
             'libVersion',
         ]);
+
+        // "hasEmulator" is not a supported Client Option, but is used
+        // internally to determine if the client is running in emulator mode.
+        // Therefore, we need to remove it from the $options array before
+        // creating the ClientOptions.
+        $hasEmulator = $this->pluck('hasEmulator', $options, false) ?: false;
```
bshaffer, line 246:
```suggestion
        $hasEmulator = $this->pluck('hasEmulator', $options, false) ?? false;
```

Hectorhammett (approved):

PR #598, opened 2024-12-02
chore: drop support for PHP 8.0

Hectorhammett:
Remove tests for PHP 8.0 as we are removing support for it.


.github/sync-repo-settings.yaml
```diff
@@ -5,9 +5,6 @@ branchProtectionRules:
 - pattern: main
   isAdminEnforced: true
   requiredStatusCheckContexts:
-    - 'PHP 8.0 Unit Test'
-    - 'PHP 8.0 Unit Test protobuf,grpc'
-    - 'PHP 8.0 Unit Test --prefer-lowest'
```
bshaffer, line 10: We need to increment the `protobuf,grpc` and `--prefer-lowest` tests to `PHP 8.1` instead of just removing them.

.github/workflows/tests.yml
```diff
@@ -11,18 +11,18 @@ jobs:
         runs-on: ubuntu-latest
         strategy:
             matrix:
-                php: [ "8.0", "8.1", "8.2", "8.3" ]
+                php: [ "8.1", "8.2", "8.3" ]
```
bshaffer, line 14: PHP 8.4 is out, so let's add that one as well.

bshaffer (changes requested):

.github/workflows/tests.yml
```diff
@@ -11,18 +11,18 @@ jobs:
...
                 extensions: [""]
                 tools: [""]
                 composerflags: [""]
                 include:
-                  - php: "8.0"
+                  - php: "8.1"
                     extensions: "protobuf,grpc"
                     tools: "pecl"
                   - php: "8.3"
```
bshaffer, line 22: Let's increment this to PHP 8.4 (and we will need to update `sync-repo-settings.yaml` as well)

.github/workflows/tests.yml
bshaffer: Done in https://github.com/googleapis/gax-php/pull/599

.github/workflows/tests.yml
bshaffer: added in [44be7ac](https://github.com/googleapis/gax-php/pull/598/commits/44be7ac778e6e5fb4f776bc18b1f386b7affbf2c)

.github/sync-repo-settings.yaml
bshaffer: Added in [44be7ac](https://github.com/googleapis/gax-php/pull/598/commits/44be7ac778e6e5fb4f776bc18b1f386b7affbf2c)

bshaffer (approved):

PR #602, opened 2024-12-10
chore(tests): move from MockBuilder to Prophecy

bshaffer:


Hectorhammett (approved):
Omg this is so much better.

Thanks for taking on chore for this 👏

PR #606, opened 2024-12-11
Add missing withResponseMetadata option

ujwaldhakal:
The withResponseMetadata option was missing from this library https://github.com/googleads/google-ads-php/blob/main/src/Google/Ads/GoogleAds/Lib/V17/UnaryGoogleAdsResponseMetadataCallable.php\#L48 because it was using gax-php.

A discussion ticket has been opened here https://github.com/googleads/google-ads-php/issues/1050

### DIFF
--- a/src/Options/CallOptions.php
+++ b/src/Options/CallOptions.php
@@ -54,6 +54,7 @@ class CallOptions implements ArrayAccess
 
     /** @var RetrySettings|array|null $retrySettings */
     private $retrySettings;
+    private bool $withResponseMetadata;
 
     /**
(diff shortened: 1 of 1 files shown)


bshaffer:
@ujwaldhakal the `withResponseMetadata` option is not supported in GAX, or in any GAPIC library as far as I know - it may be supported as a custom option by Google Ads. If that's the case, we'll need to figure out how to accommodate it.

bshaffer:
See https://github.com/googleads/google-ads-php/issues/1050#issuecomment-2536811948

PR #607, opened 2025-01-15
Return the Message error types to the ApiException class from the serializer

Hectorhammett:
This is a prototype for:
b/374174649

### DIFF
--- a/src/Serializer.php
+++ b/src/Serializer.php
@@ -273,14 +273,12 @@ class Serializer
     public static function decodeAnyMessages($anyArray)
     {
         $results = [];
-        $messages = [];
         foreach ($anyArray as $any) {
             try {
                 /** @var Any $any */
                 /** @var Message $unpacked */
                 $unpacked = $any->unpack();
                 $results[] = self::serializeToPhpArray($unpacked);
-                $messages[] = $unpacked;
             } catch (\Exception $ex) {
                 echo "$ex\n";
                 // failed to unpack the $any object - show as unknown binary data
--- a/src/Serializer.php
+++ b/src/Serializer.php
@@ -65,6 +65,18 @@ class Serializer
         'google.rpc.localizedmessage-bin' => \Google\Rpc\LocalizedMessage::class,
     ];
 
+    private static $restKnownTypes = [
+        'type.googleapis.com/google.rpc.RetryInfo' => \Google\Rpc\RetryInfo::class,
+        'type.googleapis.com/google.rpc.DebugInfo' => \Google\Rpc\DebugInfo::class,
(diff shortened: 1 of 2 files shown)


src/Serializer.php
```diff
@@ -231,12 +280,14 @@ public static function decodeMetadata(array $metadata)
...
     {
         $results = [];
+        $messages = [];
         foreach ($anyArray as $any) {
             try {
                 /** @var Any $any */
                 /** @var Message $unpacked */
                 $unpacked = $any->unpack();
                 $results[] = self::serializeToPhpArray($unpacked);
+                $messages[] = $unpacked;
```
bshaffer, line 290: These can be removed, I think

src/Serializer.php
```diff
@@ -231,12 +280,14 @@ public static function decodeMetadata(array $metadata)
     public static function decodeAnyMessages($anyArray)
     {
         $results = [];
+        $messages = [];
```
bshaffer, line 283: This can also be removed

src/Serializer.php
```diff
@@ -185,16 +225,20 @@ public static function decodeMetadata(array $metadata)
         if (count($metadata) == 0) {
             return [];
         }
-        $result = [];
+        $result = [
+            'serialized' => [],
+            'unserialized' => []
+        ];
```
bshaffer, line 231: This looks like a breaking change to me

src/Serializer.php
```diff
@@ -130,6 +142,34 @@ public function encodeMessage($message)
         }
     }
 
+    /**
+     * Encode rest metadata errors to the correct protobuf Error type
+     *
+     * @param array $metadata
+     * @return array
+     */
+    public static function encodeMetadataToProtobufErrors(array $metadata): array
```
bshaffer, line 151: rather than have a different method for REST metadata, would it be possible to make the existing `decodeMetadata` handle both REST and gRPC metadata?

src/Serializer.php
Hectorhammett: You are right, didn't thought as this method as an external one but you're right, this would break compatibility

src/Serializer.php
Hectorhammett: I thought of that but the fact that REST received a completely different structure made me go this way. Handling everything on the `decodeMetadata` method which is already a kind of loaded method, felt a bit convoluted IMO.

Maybe there is a way though, but haven't thought of it.

bshaffer:
@Hectorhammett can this PR be closed now that we have https://github.com/googleapis/gax-php/pull/611 ?

PR #608, opened 2025-01-16
chore: remove github.io reference docs

bshaffer:
Once this is merged, we should ensure that the `gh-pages` branch includes a redirect, similar to what we have in https://github.com/googleapis/google-cloud-php/blob/gh-pages/index.html


Hectorhammett (approved):

bshaffer:
See https://github.com/googleapis/google-cloud-php/pull/7797

PR #611, opened 2025-01-23
feat: Expose errors with the correct protobuf type

Hectorhammett:
Adds a method that exposes the underlying protobuf errors via the `ApiException::getErrorDetails` method.

b/374174649

### DIFF
--- a/src/ApiException.php
+++ b/src/ApiException.php
@@ -179,11 +179,13 @@ class ApiException extends Exception
         ?array $metadata = null,
         ?Exception $previous = null
     ) {
+        $errors = [];
         return self::create(
             $basicMessage,
             $rpcCode,
             $metadata,
-            Serializer::decodeMetadata((array) $metadata),
+            Serializer::decodeMetadata((array) $metadata, $errors),
+            $errors,
             $previous
         );
     }
@@ -208,6 +210,7 @@ class ApiException extends Exception
             $rpcCode,
             $metadata,
             is_null($metadata) ? [] : $metadata,
+            Serializer::encodeMetadataToProtobufErrors($metadata),
             $previous
         );
     }
--- a/src/ApiException.php
+++ b/src/ApiException.php
@@ -144,7 +144,7 @@ class ApiException extends Exception
      * Returns the unserialized errors
      * @return array
      */
-    public function getUnserializedErrors(): array
+    public function getErrorDetails(): array
     {
         return $this->protobufErrors;
     }
--- a/src/ApiException.php
+++ b/src/ApiException.php
@@ -46,6 +46,7 @@ class ApiException extends Exception
     private $metadata;
     private $basicMessage;
     private $decodedMetadataErrorInfo;
+    private array $protobufErrors;
 
     /**
      * ApiException constructor.
@@ -62,7 +63,8 @@ class ApiException extends Exception
         string $message,
         int $code,
         ?string $status = null,
-        array $optionalArgs = []
+        array $optionalArgs = [],
+        array $protobufErrors = [],
     ) {
         $optionalArgs += [
             'previous' => null,
@@ -76,6 +78,7 @@ class ApiException extends Exception
         if ($this->metadata) {
             $this->decodedMetadataErrorInfo = self::decodeMetadataErrorInfo($this->metadata);
         }
+        $this->protobufErrors = $protobufErrors;
     }
 
     public function getStatus()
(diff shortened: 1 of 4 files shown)


src/KnownTypes.php
```diff
@@ -0,0 +1,64 @@
+<?php
+/*
+ * Copyright 2017 Google LLC
```
bshaffer, line 3:
```suggestion
 * Copyright 2025 Google LLC
```

src/Serializer.php
```diff
@@ -562,7 +508,7 @@ private static function getPhpArraySerializer()
 
     public static function loadKnownMetadataTypes()
     {
-        foreach (self::$metadataKnownTypes as $key => $class) {
+        foreach (KnownTypes::allKnownTypes() as $key => $class) {
```
bshaffer, line 511: nit: Why not just 
```php
        foreach (KnownTypes::GRPC_TYPES as $key => $class) {
```

The values are the same for both GRPC and JSON types. Potentially you could also do this
```php
        foreach (KnownTypes::GRPC_TYPES + KnownTypes::JSON_TYPES as $key => $class) {
```

src/ApiException.php
```diff
@@ -137,18 +140,29 @@ public function getErrorInfoMetadata()
         return ($this->decodedMetadataErrorInfo) ? $this->decodedMetadataErrorInfo['errorInfoMetadata'] : null;
     }
 
+    /**
+     * Returns the unserialized errors
```
vchudnov-g, line 144: (not familiar with the code base) What does "unserialized" errors mean?

src/ApiException.php
```diff
@@ -263,11 +282,51 @@ private static function create(
...
+                'previous' => $previous,
+                'metadata' => $metadata,
+                'basicMessage' => $basicMessage,
+            ],
+            $protobufErrors ?? []
+        );
+    }
+
+    /**
+     * Encodes decoded metadata to the Protobuf error type
```
vchudnov-g, line 299: (not familiar with the code base) With `rest` transport, the error is returned from the server as the body of the response. So that does it mean that you're taking it from "metadata"?

src/KnownTypes.php
```diff
@@ -0,0 +1,66 @@
...
+        'type.googleapis.com/google.rpc.RetryInfo' => \Google\Rpc\RetryInfo::class,
+        'type.googleapis.com/google.rpc.DebugInfo' => \Google\Rpc\DebugInfo::class,
+        'type.googleapis.com/google.rpc.QuotaFailure' => \Google\Rpc\QuotaFailure::class,
+        'type.googleapis.com/google.rpc.BadRequest' => \Google\Rpc\BadRequest::class,
+        'type.googleapis.com/google.rpc.RequestInfo' => \Google\Rpc\RequestInfo::class,
+        'type.googleapis.com/google.rpc.ResourceInfo' => \Google\Rpc\ResourceInfo::class,
+        'type.googleapis.com/google.rpc.ErrorInfo' => \Google\Rpc\ErrorInfo::class,
+        'type.googleapis.com/google.rpc.Help' => \Google\Rpc\Help::class,
+        'type.googleapis.com/google.rpc.LocalizedMessage' => \Google\Rpc\LocalizedMessage::class,
+        'type.googleapis.com/google.rpc.PreconditionFailure' => \Google\Rpc\PreconditionFailure::class,
```
vchudnov-g, line 59: Can you handle a custom error detail type that is defined as part of the API but is not on this list? (That's what the internal [go/cloudsdk-gapic-error-details-handling-status](http://goto.google.com/cloudsdk-gapic-error-details-handling-status) is about.)

Work to support this can/should be in a follow-up PR, but it may affect how you handle the standard error details here.

vchudnov-g (commented):
Left some high-level comments, some of them just questions. I'm not an expert on this code base nor on PHP, so I'm not qualified to approve.

src/ApiException.php
bshaffer: In this case, serialized would be binary form and unserialized would be in message form (if possible)

bshaffer (changes requested):
In my testing, the error that comes back from `grpc-status-details-bin` contains an array of error details of type `Any` in `Status::getDetails()`. 

This seems to be okay, as 1. that's the expected type, and changing it throws a typing error, and 2. the user can unpack the any types like this:

```php
try {
    // some API operation
} catch(Google\ApiCore\ApiException $e) {
    $status = $e->getErrorDetails()[0]; // get the first one
    foreach ($status->getDetails() as $any) {
        $message = $any->unpack();
        var_dump($message); // outputs individual message info
    }
}
```

HOWEVER... the issue is that, this should be the same for both REST and gRPC, and it is not. REST decodes the individual messages.

So we need to make gRPC consistent and return an array of decoded messages, OR we need to make REST consistent and return `Google\Rpc\Status`.

src/KnownTypes.php
```diff
@@ -0,0 +1,66 @@
...
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+namespace Google\ApiCore;
+
+class KnownTypes
+{
+    public const GRPC_TYPES = [
+        'google.rpc.retryinfo-bin' => \Google\Rpc\RetryInfo::class,
```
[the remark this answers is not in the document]
bshaffer: After looking into this more, I think the way gRPC is handling this is incorrect - the `Status` message represents the entire response, but it's being put in the `details` array erroneously. Rather, it should be decoded and only the decoded `details` returned in the `details` array.

PR #620, opened 2025-07-31
docs: add template for operation result

bshaffer:
required for https://github.com/googleapis/gapic-generator-php/pull/767

ensures that PHPStan understands the "generic" syntax used in the GAPIC generator for LROs


src/OperationResponse.php
```diff
@@ -56,6 +56,8 @@
  * more control is required, it is possible to make calls against the
  * Operations API directly instead of via the OperationResponse object
  * using an Operations Client instance.
+ *
+ * @template T
```
bshaffer, line 60:
```suggestion
 * @template T = null
```

PR #624, opened 2025-08-27
fix: update protobuf RepeatedField to new namespace

bshaffer:
As we've dropped support for protobuf v3, we can now update our `RepeatedField` to the new (non-internal) namespace

Because this class was moved in version [`4.31`](https://github.com/protocolbuffers/protobuf/releases/tag/v31.0), we must update `conflict` in `composer.json` to be `"ext-protobuf": "<4.31.0"` in order to guaranteed to have this class.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,7 @@
         "phpstan/phpstan": "^2.0"
     },
     "conflict": {
-        "ext-protobuf": "<4.7.0"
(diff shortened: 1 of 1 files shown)


bshaffer:
Removing `next release` label as this is blocked by https://github.com/protocolbuffers/protobuf/pull/23373

PR #630, opened 2025-09-16
feat: Add the rpcName to the BIDI stream opening request

Hectorhammett:
This PR adds the `rpcName` field to the logs when opening a BIDI stream.

### DIFF
--- a/src/Transport/GrpcTransport.php
+++ b/src/Transport/GrpcTransport.php
@@ -178,7 +178,7 @@ class GrpcTransport extends BaseStub implements TransportInterface
     {
         $this->verifyUniverseDomain($options);
 
-        return new BidiStream(
+        $bidiStream = new BidiStream(
             $this->_bidiRequest(
                 '/' . $call->getMethod(),
                 [$call->getDecodeType(), 'decode'],
@@ -188,6 +188,21 @@ class GrpcTransport extends BaseStub implements TransportInterface
             $call->getDescriptor(),
             $this->logger
         );
+
+        if ($this->logger) {
+            $requestEvent = new RpcLogEvent();
+
+            $requestEvent->headers = $options['headers'];
+            $requestEvent->retryAttempt = $options['retryAttempt'] ?? null;
+            $requestEvent->serviceName = $options['serviceName'] ?? null;
+            $requestEvent->rpcName = $call->getMethod();
+            $requestEvent->processId = (int) getmypid();
+            $requestEvent->requestId = crc32((string) spl_object_id($bidiStream) . getmypid());
+
+            $this->logRequest($requestEvent);
+        }
+
+        return $bidiStream;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


src/Transport/GrpcTransport.php
```diff
@@ -188,6 +188,21 @@ public function startBidiStreamingCall(Call $call, array $options)
             $call->getDescriptor(),
             $this->logger
         );
+
+        if ($this->logger) {
+            $requestEvent = new RpcLogEvent();
+
+            $requestEvent->headers = $options['headers'];
```
bshaffer, line 195: Shouldn't this also be 
```suggestion
            $requestEvent->headers = $options['headers'] ?? [];
```

I see that on line 185 it's checking if the key is set

tests/Unit/Transport/GrpcTransportTest.php
```diff
@@ -275,6 +275,34 @@ public function testBidiStreamingSuccessSimple()
...
+        $status->code = Code::OK;
+
+        $bidiStreamingCall = $this->prophesize(\Grpc\BidiStreamingCall::class);
+
+        $transport = new MockGrpcTransport(
+            $bidiStreamingCall->reveal(),
+            logger: new StdOutLogger()
+        );
+
+        /* @var $stream \Google\ApiCore\BidiStream */
```
bshaffer, line 291: nit: 
```suggestion
        /** @var \Google\ApiCore\BidiStream $stream */
```

tests/Unit/Transport/GrpcTransportTest.php
```diff
@@ -275,6 +275,34 @@ public function testBidiStreamingSuccessSimple()
...
+            ['headers' => [
+                ['thisis' => 'a header']
+            ]]
+        );
+
+        $buffer = $this->getActualOutput();
+        $unserializedBuffer = json_decode($buffer, true);
+
+        $this->assertNotEmpty($unserializedBuffer);
+        $this->assertNotEmpty($unserializedBuffer['rpcName']);
```
bshaffer, line 303: Can we assert what the `rpc` name is? like this:
```php
$this->assertEquals('takeAction', $unserializedBuffer['rpcName']);
```

tests/Unit/Transport/GrpcTransportTest.php
```diff
@@ -301,6 +301,7 @@ public function testBidiOpeningRequestLogsRpcName()
 
         $this->assertNotEmpty($unserializedBuffer);
         $this->assertNotEmpty($unserializedBuffer['rpcName']);
+        $this->assertEquals($unserializedBuffer['rpcName'], $rpcName);
```
bshaffer, line 304: one final nit but, in PHPUnit, the expected value goes first. So this should be 
```suggestion
        $this->assertEquals($rpcName, $unserializedBuffer['rpcName']);
```

This can be important for the error output, because it will say "Expected string Foo does not match expected Bar", which can be important when debugging (I've had the order being wrong mess me up before)

Hectorhammett: Damn it! I always make this mistake!

PR #632, opened 2025-09-19
Add GOOGLE_API_TRANSPORT environment variable support

sl0wik:
This PR adds support for configuring the transport method via the `GOOGLE_API_TRANSPORT` environment variable, allowing users to set the transport to rest, grpc, or grpc-fallback at the environment level.

#  Problem

Currently, the transport method must be manually configured for each Google Cloud PHP client library implementation by passing `['transport' => 'rest']` in the options array. This approach presents several challenges:

- When managing multiple client libraries across an application, developers must remember to set the transport option for each individual client instantiation.
- The same transport configuration must be repeated throughout the codebase.

# Solution

This PR introduces the `GOOGLE_API_TRANSPORT` environment variable that allows setting the transport method globally. 

# Usage

```
GOOGLE_API_TRANSPORT=rest php my-script.php
```

### DIFF
--- a/src/ClientOptionsTrait.php
+++ b/src/ClientOptionsTrait.php
@@ -124,6 +124,14 @@ trait ClientOptionsTrait
             'logger' => null,
         ];
 
+        // Check for transport method in environment variable if not set in options
+        if (!isset($options['transport'])) {
+            $envTransport = getenv('GOOGLE_API_TRANSPORT');
+            if ($envTransport !== false && in_array($envTransport, ['rest', 'grpc', 'grpc-fallback'])) {
+                $options['transport'] = $envTransport;
+            }
+        }
+
         $supportedTransports = $this->supportedTransports();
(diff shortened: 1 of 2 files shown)


bshaffer:
@sl0wik I do sympathize with your frustration with the [current bug in gRPC](https://github.com/grpc/grpc/issues/39509), and your desire to use REST in your development environment. But we do not want to support an environment variable for this use-case, or encourage this workflow. I imagine your codebase has a way to change configurations depending on dev/prod/etc, and I would suggest you use this to set the `rest` vs `grpc` transport when creating the service client. For example:


```php
use Google\Cloud\SecretManager\V1\Client\SecretManagerClient;
$transort = Environment::getConfig('secret_manager_transport'); // either "rest" or "grpc"
$secretmanager = new SecretManagerClient(['transport' => $transport]);
```

PR #633, opened 2025-10-20
fix: add return type to `offsetGet`

DjordyKoert:
Fixes deprecation warning:

> `User Deprecated: Method "ArrayAccess::offsetGet()" might add "mixed" as a native return type declaration in the future. Do the same in implementation "Google\ApiCore\Options\ClientOptions" now to avoid errors or add an explicit @return annotation to suppress this message.`

### DIFF
--- a/src/Options/OptionsTrait.php
+++ b/src/Options/OptionsTrait.php
@@ -57,6 +57,9 @@ trait OptionsTrait
         return isset($this->$offset);
     }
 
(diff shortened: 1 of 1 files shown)


bshaffer:
Hi @DjordyKoert ! Thank you for your contribution.

Where are you seeing that error? Is that when using PHPStan or some other static analyzer?

PR #640, opened 2025-12-02
feat: Add a TransportCallMiddleware

Hectorhammett:
We added a new feature to the GapicClientTrait adding the new method: `pependMiddleware`.

Adding a middleware involves passing a closure to the `addMiddleware` and `prependMiddleware` with the first parameter being a `MiddlewareInterface` like so:
```php
$client->addMiddleware(function (MiddlewareInterface $handler) {...});
```

For `prependMiddleware` this is not always the case as the final call in the `callStack` is not a `MiddlewareInterface` but a closure.

This PR moves this closure into its own middleware so the `$callStack` variable is always of type `MiddlewareInterface` fixing the problem with the `prependMiddleware` class and adding more consistency to typing.

### DIFF
--- a/src/Middleware/TransportCallMiddleware.php
+++ b/src/Middleware/TransportCallMiddleware.php
@@ -34,12 +34,15 @@ namespace Google\ApiCore\Middleware;
 use Google\ApiCore\Call;
 use Google\ApiCore\Transport\TransportInterface;
 
+/**
+ * A Middleware in charge of handling the end of the callstack to call the transport layer.
+ * This middleware is made so the callstack in the GapicClientTrait is always a middleware.
+ *
+ * @internal
+ */
 class TransportCallMiddleware implements MiddlewareInterface
 {
-    /**
-     * A Middleware in charge of handling the end of the callstack to call the transport layer.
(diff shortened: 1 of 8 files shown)


src/Middleware/TransportCallMiddleware.php
```diff
@@ -0,0 +1,57 @@
...
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+namespace Google\ApiCore\Middleware;
+
+use Google\ApiCore\Call;
+use Google\ApiCore\Transport\TransportInterface;
+
```
bshaffer: Do we want to add `@internal`? We didn't do it for the other Middleware classes but I feel like we should

PR #643, opened 2025-12-31
fix: Use self:: instead of $this-> for static validation methods in GapicClientTrait validation 

### LINKED ISSUE #642: Too few arguments to function Google\Cloud\Tasks\V2\Client\CloudTasksClient::validateNotNull(), 2 passed in /code/vendor/google/gax/src/GapicClientTrait.php on line 264 and exactly 2 expected
php 8.5.1

        "name": "google/gax",
        "version": "v1.40.0",

        "name": "google/cloud-tasks",
        "version": "v2.1.1",

Maybe an opcache issue?
Yes diabling the OpCache fixes the issue

or also 

```
--- a/src/GapicClientTrait.php
+++ b/src/GapicClientTrait.php
@@ -261,7 +261,7 @@
         if (isset($options['serviceAddress'])) {
             $options['apiEndpoint'] = $this->pluck('serviceAddress', $options, false);
         }
-        $this->validateNotNull($options, [
+        self::validateNotNull($options, [
             'apiEndpoint',
             'serviceName',
             'descriptorsConfigPath',

```

`validateNotNull` is defined as a static function in the trait.
It was called using `$this->validateNotNull(...)`.
Standard PHP (No OpCache): PHP is forgiving. When it sees `$this->staticMethod()`, it realizes the method is static, ignores `$this`, and runs it correctly. It's a "dynamic" resolution at runtime.

OpCache tries to optimize code before it runs. When it sees `$this->`, it prepares the code execution assuming it involves an object instance.
In complex cases involving Traits + Inheritance, OpCache can get confused about how many arguments are actually on the stack.
It might count the invisible `$this` as an argument, or fail to strip it out correctly before handing off to the static function.
The engine ends up thinking: "Wait, I have a stack set up for an object method, but I'm entering a static function signature," resulting in that confusing "Too few arguments" error (even when the numbers look right).

maybe add some php85 unit test in the pipeline and also maybe use something with opcache enabled 

```
opcache.enable = 1
opcache.enable_cli = 1
opcache.jit_buffer_size = 128M
auto_globals_jit = On
pcre.jit = 1
opcache.jit = 1255
```

Image

maybe not the entire solution

romanstingler:
This PR fixes a critical issue where enabling OpCache (specifically with JIT or strict optimization levels) causes a fatal error: `Too few arguments to function `Google\Cloud...\GapicClientTrait::validateNotNull()`, 2 passed ... and exactly 2 expected`

The methods `validateNotNull` and `traitValidate` (aliased from `ValidationTrait::validate`) are defined as static. However, they are currently called using instance syntax (`$this->...`).

While valid in standard PHP, some OpCache/JIT configurations misinterpret the stack frame when a static method is invoked via `$this`, potentially counting the implicit object instance as a misaligned argument or failing to discard it correctly before the static call.

I have updated the calls to use `self::` instead of `$this->`. 
This explicitly invokes them as static methods, ensuring the correct opcode generation and preventing the argument count mismatch error in optimized environments.

closes #642

what is this code? 
opening it in an IDE is like a Christmas tree 
```
'str*()' calls can be replaced with PHP 8 'str_*()' calls
::class constant usage correctness
Callable parameter usage violates definition
Deprecated
Fully qualified name usage
Grammar
Incompatible return type
JSON encoding API usage
Missing @throws tag(s)
Missing parameter's type declaration
Missing return type declaration
Missing strict types declaration
Nested positive ifs
Null reference
One-time use variables
PHPDoc comment matches function/method signature
PHPDoc comment signature is not complete
Possible polymorphic call
Referencing objects
Return type hint can be used
Return type in PHPDoc comment does not match the actual return type
Spelling
Type unsafe comparison
Undefined class
Unhandled exception
Unnecessary fully qualified name
Unqualified function/constant reference
```

### DIFF
--- a/src/GapicClientTrait.php
+++ b/src/GapicClientTrait.php
@@ -261,7 +261,7 @@ trait GapicClientTrait
         if (isset($options['serviceAddress'])) {
             $options['apiEndpoint'] = $this->pluck('serviceAddress', $options, false);
         }
-        $this->validateNotNull($options, [
+        self::validateNotNull($options, [
             'apiEndpoint',
             'serviceName',
             'descriptorsConfigPath',
@@ -270,7 +270,7 @@ trait GapicClientTrait
             'credentialsConfig',
             'transportConfig',
         ]);
-        $this->traitValidate($options, [
+        self::traitValidate($options, [
             'credentials',
             'transport',
             'gapicVersion',


bshaffer:
> what is this code?
opening it in an IDE is like a Christmas tree

@romanstingler the items you listed do not provide a clear way to make changes. But please feel free to share more specifically how this code can be improved and we will be happy to take action

PR #664, opened 2026-04-12
fix(serializer): catch Throwable to handle PHP 7+ Errors

mmucklo:
PHP7+ introduced the following hierarchy: https://www.php.net/manual/en/language.errors.php7.php

There's a base Throwable class which both Error and Exception derive from.

When using the protobuf.so C extension, it's possible that an error message such as "Class zetasql.ErrorMessageModeForPayload hasn't been added to descriptor pool" can manifest as a TypeError or Error which don't derive from Exception and thus cause a fatal instead of being caught.

### DIFF
--- a/src/Serializer.php
+++ b/src/Serializer.php
@@ -252,7 +252,7 @@ class Serializer
                 /** @var Message $unpacked */
                 $unpacked = $any->unpack();
                 $results[] = self::serializeToPhpArray($unpacked);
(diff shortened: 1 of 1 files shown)


bshaffer (approved):
Thank you for finding and fixing this issue!

Hectorhammett (approved):

Hectorhammett:
Good catch! Pun intended!

PR #676, opened 2026-06-04
fix: RetryMiddleware should sleep on retry

bshaffer:
Retry Middleware was calculating a delay but was not using it to sleep

fixes https://github.com/googleapis/gax-php/issues/673


src/Middleware/RetryMiddleware.php
```diff
@@ -48,6 +48,8 @@ class RetryMiddleware implements MiddlewareInterface
     private $nextHandler;
     private RetrySettings $retrySettings;
     private ?float $deadlineMs;
+    /** @var callable */
```
Hectorhammett, line 51: Maaan if only Callable was a type :(

Hectorhammett (approved):

PR #679, opened 2026-06-09
fix: Add the rpc uri to the logs

Hectorhammett:
closes: googleapis/google-cloud-php#9254


bshaffer (changes requested):
This is great! But it needs to be resubmitted to https://github.com/googleapis/google-cloud-php:

```bash
cd /path/to/google-cloud-php
SHA=472829f3d15ab78f7e1dfacfd128485e1ac5f48e
git -C /path/to/gax-php format-patch -1 $SHA --stdout | git apply --directory=Gax
```

